PR #105, opened 2022-04-25
fix: update to use postgres for test

mkobetic:
Switch to use latest node-go and consequently the postgres message store.

The change highlighted some flakiness in our tests, presumably due to higher latencies to the now external DB. Switching to lightpush from relay for sending contacts and private keys to the network improved it, although I still see some flakiness locally. Presumably because the lightpush confirmation doesn't necessarily mean that the message has actually reached the store.

# Questions

1. Do we want to use lightpush instead of relay for these?
2. If yes, do we want to check the lightpush response and throw in some cases?
3. Do we want to further fortify the tests against flakiness? I've seen mostly the Conversations tests flaking out [with "Address X is not on the network"](https://github.com/xmtp/xmtp-js/blob/main/src/conversations/Conversations.ts#L89) presumably because they were too fast grabbing the contact info

### DIFF
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -1,12 +1,22 @@
 services:
   waku-node:
-    image: xmtp/node-go@sha256:489f72a4049272989b42032f6fe23aac91cfaf13c93605017d4dce93d21f4993
+    image: xmtp/node-go:latest
     environment:
       - GOWAKU-NODEKEY=8a30dcb604b0b53627a5adc054dbf434b446628d4bd1eccc681d223f0550ce67
     command:
       - --ws
       - --store
+      - --message-db-connection-string=postgres://postgres:xmtp@db:5432/postgres?sslmode=disable
       - --lightpush
       - --ws-port=9001
+      - --wait-for-db=30s
     ports:
       - 9001:9001
+    depends_on:
+      - db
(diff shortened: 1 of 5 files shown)


dev/docker-compose.yml
```diff
@@ -1,12 +1,22 @@
...
+      - --wait-for-db=30s
     ports:
       - 9001:9001
+    depends_on:
+      - db
+  db:
+    image: postgres:13
+    environment:
+      POSTGRES_PASSWORD: xmtp
+    ports:
```
neekolas, line 21: Don't need to expose this port to the host, which'll lower the chance of port conflicts with other instances locally.

neekolas:
Some delay is to be expected now that we are writing to an actual DB.

## Answers to questions
1. Can't hurt. I think we generally want to move to LightPush across the board so that we can remove Relay from the client. Would be surprised if this solves the issue fully, as the LightPush flow in Waku is `receive lightpush message -> push to Relay -> send message to store channel -> write to database`. Switching to LightPush only blocks until the end of the first step there.
2. I think so
3. Even very short delays at the right point (maybe `50ms`) could resolve the issue. This should be relatively snappy to execute in practice

neekolas (approved):

PR #110, opened 2022-05-27
Use waku filter

neekolas:
## Summary
- Replaces usage of Waku Relay in the Stream module with usage of Waku Filter.
- Adds support for the `dev` environment and sets it as the default
- Refactor tests to not copy/paste all the client config

## Notes
- This is a breaking change, as initiating the stream is now an asynchronous operation (it has to send a request to the server to start the stream) as opposed to the Relay way where the client was already getting the firehose of messages and it just needed to add an event listener. I have updated the tests and README appropriately.
- This PR does not yet turn off Relay. That will come next, and should be a non-breaking change. That _may_ require a separate, but smaller, PR to `js-waku` to enable in a nice way. I'm going to open up an issue to discuss options with the `js-waku` team. There are a number of possible ways to implement.
- This will only work well against the AWS-backed dev environment, as the `testnet` still has the severe bug in the Waku Filter protocol. Merging this PR will switch all clients to default to connecting to the new `dev` environment

### DIFF
--- a/test/helpers.ts
+++ b/test/helpers.ts
@@ -5,9 +5,13 @@ import {
   ContentCodec,
   ContentTypeId,
   TextCodec,
+  Client,
 } from '../src'
 import Stream from '../src/Stream'
 import { promiseWithTimeout } from '../src/utils'
+
+const LOCAL_DOCKER_MULTIADDR =
+  '/ip4/127.0.0.1/tcp/9001/ws/p2p/16Uiu2HAmNCxLZCkXNbpVPBpSSnHj9iq4HZQj7fxRzw2kj1kKSHHA'
 
 export const sleep = (ms: number): Promise<void> =>
   new Promise((resolve) => setTimeout(resolve, ms))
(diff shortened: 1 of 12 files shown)


test/helpers.ts
```diff
@@ -89,3 +93,11 @@ export class CodecRegistry {
...
   }
 }
+
+export const newLocalDockerClient = (): Promise<Client> =>
+  Client.create(newWallet(), {
+    bootstrapAddrs: [LOCAL_DOCKER_MULTIADDR],
+  })
+
+export const newDevClient = (): Promise<Client> =>
+  Client.create(newWallet(), { env: 'dev' })
```
snormore, line 103: Where does `dev` translate into a node list?

neekolas: Oh ya, we have our own fleets file. https://github.com/xmtp-labs/xmtp-nodes-service

snormore (approved):
LGTM

neekolas:
For @snormore's benefit, this PR is the followup to [this project](https://github.com/status-im/js-waku/pull/711) bringing Waku Filter to the browser. We got a lot of support and engagement from the VAC team on it, which has been nice.

mkobetic (approved):

Issue #112, opened 2022-06-03
Test 'filters out spoofed messages' failing intermittently 


snormore:
The test seems to be flakey, failing more often than not - noticed mostly in the `test:jsdom` env so far.

```
  ● Client › local docker node › filters out spoofed messages

    thrown: "Exceeded timeout of 5000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      264 |       })
      265 |
    > 266 |       it('filters out spoofed messages', async () => {
          |       ^
      267 |         const stream = await bob.streamConversationMessages(alice.address)
      268 |         // mallory takes over alice's client
      269 |         const malloryWallet = newWallet()
```

https://github.com/xmtp/xmtp-js/runs/6725272984?check_suite_focus=true

snormore:
Seems to hang on this line https://github.com/xmtp/xmtp-js/blob/main/test/Client.test.ts#L277

snormore:
@neekolas it looks like if I revert https://github.com/xmtp/xmtp-js/pull/110 then this stops happening

snormore:
Doesn't seem to be mitigated by increasing the test timeout, so it seems that the message possibly isn't being emitted sometimes.

snormore:
Running the test on it's own seems to work pretty consistently, at least the number of times I've tried it:
```sh
npm run test:jsdom -- -t 'Client local docker node filters out spoofed messages'
```

Running the group of client local node tests will reproduce the intermittent failures though:
```sh
npm run test:jsdom -- -t 'Client local docker node'
```

neekolas:
My hunch is that there is a race between the filter being established and the message being received on the backend. 

Try adding a 50ms sleep before the first message is sent and see if that resolves it.

snormore:
Just tried that and also adding 1s sleeps between pretty much everything but still seeing it

snormore:
Ok adding a delay before the the first line (`const stream = await bob.streamConversationMessages(alice.address)`) in the test seems to mitigate it - so something about sharing/restreaming across tests

neekolas:
That's surprising. 

Yes, the filter gets initiated in that function by sending a FilterRequest to the node.

PR #113, opened 2022-06-03
Add missing await to stream.return() in client test

snormore:
Fixes https://github.com/xmtp/xmtp-js/issues/112

### DIFF
--- a/test/Client.test.ts
+++ b/test/Client.test.ts
@@ -260,7 +260,7 @@ describe('Client', () => {
           alice.sendMessage(bob.address, key, { contentType: type2 })
         ).rejects.toThrow('unknown content type xmtp.test/public-key:2.0')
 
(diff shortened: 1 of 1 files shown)


neekolas (approved):

neekolas:
Not quite sure I grok how this fixes that flaky test, but it's a bug regardless

snormore:
@neekolas yea it could just be the slight delay it's adding by waiting on the return that happens immediately before it - does seem to result in consistently passing tests though - we can keep the issue open if you want to dig into it some more

snormore:
It looks like the flakey test subscription is racing with the subscription that happens in `unsubscribeFn` in this return, since they both use the same topic - so that unsubscribe is draining the message before the flakey test subscribe can consume it.

PR #117, opened 2022-06-16
Disable relay listener by default with option to enable

snormore:
This PR adds support for a flag that disables/removes the waku relay listeners, specifically [this one](https://github.com/status-im/js-waku/blob/master/src/lib/waku_relay/index.ts#L216 ). In our case we want the flag to disable by default, so it's an enable/opt-in flag - since our use at this level is exclusively the filter interface.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "cross-fetch": "^3.1.5",
         "ethers": "^5.5.3",
         "js-waku": "^0.24.0",
-        "protobufjs": ">=6.11.3"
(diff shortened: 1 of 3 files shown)


snormore:
Makes sense, seems reasonable to start by removing [this whole part](https://github.com/status-im/js-waku/blob/master/src/lib/waku_relay/index.ts#L216-L251) in the meantime though right?

neekolas:
Ya. I mean, this will help some.

neekolas (approved):

PR #120, opened 2022-06-23
SDK Authn Protocol Integration

jazzz:
Js SDK will need to provide authentication details to the node in order to be allowed to send messages. This PR implements the core functionality required to achieve that. 

Not implemented:
- Exception handling :: Currently there is no mechanism to inform the client that the send request was denied because the client is not authenticated. For now an error is thrown.


test/testClient.ts
```diff
@@ -0,0 +1,41 @@
+import { Signer } from 'ethers'
```
snormore, line 1: Maybe call this file `TestClient.ts` to match the class that's in it?

src/authn/Authenticator.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.walletAddress = identityKey.publicKey.walletSignatureAddress()
+    this.sender = sender
+  }
+
+  static create(
+    libp2p: Libp2p,
+    identityKey: PrivateKey,
+    authOpts?: AuthOptions
+  ): Authenticator {
+    const sender = authOpts?.alternativeSender
```
snormore, line 44: I was a little confused by this `alternativeSender` option at first too; why not just call it `sender`?

src/Client.ts
```diff
@@ -279,7 +297,22 @@ export default class Client {
   }
 
   private async sendWakuMessage(msg: WakuMessage): Promise<void> {
-    const ack = await this.waku.lightPush.push(msg)
+    // Waku randomly selects a peer from the Peerstore to send the message to. To ensure this is the
+    // same peer to which we authenticated to, a random peer is selected at this context and then
```
snormore, line 301: Maybe a dumb question but why does it need to be the same peer/node?

test/exports/Export.test.ts
```diff
@@ -0,0 +1,19 @@
+// This test suite is used to generate test cases to be exported to other implementations and is not run as part of
```
snormore, line 1: I was confused by the export thing at first but I think this comment helps. I'm not sure how I feel about it tbh, still maybe don't fully get it - is it being used for anything right now?

snormore (commented):
A few comments mostly non-functional and a couple questions. Looks good from my pov. Will do another pass tomorrow or whenever it makes sense.

src/Client.ts
jazzz: Authentication state is not replicated between nodes. So if a client authenticates with Node0 but sends the message to Node1 then the message will fail. 


I'll update the comment with more detail

test/exports/Export.test.ts
jazzz: Yeah I'm not super happy about this either, as its co-opting a test system to run tooling/scripts and I am very open to a better solution. 

The test cases are used to detect breaking changes in the node software. As this library is currently the canonical reference implementation, known good requests are exported and run against the node software to ensure the end-to-end functionality. This acts as a 'reverse integration' test of sorts. 

Options:
- Run it as a external script :: would require adding additional dependencies( ts-node ) 
- Move into separate repo 
- Remove the tests entirely.

src/crypto/PrivateKey.ts
```diff
@@ -54,7 +54,7 @@ export default class PrivateKey implements proto.PrivateKey {
   }
 
   // sign provided digest
-  async sign(digest: Uint8Array): Promise<Signature> {
+  async sign(digest: Uint8Array | string): Promise<Signature> {
```
mkobetic, line 57: Do we really need this? I'd rather not have that. If we do then we should be clear on how is the string interpreted (i.e. converted to bytes).

jazzz: Thats fair. Will remove

src/Client.ts
```diff
@@ -279,7 +297,22 @@ export default class Client {
...
+    // passed in to LightPush to ensure a match
+    const dstPeer = await this.waku.lightPush.randomPeer
```
mkobetic, line 303: Hm, is this going to make clients use all lightpush nodes? Wouldn't we want to have the clients stick with fewer rather than more?

src/proto/authReq.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
neekolas: Maybe `authn.proto`?

src/Client.ts
jazzz: I agree. I think not randomly selecting a peer would be preferred, but that seems like a bigger change that I'd want to handle outside of this PR. Waku randomly selects the peer by default, so this intends to matches that behavior. 

I also think in practice this is not a big deal as it seems. I traced the SDK looking for where peers get added to the `libp2p.peerStore` (which is where randomPeer samples from) and couldn't find any references in`xmtp-js` and `js-waku`. 

I've yet to confirm but it appears the only place place a peer gets added to the `peerStore.addressBook` is in `Libp2p.dialer.dial`. This would mean that its only randomly selecting peers to which its previously connected to. 

ref: https://github.com/libp2p/js-libp2p/blob/e6f646ed361d231ec2306fee40fcb87c18af0c41/src/connection-manager/dialer/index.ts#L171

neekolas: Yes. In practice it is always one node per client. That node is picked randomly from the bootstrap list, but then stays consistent for the browser session. Maybe Waku is going to fix this at some point, but that's the way it works right now.

mkobetic: Sounds good 👍

src/authn/AuthRequests.ts
```diff
@@ -0,0 +1,121 @@
...
+
+import { PrivateKey } from '../crypto'
+import * as proto from '../proto/authReq'
+
+export class AuthRequest {
+  public constructor(public proto: proto.ClientAuthRequest) {
+    this.proto = proto
+  }
+
+  static async createRequest(
```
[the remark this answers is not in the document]
jazzz: fair enough

src/proto/authReq.proto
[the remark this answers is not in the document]
jazzz: changed to authn.proto

test/exports/Export.test.ts
jazzz: I've removed the TestCase exporting as needs more work. Its presence is largely counter productive and adds more confusion than anything. If needed we can revisit.

src/authn/Authenticator.ts
```diff
@@ -0,0 +1,89 @@
+import Libp2p from 'libp2p'
+import PeerId from 'peer-id'
+
+import { AuthnRequest } from './AuthnRequest'
+import { AuthnResponse } from './AuthnResponse'
+import { AuthnSender, ProductionAuthnSender } from './AuthnSender'
+import { PrivateKey } from '../crypto'
+
+const PROTO_AUTHN = '/xmtplabs/xmtp-v1/clientauthn/0.1.0-beta1'
```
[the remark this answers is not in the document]
neekolas: If we are going to drop it later, why not drop it now?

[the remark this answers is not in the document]
jazzz: If any further changes are required to the nodes, I'd rather batch those together. Figured I'd defer the change a couple of hours to see if anything pops up in review that could be addressed at the same time.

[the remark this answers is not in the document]
neekolas: I guess my question is "why does the version matter if the protocol isn't actually in use by anyone?"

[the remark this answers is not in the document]
jazzz: > why does the version matter if the protocol isn't actually in use by anyone?

It does not matter as long as theres a match between the node software and the client. Removing the "beta" tag needs to be done in two places. This has no technical impact, only a workflow order of operations. 

For ease, I'll push that change now so it's not a distraction.

neekolas (approved):

PR #121, opened 2022-06-27
KeyBundle Versioning

jazzz:
PrivateKeyBundles stored on the network do not contain any versioning, which will complicate handling of future changes to the bundles.

This PR adds a version field to the PrivateKeyBundle and creates a new ContactBundle object to handle stored public key bundles on the network.


src/proto/messaging.proto
```diff
@@ -32,6 +32,16 @@ message PublicKeyBundle {
   PublicKey preKey = 2;
 }
 
+message ContactBundleV1 {
+  PublicKeyBundle keyBundle = 1;
+}
+
+message ContactBundle {
```
snormore, line 39: How about just `Contact`?

snormore (approved):

src/proto/messaging.proto
jazzz: What would be goal be moving from ContactBundle to Contact?

snormore: It's simpler and reflects the topic that it lives on (`contact-*`)

jazzz: the term `Contact` has multiple definitions and is already a source of ambiguity, and I was hoping to avoid it. 

Contact is used at a high level to refer an individual (who is either on the network or is a participant in an existing conversation), most often referenced via a walletAddress.

the `contactBundle` in this case refers to the object persisted to the network which contains the information required to the reach a contact.

PR #122, opened 2022-06-28
Remove default start time on message query

snormore:
Removes the default `startTime` on historic message queries of the recent 7 days.


test/Client.test.ts
```diff
@@ -176,6 +176,37 @@ describe('Client', () => {
         convos.forEach((m, i) => assert.equal(m.content, messages[i]))
       })
 
+      it('query historic messages', async () => {
+        const c1 = await testCase.newClient()
+        const c2 = await testCase.newClient()
```
jazzz, line 181: Is there a reason to not use the Alice and Bob Client instances that are automatically created for this test case?

 Initializing new clients will result in excess messages on the network as they publish contactBundles on initialization

test/Client.test.ts
```diff
@@ -176,6 +176,37 @@ describe('Client', () => {
...
+        await c1.getUserContactFromNetwork(c2.address)
```
jazzz, line 182: GetUserContactFromNetwork bypasses the contact bundle cache. Is that the intended behaviour here?

test/Client.test.ts
snormore: I'm not a huge fan of reusing state across tests tbh but I'll update this to be consistent with the others.

test/Client.test.ts
snormore: Yea I had to put this here to wait for the contact to be available on the network, otherwise it would race with the publish and fail.

test/Client.test.ts
jazzz: My bad, I mistakenly thought that the clients were being generated once per test. Makes sense to me to start with a new client to ensure a clean state

jazzz (approved):

PR #125, opened 2022-07-06
Jazzz/bundle version prep

jazzz:
This PR adds decoding for versioned bundles. If a versioned bundle cannot be decoded it attempts to parse it as a 'LegacyBundle' as a fallback. 

Once majority of clients have had a change to update, versioned bundles will be enabled by default.

### DIFF
--- a/src/proto/messaging.proto
+++ b/src/proto/messaging.proto
@@ -32,6 +32,16 @@ message PublicKeyBundle {
   PublicKey preKey = 2;
 }
 
+message ContactBundleV1 {
+  PublicKeyBundle keyBundle = 1;
+}
+
+message ContactBundle {
+  oneof version {
+    ContactBundleV1 v1 = 1;
+  }
+}
+
 // ContentTypeId is used to identify the type of content stored in a Message.
 message ContentTypeId {
   string authorityId = 1;  // authority governing this content type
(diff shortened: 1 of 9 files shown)


src/proto/messaging.proto
```diff
@@ -32,6 +32,16 @@ message PublicKeyBundle {
   PublicKey preKey = 2;
 }
 
+message ContactBundleV1 {
+  PublicKeyBundle keyBundle = 1;
+}
+
+message ContactBundle {
+  oneof version {
+    ContactBundleV1 v1 = 1;
```
mkobetic, line 41: Couldn't this just be `PublicKeyBundle v1 = 1;` ? It's not clear to me what we gain with adding ContactBundleV1 here.

test/ContactBundle.test.ts
```diff
@@ -0,0 +1,18 @@
+import * as assert from 'assert'
+import ContactBundle from '../src/ContactBundle'
+import * as ethers from 'ethers'
+import { PrivateKeyBundle } from '../src'
+
+describe('ContactBundles', function () {
+  it('roundtrip', async function () {
+    const priv = PrivateKeyBundle.generate()
+
+    const cb1 = new ContactBundle((await priv).getPublicKeyBundle())
```
mkobetic, line 10: Just so that I understand what's going on here, what does the `(await priv).` bit mean? Is it the same as putting the await on line 8?

jazzz: Yes. Moved it up a line to be more standard

PR #130, opened 2022-07-15
Deploy versioned bundles

jazzz:
This PR enables versioned EncryptedPrivateBundles and ContactBundles. Clients will now create this bundles by default, and previous EncryptedPrivateBundles will be upgraded to the new format

### DIFF
--- a/src/store/EncryptedStore.ts
+++ b/src/store/EncryptedStore.ts
@@ -43,7 +43,7 @@ export default class EncryptedStore {
       )
     } catch (e) {
       // If a versioned bundle is not found, the legacy bundle needs to be resaved to the store in
-      // the new format. One all bundles have been upgraded, this migration code can be removed.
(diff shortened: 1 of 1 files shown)


src/store/PrivateTopicStore.ts
```diff
@@ -15,7 +15,7 @@ export default class NetworkStore implements Store {
     const contents = (
       await this.waku.store.queryHistory([buildUserPrivateStoreTopic(key)], {
         pageSize: 1,
-        pageDirection: PageDirection.FORWARD,
+        pageDirection: PageDirection.BACKWARD,
         callback: function (msgs: WakuMessage[]): boolean {
           return Boolean(msgs[0].payload)
```
neekolas, line 20: I guess we don't need to rush into a fix for malicious/broken bundles written to the private store topic. But we should probably figure out what our strategy is there.

neekolas (approved):

PR #132, opened 2022-07-20
Allow key bundle import/export

neekolas:
## Summary
- Adds a new static method to the client `getKeys()` which will either create or load the keys for the given wallet and return them as a `Uint8Array`. This is to ensure that usage of this method does not lead to key fragmentation for users with existing keys on the network
- Renames `encode()` and `decode()` methods on `PrivateKeyBundle` to `toEncryptedBytes` and `fromEncryptedBytes` to more clearly denote that the methods are not just encoding them as a protobuf, but are also encrypting the payload using the wallet
- Makes the `wallet` argument nullable when creating a client. Will throw an exception if neither `wallet` or `privateKeyOverride` is provided on client startup
- Adds tests
- Updates README

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,18 +6,33 @@
 
 ![x-red-sm](https://user-images.githubusercontent.com/510695/163488403-1fb37e86-c673-4b48-954e-8460ae4d4b05.png)
 
-**Pre-stable XMTP client implementation for JavaScript applications**. Test sending and receiving messages on behalf of Ethereum wallets via the [XMTP Labs](https://xmtp.com) development network in your own app. For a complete demonstration, see the [example React app](https://github.com/xmtp/example-chat-react). For key concepts and answers to frequently asked questions, see <https://docs.xmtp.org/>.
+**XMTP client SDK for JavaScript applications**
 
-## 🚧 **XMTP-JS is in active development** 🚧
+`xmtp-js` provides a TypeScript implementation of the XMTP client protocol for use with JavaScript and React applications.
 
-> ![Security](https://img.shields.io/badge/security-unaudited-orange) ![Stability](https://img.shields.io/badge/code%20stability-low-orange) ![Message Retention](https://img.shields.io/badge/message%20retention-7%20days-orange)
->
-> This pre-stable development release is publicly available for evaluation, feedback, and community contribution. All wallets and messages are forcibly deleted from the development network on Mondays.
->
-> - **DO NOT** use this package version in production.
-> - **DO NOT** share sensitive information via the development network.
-> - **DO** expect significant, frequent breaking revisions.
-> - **DO** contribute issues and PRs in this repo. The core team has limited bandwidth and may need a few days to review.
+Build with `xmtp-js` to provide messaging between blockchain wallet addresses, delivering on use cases such as wallet-to-wallet messaging and dapp-to-wallet notifications.
+
+For a demonstration of the core concepts and capabilities of the `xmtp-js` client SDK, see the [example React app](https://github.com/xmtp/example-chat-react).
+
+`xmtp-js` has not undergone a formal security audit.
+
+To learn more about XMTP and get answers to frequently asked questions, see <https://docs.xmtp.org/>.
+
+## 🏗 **Breaking revisions**
+
+Because `xmtp-js` is in active development, you should expect breaking revisions that might require you to adopt the latest SDK release to enable your app to continue working as expected.
+
+XMTP communicates about breaking revisions in the XMTP Discord community ([request access](https://xmtp.typeform.com/to/yojTJarb?utm_source=docs_home)), providing as much advance notice as possible. Additionally, breaking revisions in an `xmtp-js` release are described on the [Releases page](https://github.com/xmtp/xmtp-js/releases).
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -254,6 +254,22 @@ conversation.send('#'.repeat(1000), {
 })
 ```
 
+#### Manually handling private key storage
+
+The SDK will handle key storage for the user by encrypting the private key bundle using a signature generated from the wallet, and storing the encrypted payload it on the XMTP network. This can be awkward for some server-side applications, where you may only want to give the application access to the XMTP keys but not your wallet keys. Mobile applications may also want to store keys in a secure enclave rather than rely on decrypting the remote keys on the network each time the application starts up.
```
jazzz: > storing the encrypted payload it on the XMTP network.

Perhaps an extra 'it' here

src/crypto/PrivateKeyBundle.ts
```diff
@@ -112,7 +112,7 @@ export default class PrivateKeyBundle implements proto.PrivateKeyBundleV1 {
   }
 
   // encrypts/serializes the bundle for storage
-  async encode(wallet: ethers.Signer): Promise<Uint8Array> {
+  async toEncryptedBytes(wallet: ethers.Signer): Promise<Uint8Array> {
```
jazzz, line 115: 👍 glad to see this changed to something more accurate

README.md
mkobetic, line 259:
```suggestion
The SDK will handle key storage for the user by encrypting the private key bundle using a signature generated from the wallet, and storing the encrypted payload on the XMTP network. This can be awkward for some server-side applications, where you may only want to give the application access to the XMTP keys but not your wallet keys. Mobile applications may also want to store keys in a secure enclave rather than rely on decrypting the remote keys on the network each time the application starts up.
```

README.md
```diff
@@ -254,6 +254,22 @@ conversation.send('#'.repeat(1000), {
...
+
+You can export the key bundle using the static method `Client.getKeys`, save them somewhere secure, and then provide those keys at a later time to initialize a new client using the exported XMTP identity.
```
mkobetic, line 261:
```suggestion
You can export the unencrypted key bundle using the static method `Client.getKeys`, save it somewhere secure, and then provide those keys at a later time to initialize a new client using the exported XMTP identity.
```

mkobetic (approved):

src/Client.ts
```diff
@@ -114,6 +115,7 @@ export type ClientOptions = NetworkOptions & KeyStoreOptions & ContentOptions
 export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
   const _defaultOptions: ClientOptions = {
     keyStoreType: KeyStoreType.networkTopicStoreV1,
+    privateKeyOverride: undefined,
```
jazzz, line 118: [PEBBLE] Long term I'm not in love with this pattern.

1) It does not provide guidance on how these keys ought to be protected. 
2) It does not provide a mechanism to safe guard these keys in the future.

What are your thoughts on repurposing the KeyStore object to better suit developers needs? The idea being that we could provide secure storage mechanisms that satisfy 90% of the use cases and facilitate developers building their own if needs be. 

If a "OfflineBroswerKeyStore" was available in the SDK which did most of the heavy lifting, there would be significantly less burden on developers. We would then be able to push out security patches and updates. 

Thoughts?

neekolas: Can you elaborate on what exactly the `OfflineBrowserKeyStore` would do?

neekolas: I don't think the two are mutually exclusive. But we should have some escape hatch for developers whose needs we can't currently serve with a bespoke approach. There are a lot of different use-cases here (server, mobile, extension, etc, etc)

neekolas: There's maybe a secondary question about whether we should express this as some sort of "StaticStore" instead of just a client option. I wouldn't be opposed to that, although the API doesn't quite line up with the other providers since it would always resolve to a single wallet and couldn't accept writes.

jazzz: I very much agree that we will need an escape hatch, but I also hope it is rarely used ( and if it is, then it can be shared as a package so others can benefit from it as well). Instead major use cases are covered by existing keyStores, and developers are incentivized to use those instead. 

Using a keyStore interface for supplying keys would help keep some consistency across the different approaches, while also enforcing a common interface.  Symmetry is the focus here.

The StaticStore, is exactly how I would think about the lowest level escape hatch. Developers could take full control, if really needed.

jazzz: I'm cognizant that this would require a fair amount of work to revamp our flow and define an interface, so I'm not putting a block on anything. But as I look forward I see the need for clients to handle keyRotations and ratcheting, which will require more than just passing in a private key.

jazzz (approved):

src/Client.ts
neekolas: I reformatted this PR this to use a `StaticKeyStore`. It doesn't change much in the present, but puts us on better footing to make the system more full-featured down the line.

Couldn't agree more that I hope the escape hatch is very rarely used.

PR #135, opened 2022-07-21
fix: import node-localstorage in tests only

killthebuddh4:
*Context/Motivation behind change*:

* Deployed a Next.js app to Vercel
* Every API call threw a weird runtime error: 
image
* Ultimately I tracked down the error to the `import 'node-localstorage/register` in `LocalStorageStore.ts`
* Admittedly I didn't investigate much further, mostly because of the "this line is for tests" comment in `LocalStorageStore.ts` that suggested a potentially straightforward solution. My guess is that something about the serverless runtime Vercel uses breaks `node-localstorage`, but I'm not sure.

I wasn't sure whether to create an issue or open a conversation on Discord, figured I'd throw up this PR in case the solution I implemented makes sense. I ran tests locally before and after making these changes, all tests pass.

### DIFF
--- a/src/store/LocalStorageStore.ts
+++ b/src/store/LocalStorageStore.ts
@@ -1,6 +1,3 @@
-// This will create a global localStorage object on Node.js for use in tests
-// If we want to save some bytes from the bundle, we can have Webpack replace this with an empty module for the browser
-import 'node-localstorage/register'
 import { Store } from './Store'
 
 const KEY_PREFIX = '/xmtp/'
--- a/test/Client.test.ts
+++ b/test/Client.test.ts
@@ -1,3 +1,7 @@
+// This will create a global localStorage object on Node.js for use in tests
+// If we want to save some bytes from the bundle, we can have Webpack replace
+// this with an empty module for the browser
+import 'node-localstorage/register'
 import assert from 'assert'
 import {
   pollFor,
--- a/test/store/store.test.ts
+++ b/test/store/store.test.ts
@@ -1,3 +1,7 @@
+// This will create a global localStorage object on Node.js for use in tests
+// If we want to save some bytes from the bundle, we can have Webpack replace
+// this with an empty module for the browser
+import 'node-localstorage/register'
 import { EncryptedStore, LocalStorageStore } from '../../src/store'
 import assert from 'assert'
 import { WakuMessage } from 'js-waku'


neekolas (commented):
@snormore do you remember why we didn't do this initially? IIRC we did for a minute but had to change it.

neekolas:
Maybe usage in Node apps?

snormore:
@neekolas not sure, but this seems reasonable to me

neekolas:
I do think this is going to break in Node.js apps, other than our test runner.

Maybe we can do some sort of conditional require. Something like
```
if (typeof localStorage === "undefined" || localStorage === null) {
   require('node-localstorage/register')
}
```

killthebuddh4:
@neekolas If we'd like to keep the import inside `LocalStorageStore.ts` (which of course makes sense if it's required outside the tests) I think I should look a little bit more into the root cause of the error. I think the various possible root causes might influence the solution.

killthebuddh4:
@neekolas So I looked into the error a bit more. In the Vercel serverless runtime, the filesystem is set to read-only so it's not possible to use `node-localstorage/register`. I guess it might be "possible" but would require non-standard config since serverless functions are by convention stateless. I think it will be a fairly common case where the default runtime configuration does not allow filesystem access, in particular in many "serverless" environments. Do you think it would make sense to make the local storage store an opt-in config?

neekolas:
@killthebuddh4 ahh. Serverless runtimes aren't something we've been focused on.

Generally in serverless runtimes the `/tmp` folder is available for writes. Maybe we can allow an override where you can specify the storage directory so that in a serverless environment you can make use of the tmp folder. Something like

```ts
if (typeof localStorage === "undefined" || localStorage === null) {
  var LocalStorage = require('node-localstorage').LocalStorage;
  var localStorage = new LocalStorage( path.join(process.env.LOCALSTORAGE_DIR || '.' , 'localstorage') );
}
```

killthebuddh4:
> @killthebuddh4 ahh. Serverless runtimes aren't something we've been focused on.
> 
> Generally in serverless runtimes the `/tmp` folder is available for writes. Maybe we can allow an override where you can specify the storage directory so that in a serverless environment you can make use of the tmp folder. Something like
> 
> ```ts
> if (typeof localStorage === "undefined" || localStorage === null) {
>   var LocalStorage = require('node-localstorage').LocalStorage;
>   var localStorage = new LocalStorage( path.join(process.env.LOCALSTORAGE_DIR || '.' , 'localstorage') );
> }
> ```

I tested and this seems to work on Vercel at least, so solves my immediate problem and I'm happy with it. I'll update this PR soon with a solution. Right now we're running a forked version of `xmtp-js`, so it's fairly high priority for me.

mkobetic:
I'm not sure which PR, but we ripped localstorage out recently, correct @neekolas? Is this obsolete now?

neekolas:
Ya. It was removed as part of V7

mkobetic:
@killthebuddh4 ok to close this?

killthebuddh4:
> 

Oh boy, I can't believe this is still open. Thank you! Closed.

PR #137, opened 2022-07-23
docs: update docs url

saulmc:
Docs were recently moved from docs.xmtp.org to xmtp.org/docs.


jhaaaa (approved):
Thank you so much for this, Saul! 👍

PR #140, opened 2022-07-27
[WIP] Add stream for all messages

elisealix22:
This PR is a work in progress for adding a stream for all messages across all conversations and should not be merged as-is.

Currently this is blocked on unsubscribing from a list of content topics and then subscribing with a new list of content topics. To keep it simple, I set up the content topics list to stay the same between the unsubscribe and the second subscribe for the time being. The unsubscribe appears to work ok, but subscribing again with the same list of content topics causes malformed messages to return.

To reproduce:
1. Check out `ea/test-all-messages` branch in the example react app which [integrates with the new `streamAllMessages()`](https://github.com/xmtp/example-chat-react/compare/ea/test-all-messages) method: https://github.com/xmtp/example-chat-react/tree/ea/test-all-messages
2. Rebuild this library with `npm run build` and then run it
3. Open 2 chat windows with separate wallets and send send a message from one wallet to the other.
4. You should see a new conversation trigger the `resubscribeToTopics` method which is causing the error below for me currently:

Screen Shot 2022-07-27 at 3 27 49 PM

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -500,9 +500,8 @@ export default class Client {
       for (const connections of this.waku.libp2p.connectionManager.connections.values()) {
         for (const connection of connections) {
           if (!connection.streams.length) {
-            console.log('### Shut it down in the client')
             console.log(`Closing connection to ${connection.remoteAddr}`)
-            // connectionsToClose.push(connection.close())
+            connectionsToClose.push(connection.close())
           }
         }
       }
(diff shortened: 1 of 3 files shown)


neekolas:
I can't see how this code would lead to the problem you're having, so it feels like it must be on the server.

elisealix22:
Closing this because we found the source of the issue: https://github.com/xmtp/go-waku/pull/32. Will re-open when it's g2g.

PR #141, opened 2022-07-29
Add new API Client

neekolas:
## Summary
I realized there is a piece of the GRPC work that I can break out and merge in now. Should help cut down the size of the already-huge PR to replace Waku.

- Creates new API client wrapper for the MessageApi service
- Adds two new methods for using the `Query` API. `query` and `queryIterator`
- Adds tests for the API using `jest.mock` to mock out the backend
- Handles conversion of Date objects to `uint64`s in the expected format of GRPC Gateway
- Adds retries to all API methods

## Notes
Because this client uses entirely mocked API calls for the tests, and is not imported anywhere, it is safe to merge in advance of the node software being deployed

### DIFF
--- a/src/ApiClient.ts
+++ b/src/ApiClient.ts
@@ -1,0 +1,204 @@
+import {
+  Cursor,
+  Envelope,
+  MessageApi,
+  PagingInfo,
+  PublishRequest,
+  QueryRequest,
+  SortDirection,
+  SubscribeRequest,
+} from '@xmtp/proto'
+import { retry } from './utils'
+import { NotifyStreamEntityArrival } from '@xmtp/proto/ts/dist/types/fetch.pb'
+
+const RETRY_SLEEP_TIME = 100
+
+export type QueryParams = {
+  startTime?: Date
+  endTime?: Date
+  contentTopics: string[]
+}
+
+export type QueryAllOptions = {
+  direction?: SortDirection
+  limit?: number
+}
+
+export type QueryStreamOptions = Omit<QueryAllOptions, 'limit'> & {
+  pageSize?: number
+}
+
+export type PublishParams = {
+  contentTopic: string
+  message: Uint8Array
+  timestamp?: Date
+}
+
+export type SubscribeParams = {
+  contentTopics: string[]
+}
+
+export type ApiClientOptions = {
+  maxRetries?: number
+}
+
+export type SubscribeCallback = NotifyStreamEntityArrival<Envelope>
+
+const toNanoString = (d: Date | undefined): undefined | string => {
+  return d && (d.valueOf() * 1_000_000).toFixed(0)
+}
+
+export default class ApiClient {
+  pathPrefix: string
+  maxRetries: number
+
+  constructor(pathPrefix: string, opts?: ApiClientOptions) {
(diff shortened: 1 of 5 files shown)


src/ApiClient.ts
```diff
@@ -0,0 +1,204 @@
...
+  contentTopics: string[]
+}
+
+export type QueryAllOptions = {
+  direction?: SortDirection
+  limit?: number
+}
+
+export type QueryStreamOptions = Omit<QueryAllOptions, 'limit'> & {
+  pageSize?: number
```
snormore, line 28: Should we just call this `limit` too?

src/ApiClient.ts
```diff
@@ -0,0 +1,204 @@
...
+          break
+        }
+      }
+    }
+    return out
+  }
+
+  // Will produce an AsyncGenerator of Envelopes
+  // Uses queryStreamPages under the hood
+  async *queryStream(
```
snormore, line 119: What does the `*` prefix do on these?

src/utils.ts
```diff
@@ -48,6 +48,29 @@ export const promiseWithTimeout = <T>(
...
+  maxRetries: number,
+  sleepTime: number,
+  retryCount = 1
+): Promise<Awaited<ReturnType<T>>> {
+  const currRetry = typeof retryCount === 'number' ? retryCount : 1
+  try {
+    const result = await fn(...args)
+    return result
+  } catch (e) {
+    console.error(e)
```
snormore, line 65: Should we just `log` or `debug` here instead?

snormore (commented):
Nice

src/ApiClient.ts
elisealix22, line 119: Just for my own knowledge: Is `QueryStream` a standard term? Asking because at first glance I'd expect a "stream" to work similarly to `subscribe()`. Wondering if maybe this could just be `query()` and `queryPages()`?

elisealix22 (commented):
Overall looks fantastic to me! Just one question around naming the APIs

src/ApiClient.ts
neekolas: You know, maybe the term is confusing.

I could go for `query` and `queryIterator`?

neekolas: It defines the function as a generator, which allows me to use the `yield` keyword. They work a lot like Python generators. If you're curious to learn more, I found [this resource](https://javascript.info/async-iterators-generators) helpful when building.

src/utils.ts
neekolas: I can switch to a log

src/ApiClient.ts
neekolas: I wanted to make it a little more explicit that you would still receive 100% of the items that match the query in this case.

src/ApiClient.ts
neekolas: I renamed

neekolas:
Addressed all feedback, including renaming the methods. I like `query` and `queryIterator` more than the original. Thanks for the push @elisealix22

elisealix22 (approved):
🚀

Issue #92, opened 2022-03-23
Listen for new messages in all conversations


saulmc:
## Problem

Usually, developers need to listen for new messages in all of a user's conversations. The client is currently only capable of streaming messages from an individual conversation. Therefore, developers must aggregate conversation streams in their applications. To avoid duplicate effort, this capability should exist in the client.

### Requirements
- Stream incoming messages for all conversations that have started
- Stream incoming messages for new conversations that started since the stream was initiated

### Non-goals
- Provide control over whether inbound or outbound messages are returned

elisealix22:
Wanted to provide a status update on this! Last week @neekolas and I fixed a couple of issues in `go-waku` around unsubscribing to `Stream`s that were blocking the second requirement of streaming incoming messages for new conversations:
- https://github.com/xmtp/go-waku/pull/29
- https://github.com/xmtp/go-waku/pull/32

Now that those are unblocked, I restarted feature development today and went down the path of establishing two streams: one to listen for existing conversations' messages and one to listen for new conversations' messages. The idea was that we could detect a new conversation in one stream, unsubscribe the previous list of conversation topics in the other stream and then resubscribe appending the newly detected conversation address from the first stream. However, I'm fearing that this is actually quite complex because of having to merge the output of two async iterators.

Instead of pushing that further, I'm going back to the drawing board to try a new approach where we only have a single stream to listen for all new and existing conversations. The hope is that we can differentiate between the two when new messages come in so we can correctly unsubscribe / resubscribe in order to append a new dm topic when a new conversation is detected (to prevent missing more messages for that convo once it's been started).

I'll report back again tomorrow on if this a feasible approach!

elisealix22:
Was able to get an end-to-end single stream approach working for both new and existing conversations today! 🎉  I still have tests to write and some clean-up to do but will open up a PR tomorrow for sure.

elisealix22:
@KishanBagaria @emilio-im this is now available in [v5.3.0](https://github.com/xmtp/xmtp-js/releases/tag/v5.3.0)! Let me know if you run into any issues if you try it out.

PR #142, opened 2022-08-03
Introduce API to stream all messages for all conversations

elisealix22:
Closes https://github.com/xmtp/xmtp-js/issues/92

### Summary

This PR introduces a `Conversations#streamAllMessages()` API that will listen for new messages for both new and existing conversations.

The implementation is to subscribe to multiple content topics - any existing DM topics and the intro topic. The intro topic is to support the fresh account scenario and the new conversation scenario. When we detect a new conversation, we update the stream's content topics to include the new DM topic.

Note: In order to detect a new conversation, I save the `contentTopic` to the `Message` so I can tell if it is an intro topic when receiving it.

### Tested scenarios
- [x] New account without any conversations receives new messages
- [x] Existing account receives new messages for new conversations
- [x] Existing account receives new messages for existing conversations

### DIFF
--- a/src/Stream.ts
+++ b/src/Stream.ts
@@ -77,6 +77,7 @@ export default class Stream<T> {
       this.callback,
       this.topics
     )
+    console.log('Subscribe to topics in start!')
     await this.listenForDisconnect()
   }
 
@@ -85,12 +86,14 @@ export default class Stream<T> {
     // Save the callback function on the class so we can clean up later
     this._disconnectCallback = async (connection: Connection) => {
       if (connection.remotePeer.toB58String() === peer?.id?.toB58String()) {
+        console.log('### Disconnect callback called!')
         console.log(`Connection to peer ${connection.remoteAddr} lost`)
         while (true) {
           try {
             if (!this.callback) {
               return
             }
+            console.log('Subscribe to topics in disconnect callback!')
             this.unsubscribeFn = await this.client.waku.filter.subscribe(
               this.callback,
               this.topics
@@ -105,6 +108,7 @@ export default class Stream<T> {
       }
     }
 
+    console.log('### Init disconnect callback!')
     this.client.waku.libp2p.connectionManager.on(
       'peer:disconnect',
       this._disconnectCallback
@@ -113,11 +117,11 @@ export default class Stream<T> {
 
   static async create<T>(
     client: Client,
-    topic: string,
+    topics: string[],
     messageTransformer: MessageTransformer<T>,
     messageFilter?: MessageFilter
   ): Promise<Stream<T>> {
-    const stream = new Stream(client, topic, messageTransformer, messageFilter)
+    const stream = new Stream(client, topics, messageTransformer, messageFilter)
     await stream.start()
     return stream
   }
@@ -166,5 +170,15 @@ export default class Stream<T> {
     }
     // otherwise return empty Promise and queue its resolver
     return new Promise((resolve) => this.resolvers.unshift(resolve))
+  }
+
+  // Unsubscribe and resubscribe with new topics.
+  async resetTopics(topics: string[]): Promise<void> {
+    console.log(`New topics: ${topics}`)
+    this.topics = topics
+    if (this.unsubscribeFn) {
+      console.log('### Unsubscribe!')
+      await this.unsubscribeFn()
+    }
   }
 }
--- a/src/Stream.ts
+++ b/src/Stream.ts
@@ -51,6 +51,7 @@ export default class Stream<T> {
       if (!wakuMsg.payload) {
         return
       }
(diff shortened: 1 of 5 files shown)


src/Stream.ts
```diff
@@ -31,31 +37,47 @@ export default class Stream<T> {
...
+      )
       // If there is a filter on the stream, and the filter returns false, ignore the message
       if (filter && !filter(msg)) {
         return
       }
+      // Check to see if we should update the stream's content topic subscription
+      if (contentTopicUpdater) {
+        const newTopics = contentTopicUpdater(msg)
+        if (newTopics) {
+          this.resubscribeToTopics(newTopics)
```
mkobetic, line 78: `newTopics` is a bit confusing here. This is all the topics, right? The old ones and whatever new ones came in, correct?

neekolas (approved):
This looks great. I think the tests cover the tricky parts of the implementation.

src/Stream.ts
neekolas: Oh ya, I was also a little thrown off by the variable name

src/conversations/Conversations.ts
```diff
@@ -74,12 +78,76 @@ export default class Conversations {
...
+      }
+      const peerAddress = this.getPeerAddress(msg)
+      if (
+        dmAddresses.includes(peerAddress) ||
+        peerAddress === this.client.address
+      ) {
+        // No need to update if we're already subscribed
+        return undefined
+      }
+      dmAddresses.push(this.getPeerAddress(msg))
```
mkobetic, line 111:
```suggestion
      dmAddresses.push(peerAddress)
```

src/conversations/Conversations.ts
```diff
@@ -74,12 +78,76 @@ export default class Conversations {
...
+      return this.buildTopicsForAllMessages(dmAddresses, introTopic)
+    }
+
+    // Filter intro topics if already streaming direct messages for that address to avoid duplicates
+    const filter = (msg: Message): boolean => {
```
mkobetic, line 116: Can't we filter all introTopic messages? IIRC the intro message should be duplicated into the dm topic as well, no?

mkobetic (approved):
Some minor comments, but otherwise very nice. 👍

src/Stream.ts
elisealix22: Great point! Yes, it's all topics including the new topic so I'll update the name for sure

src/conversations/Conversations.ts
elisealix22: I don't think it's possible to catch new messages coming from new conversations because we aren't subscribed to that DM topic yet (since it didn't exist when the stream started). From my testing, if a fresh convo is kicked off, the intro topic is the only message received and then we subscribe to the DM topic after that. We only protect against dups here in case the convo existed before the intro topic was sent and we were already subscribing to the DM. Lmk if anything seems off with that though!

mkobetic: Ah, I see, got it now. 👍

mkobetic: That feels a bit racy though, since there may be more messages in the new convo topic. Prudent thing would probably be to list the new conversation after subscribing to make sure we didn't drop anything.

mkobetic: Then again, we might fetch some old messages if we list and old dm topic. Hm, not sure if we can actually do much better than you're already doing.

elisealix22: >That feels a bit racy though, since there may be more messages in the new convo topic.

I _think_ this should be covered by us creating and appending a new DM topic for the new address as we receive the intro message and including that in the new subscription (still within the new message callback for the intro message):

https://github.com/xmtp/xmtp-js/blob/0bf89e9081b2eb374b333e485fbc0cc01c75ef03/src/Stream.ts#L78

I could absolutely be missing something though so please lmk if I'm misunderstanding!

mkobetic: My point was that I don't think there's anything preventing the sender from sending multiple messages before the recipient subscribes to the new conversation (in fact could be even before the recipient receives the intro message). So in theory we could end up showing the very first message (from the intro topic), then miss a few that arrived to the dm topic before we subscribed to it, and then start showing whatever comes after. That's why I suggested to list the topic first. But if the two parties communicated before then it could pick up messages that shouldn't be part of this stream. So I'm not sure if we can make this really bullet proof.

mkobetic: I'd say don't let this stop you from merging this. We can ponder the tricky stuff later if it ever becomes relevant.

elisealix22: Ahhh got it. Appreciate the explanation @mkobetic 🙏 Will keep an eye out for feedback here for sure.

PR #143, opened 2022-08-03
Message grpc client integration

neekolas:
## Summary
- Completely removes Waku dependency from the `js-sdk` and replaces with the GRPC API client
- Uses the new GRPC API Client for all interactions with nodes
- Replaces the fleets file method of peer discovery with a fixed URL for each environment
- Implements new `authn` header protocol to allow for per wallet rate limiting 
- Many thanks to @snormore, who setup the foundation for this change and prototyped the initial version.

## Remaining items before this can be merged
- ~Need to ensure backend is deployed to all environments behind a load balancer with TLS, and test end-to-end in the chat app~
- ~I am having issues catching the `AbortError` caused by canceling streaming methods early. It bubbles up too high in the stack and cannot be managed with a try/catch.~
- ~Need to figure out a solution to `authn`~
- Ensure all documentation is accurate
- Review tests and make sure everything new is covered


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
neekolas, line 1: This was helpful for debugging some issues, so sharing with everyone.

package.json
```diff
@@ -2,10 +2,11 @@
...
   "version": "0.0.0-development",
   "description": "XMTP client SDK for interacting with XMTP networks.",
-  "main": "dist/cjs/src/index.js",
-  "module": "dist/esm/src/index.js",
-  "umd:main": "dist/umd/src/index.js",
-  "types": "dist/types/src/index.d.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "umd:main": "dist/umd/index.js",
+  "browser": "dist/umd/index.js",
```
neekolas, line 8: Added the `browser` field to help webpack find the UMD build

src/Client.ts
```diff
@@ -28,31 +23,30 @@ import {
...
+  production: 'https://production.xmtp.network',
+} as const
+
 // Parameters for the listMessages functions
 export type ListMessagesOptions = {
   checkAddresses?: boolean
-  pageSize?: number
   startTime?: Date
   endTime?: Date
+  limit?: number
```
neekolas, line 49: Breaking change

src/Client.ts
```diff
@@ -74,15 +68,10 @@ export type SendOptions = {
  * Network startup options
  */
 type NetworkOptions = {
-  /** List of multiaddrs for boot nodes */
-  bootstrapAddrs?: string[]
```
neekolas, line 78: Breaking change

src/Client.ts
```diff
@@ -74,15 +68,10 @@ export type SendOptions = {
...
   // Allow for specifying different envs later
-  env: keyof NodesList
-  /**
-   * How long we should wait for the initial peer connection
-   * to declare the startup as successful or failed
-   */
-  waitForPeersTimeoutMs: number
```
neekolas, line 85: Breaking change

src/Client.ts
```diff
@@ -174,9 +161,10 @@ export default class Client {
     opts?: Partial<ClientOptions>
   ): Promise<Client> {
     const options = defaultOptions(opts)
-    const waku = await createWaku(options)
-    const keys = await loadOrCreateKeysFromOptions(options, wallet, waku)
-    const client = new Client(waku, keys)
+    const apiClient = createApiClientFromOptions(options)
+    const keys = await loadOrCreateKeysFromOptions(options, wallet, apiClient)
+    apiClient.setAuthenticator(new Authenticator(keys.identityKey))
```
neekolas, line 166: I need the `apiClient` to get the keys from the network topic store, only once I have the keys can I add the authenticator. So, I have to go through this awkward flow of adding the authenticator later.

src/Client.ts
```diff
@@ -446,64 +434,47 @@ export default class Client {
...
+      }
     )
-    if (opts?.checkAddresses) {
+
+    let msgs: Message[] = []
+    for (const env of res) {
+      if (!env.message) continue
+      try {
+        const msg = await this.decodeMessage(
+          b64Decode(env.message as unknown as string),
```
neekolas, line 452: Have to coerce the types here since the generated types for `UInt8Array` are a lie on the return value

src/Client.ts
```diff
@@ -220,24 +206,22 @@ export default class Client {
...
-      },
-    })
+    const stream = this.apiClient.queryIterator(
+      { contentTopics: [buildUserContactTopic(peerAddress)] },
+      { pageSize: 5 }
+    )
+
+    for await (const env of stream) {
+      if (!env.message) continue
+      const bundle = ContactBundle.fromBytes(b64Decode(env.message.toString()))
```
mkobetic, line 216: I don't understand where does the `b64Decode` come from, are we changing the format of the contact bundles?

mkobetic: Oh, is that the JSON encoding between the grpc gateway and the client?

src/authn/AuthCache.ts
```diff
@@ -0,0 +1,31 @@
+import Authenticator from './Authenticator'
+import Token from './Token'
+
+// Default to 10 seconds less than expected expiry to give some wiggle room near the end
```
mkobetic, line 4: Could you add a link to the node side constant that should correlate with this? 🙏

src/authn/Token.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  static fromBytes(bytes: Uint8Array): Token {
+    return new Token(authn.Token.decode(bytes))
+  }
+
+  toBase64(): string {
+    return (
+      Buffer.from(this.toBytes())
+        // base64url is not widely supported in browsers so convert to base64 and then replace the non url-safe characters
```
mkobetic, line 49: Should we just go to straight b64 to simplify this? I'm not sure we need a URL safe version for the header.

mkobetic (approved):
This is large and I can't claim thorough understanding, but it looks good as far as I can wrap my head around it.

mkobetic:
Is there a good way to surface the breaking changes to devs somehow?

src/authn/AuthCache.ts
neekolas: That's a good idea

src/authn/Token.ts
neekolas: Ya. I would think all base64 characters would be safe in the header across any reasonable implementation. I'm OK to switch if you are.

neekolas: Sounds great. I'll adjust mine

PR #145, opened 2022-08-08
Return message upon sending

neekolas:
## Summary
Through conversations with developers we realized that there is demand for returning the sent message from the `sendMessage` function. This can be used for immediately updating the UI before confirmation.

This PR adds a return value to `sendMessage`. This should be a non-breaking change as the API previously returned `void`.

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -312,16 +312,18 @@ export default class Client {
     }
     const timestamp = options?.timestamp || new Date()
     const msg = await this.encodeMessage(recipient, timestamp, content, options)
+    const msgBytes = msg.toBytes()
+
     await Promise.all(
       topics.map(async (topic) => {
-        const wakuMsg = await WakuMessage.fromBytes(msg.toBytes(), topic, {
+        const wakuMsg = await WakuMessage.fromBytes(msgBytes, topic, {
           timestamp,
         })
         return this.sendWakuMessage(wakuMsg)
       })
     )
 
-    return msg
+    return this.decodeMessage(msgBytes, topics[0])
   }
 
   private async sendWakuMessage(msg: WakuMessage): Promise<void> {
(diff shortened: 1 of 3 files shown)


test/Client.test.ts
```diff
@@ -94,7 +94,8 @@ describe('Client', () => {
         const aliceBob = await alice.streamConversationMessages(bob.address)
 
         // alice sends intro
-        await alice.sendMessage(bob.address, 'hi bob!')
+        const sentMessage = await alice.sendMessage(bob.address, 'hi bob!')
+        assert.equal(sentMessage.content, 'hi bob!')
```
mkobetic, line 98: should we check that the Id was filled in, or whatever was the thing the devs were looking for?

neekolas: Sure. I added some more assertions just to make sure

mkobetic (approved):

src/Client.ts
```diff
@@ -312,14 +312,18 @@ export default class Client {
...
       topics.map(async (topic) => {
-        const wakuMsg = await WakuMessage.fromBytes(msg.toBytes(), topic, {
+        const wakuMsg = await WakuMessage.fromBytes(msgBytes, topic, {
           timestamp,
         })
         return this.sendWakuMessage(wakuMsg)
       })
     )
+
+    return this.decodeMessage(msgBytes, topics[0])
```
neekolas, line 326: I wish I didn't have to do this, but we only include things like the `contentType` and the decrypted content in the `decodeMessage` function. The Message only stores the resulting bytes. I'd rather not copy all the code for deriving that into a second place for maintainability's sake.

We should look at refactoring the message encoding/decoding so that this logic lives inside the message instead of the client.

mkobetic: Uff, I really dislike this. That's a lot of overhead for something that may or may not end up being used.

PR #146, opened 2022-08-11
Create Authn Tools

neekolas:
## Summary

Decided to break out some of the `authn` work and make a PR against my working GRPC branch. Figure this deserves a thorough review on its own.

- Creates an `Authenticator` class responsible for generating tokens based on a `PrivateKey`
- Creates some helper classes to help with serialization of the generated protos
- Uncomments a bunch of commented out `authn` code


src/authn/Token.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export default class Token implements authn.Token {
+  identityKey: authn.PublicKey
+  authDataBytes: Uint8Array
+  authDataSignature: authn.Signature
+  private _authData?: AuthData
+
+  constructor({ identityKey, authDataBytes, authDataSignature }: authn.Token) {
+    if (!identityKey) {
+      throw new Error('Missing identity key in token')
```
mkobetic, line 12: If the key signature is optional, we should check that it is indeed signed as well.

test/authn/Authn.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    )
+    expect(token.authData.createdNs).toEqual(+timestamp * 1000)
+    expect(token.identityKey.timestamp).toEqual(privateKey.publicKey.timestamp)
+    expect(token.identityKey.signature).toEqual(privateKey.publicKey.signature)
+    expect(token.identityKey.secp256k1Uncompressed).toEqual(
+      privateKey.publicKey.secp256k1Uncompressed
+    )
+  })
+
+  it('rejects unsigned identity keys', async () => {
```
mkobetic, line 36: Interesting, you already have a test for the key signature, not sure how that works.

mkobetic:
Very nice!

PR #147, opened 2022-08-11
Prepare for GRPC Upgrade

neekolas:
## Summary
- Uses the `@xmtp/proto` package for all protobufs
- Replaces all references to local protobufs
- Removes protobuf generation code and `buf`
- Temporarily removes `authn` service from the client (will come back reborn with GRPC)
- Updates to latest version of `ApiClient` from GRPC branch


mkobetic (approved):
Looks good.

src/ApiClient.ts
```diff
@@ -58,36 +66,63 @@ export default class ApiClient {
...
+
+    const doSubscribe = () => {
+      abortController = new AbortController()
+      const startTime = +new Date()
+
+      MessageApi.Subscribe(req, cb, {
+        pathPrefix: this.pathPrefix,
+        signal: abortController.signal,
+        mode: 'cors',
+      }).catch(async (err: any) => {
```
elisealix22, line 107: This warning seems valid possibly?

elisealix22 (approved):
Beautiful 🥲

src/ApiClient.ts
neekolas: The reason I never re-throw the error here is because nothing is able to await this promise (since it would block until the stream is done). If I throw, it would lead to an `UnhandledPromiseRejection` error, which is can crash the whole app. So, I just log the error and swallow it.

PR #148, opened 2022-08-12
fix: createToken wallet address

mkobetic:
I tried to use the branch to generate a token sample to bring over to xmtp-node-go to test compatibility. The wallet address wasn't matching so that's how I noticed this.

Also adding `toBase64()` since that's how I'm assuming we'll be embedding the token in the request header. Not sure if this is the right way to do this.

### DIFF
--- a/src/authn/AuthData.ts
+++ b/src/authn/AuthData.ts
@@ -13,7 +13,7 @@ export default class AuthData implements authnProto.AuthData {
     timestamp = timestamp || new Date()
     return new AuthData({
       walletAddr: walletAddr,
-      createdNs: timestamp.getTime() * 1000,
+      createdNs: timestamp.getTime() * 1000000,
     })
   }
 
--- a/src/authn/Authenticator.ts
+++ b/src/authn/Authenticator.ts
@@ -17,7 +17,7 @@ export default class Authenticator {
 
   async createToken(timestamp?: Date): Promise<Token> {
     const authData = AuthData.create(
-      this.identityKey.publicKey.getEthereumAddress(),
+      this.identityKey.publicKey.walletSignatureAddress(),
       timestamp || new Date()
     )
     const authDataBytes = authData.toBytes()
--- a/src/authn/Token.ts
+++ b/src/authn/Token.ts
@@ -35,4 +35,8 @@ export default class Token implements authn.Token {
   static fromBytes(bytes: Uint8Array): Token {
     return new Token(authn.Token.decode(bytes))
   }
+
+  toBase64(): string {
+    return Buffer.from(this.toBytes()).toString('base64url')
+  }
 }
--- a/test/authn/Authn.test.ts
+++ b/test/authn/Authn.test.ts
@@ -23,7 +23,7 @@ describe('authn', () => {
     const token = await authenticator.createToken(timestamp)
 
     expect(token.authData.walletAddr).toEqual(
-      privateKey.publicKey.getEthereumAddress()
+      privateKey.publicKey.walletSignatureAddress()
     )
     expect(token.authData.createdNs).toEqual(+timestamp * 1000)
     expect(token.identityKey.timestamp).toEqual(privateKey.publicKey.timestamp)
--- a/test/authn/Authn.test.ts
+++ b/test/authn/Authn.test.ts
@@ -43,7 +43,8 @@ describe('authn', () => {
 
   it('round trips safely', async () => {
     const originalToken = await authenticator.createToken()
-    const newToken = Token.fromBytes(originalToken.toBytes())
+    const bytes = originalToken.toBytes()
+    const newToken = Token.fromBytes(bytes)
     expect(originalToken.authData).toEqual(newToken.authData)
     expect(originalToken.toBytes()).toEqual(newToken.toBytes())
   })
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@noble/secp256k1": "^1.5.2",
         "@stardazed/streams-polyfill": "^2.4.0",
         "@xmtp/proto": "^1.2.0",
+        "big-integer": "^1.6.51",
         "ethers": "^5.5.3",
         "node-localstorage": "^2.2.1"
       },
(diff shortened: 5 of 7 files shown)


src/authn/Authenticator.ts
```diff
@@ -17,7 +17,7 @@ export default class Authenticator {
 
   async createToken(timestamp?: Date): Promise<Token> {
     const authData = AuthData.create(
-      this.identityKey.publicKey.getEthereumAddress(),
+      this.identityKey.publicKey.walletSignatureAddress(),
```
mkobetic, line 20: Not the right key to generate the address from.

src/authn/Token.ts
```diff
@@ -35,4 +35,8 @@ export default class Token implements authn.Token {
   static fromBytes(bytes: Uint8Array): Token {
     return new Token(authn.Token.decode(bytes))
   }
+
+  toBase64(): string {
+    return Buffer.from(this.toBytes()).toString('base64url')
```
mkobetic, line 40: I'm not sure which version of base64 we want to use, this is what I'm using at the moment on the node side.

test/authn/Authn.test.ts
```diff
@@ -43,7 +43,8 @@ describe('authn', () => {
 
   it('round trips safely', async () => {
     const originalToken = await authenticator.createToken()
-    const newToken = Token.fromBytes(originalToken.toBytes())
+    const bytes = originalToken.toBytes()
```
mkobetic, line 46: I've used this to capture the bytes, but it's not really necessary, now that I've learned how to use the VC's debug console :).

src/authn/Authenticator.ts
neekolas: Yikes. Thanks

neekolas (approved):

src/authn/AuthData.ts
```diff
@@ -13,7 +13,7 @@ export default class AuthData implements authnProto.AuthData {
     timestamp = timestamp || new Date()
     return new AuthData({
       walletAddr: walletAddr,
-      createdNs: timestamp.getTime() * 1000,
+      createdNs: timestamp.getTime() * 1000000,
```
mkobetic, line 16: 1 more fix. The timestamp should be nano not micro. But this gives me this error in tests which I'm not sure how to deal with.
```
● authn › can create a token

    Value is larger than Number.MAX_SAFE_INTEGER

      19 |
      20 |   static fromBytes(bytes: Uint8Array): AuthData {
    > 21 |     const res = authnProto.AuthData.decode(bytes)
         |                                     ^
      22 |     return new AuthData(res)
      23 |   }
      24 |

      at longToNumber (node_modules/@xmtp/proto/ts/message_api/v1/authn.pb.ts:261:11)
      at Object.decode (node_modules/@xmtp/proto/ts/message_api/v1/authn.pb.ts:164:31)
      at Function.fromBytes (src/authn/AuthData.ts:21:37)
      at Token.get authData [as authData] (src/authn/Token.ts:25:33)
      at test/authn/Authn.test.ts:25:18
      at fulfilled (test/authn/Authn.test.ts:5:58)

  ● authn › round trips safely

    Value is larger than Number.MAX_SAFE_INTEGER

      19 |
      20 |   static fromBytes(bytes: Uint8Array): AuthData {
    > 21 |     const res = authnProto.AuthData.decode(bytes)
         |                                     ^
      22 |     return new AuthData(res)
      23 |   }
      24 |

      at longToNumber (node_modules/@xmtp/proto/ts/message_api/v1/authn.pb.ts:261:11)
      at Object.decode (node_modules/@xmtp/proto/ts/message_api/v1/authn.pb.ts:164:31)
      at Function.fromBytes (src/authn/AuthData.ts:21:37)
      at Token.get authData [as authData] (src/authn/Token.ts:25:33)
      at test/authn/Authn.test.ts:48:26
      at fulfilled (test/authn/Authn.test.ts:5:58)
```

mkobetic: @neekolas can I let you deal with this one and finish this up? 🙏  No rush, this doesn't block me, I've got my token sample.

src/authn/Token.ts
neekolas: The GRPC Gateway client implementation uses a `b64Encode` function that looks like this. I was considering using the same, since there must be an equivalent deserialization function available somewhere in the gateway code.

It looks like this:
```ts
// Base64 encoding table
const b64 = new Array(64);

// Base64 decoding table
const s64 = new Array(123);

// 65..90, 97..122, 48..57, 43, 47
for (let i = 0; i < 64;)
    s64[b64[i] = i < 26 ? i + 65 : i < 52 ? i + 71 : i < 62 ? i - 4 : i - 59 | 43] = i++;

export function b64Encode(buffer: Uint8Array, start: number, end: number): string {
	let parts: string[] = null;
  const chunk = [];
  let i = 0, // output index
    j = 0, // goto index
    t;     // temporary
  while (start < end) {
    const b = buffer[start++];
    switch (j) {
      case 0:
        chunk[i++] = b64[b >> 2];
        t = (b & 3) << 4;
        j = 1;
        break;
      case 1:
        chunk[i++] = b64[t | b >> 4];
        t = (b & 15) << 2;
        j = 2;
        break;
      case 2:
        chunk[i++] = b64[t | b >> 6];
        chunk[i++] = b64[b & 63];
        j = 0;
        break;
    }
    if (i > 8191) {
      (parts || (parts = [])).push(String.fromCharCode.apply(String, chunk));
      i = 0;
    }
  }
  if (j) {
    chunk[i++] = b64[t];
    chunk[i++] = 61;
    if (j === 1)
      chunk[i++] = 61;
  }
  if (parts) {
    if (i)
      parts.push(String.fromCharCode.apply(String, chunk.slice(0, i)));
    return parts.join("");
  }
  return String.fromCharCode.apply(String, chunk.slice(0, i));
}
```

mkobetic: Not quite sure, is that standard b64? It doesn't really matter much which it is I think, http headers should be able to accommodate most of them, no? The easiest for me on the node side are any of these https://pkg.go.dev/encoding/base64#pkg-variables

mkobetic: Doesn't actually have to be b64 either if you have other suggestions.

src/authn/AuthData.ts
neekolas: No problem. I've switched to using a `BigInt` for the intermediate number, which is natively supported in browsers and polyfillable everywhere else.

mkobetic: Won't be an issue in RN?

neekolas: Not this way. I've explicitly imported the polyfill, so it'll work out of the box.

src/authn/Token.ts
neekolas: I'll leave it as-is. I think all roads lead to the same place, given the fact that the input is a `Uint8Array` and won't contain any special characters.

PR #153, opened 2022-08-24
Pass a `Headers` object in our publish requests

elisealix22:
This resolves an issue with React Native's `fetch` API specific to Android where before it would error with `TypeError: Network request failed`. Luckily, I found [someone else who ran into this](https://stackoverflow.com/a/55238033) and passing in a new `Headers` object in instead of the json fixes the issue. I confirmed it still works on iOS as well.

### DIFF
--- a/src/ApiClient.ts
+++ b/src/ApiClient.ts
@@ -109,7 +109,7 @@ export default class ApiClient {
           {
             pathPrefix: this.pathPrefix,
             mode: 'cors',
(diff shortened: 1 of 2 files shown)


neekolas:
Not sure how the commitlint didn't catch it, but you'll want to prefix your commit message with "fix:" to trigger a release after this is merged.

neekolas (approved):

PR #155, opened 2022-09-06
Upgrade to V3 of XMTP Proto

neekolas:
## Summary

This is the minimum viable implementation of the V3 version of the proto library. It doesn't add support for any of the new types, but does update all references to pull in the relevant existing types.

This change should introduce no changes to the wire format of messages or change any behaviour in the client.


src/crypto/Signature.ts
```diff
@@ -1,13 +1,14 @@
...
 
 // Signature represents an ECDSA signature with recovery bit.
-export default class Signature implements proto.Signature {
-  ecdsaCompact: proto.Signature_ECDSACompact | undefined // eslint-disable-line camelcase
+export default class Signature implements signature.Signature {
+  ecdsaCompact: signature.Signature_ECDSACompact | undefined // eslint-disable-line camelcase
+  walletEcdsaCompact: signature.Signature_WalletECDSACompact | undefined // eslint-disable-line camelcase
 
-  constructor(obj: proto.Signature) {
+  constructor(obj: Partial<signature.Signature>) {
```
mkobetic, line 11: Why the Partial?

mkobetic (approved):

src/crypto/Signature.ts
neekolas: Great question. Without the `Partial` it expects both the `ecdsaCompact` and the `walletEcdsaCompact`. The alternative is to make any caller always provide both with one undefined.

neekolas: Ya, it's one of Typescript's weird quirks. https://www.becomebetterprogrammer.com/typescript-question-mark/

PR #158, opened 2022-09-08
docs: Edit a few links to enable single-sourcing of README.md

jhaaaa:
- Change a relative link to a full URL in README.md so that we can single-source this great quickstart content to the `xmtp-dot-org` repo and publish it to the Dev Portal docs. (Relative links break in the context of the `xmtp-dot-org` repo.)
- Change a link to the docs to make for a more natural transition when displayed directly in the `xmtp-dot-org` Docs context
- Update links to the XMTP Discord community, which no longer requires an initial access request

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,29 +8,29 @@
 
 **XMTP client SDK for JavaScript applications**
 
-`xmtp-js` provides a TypeScript implementation of the XMTP client protocol for use with JavaScript and React applications.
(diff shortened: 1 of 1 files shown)


saulmc (approved):
Nice catch on those early access links. I've been wondering how folks are still finding that form.

You might need to prepend the commit message with `docs:` for semantic versioning.

PR #159, opened 2022-09-09
Signed keys and wallet signature

mkobetic:
Implements https://github.com/xmtp-labs/hq/issues/678

This is the first step towards implementing xmtp/proto 3.0.0, as most of the other changes depend on the signed key types. In the process I realized that it wasn't possible to accommodate the new signed public key type without introducing a new private key type as well, thus the [v3.1.0](https://github.com/xmtp/proto/pull/16). I tried a less disruptive route of keeping the new key types for marshalled format only, but it proved impossible as information is lost when I tried to roll everything into the pre-existing key types.

This PR
* adds new key classes and marks the pre-existing ones as legacy, i.e.
  * `PublicKey`  =>  `UnsignedPublicKey` & `SignedPublicKey` (note that `PublicKey` now extends `UnsignedPublicKey`)
  * `PrivateKey` => `SignedPrivateKey` (it didn't seem worth it to extract shared bits to a common superclass, there wasn't much to share, the type signatures were off)
* Adds some private object types to replace direct use of the proto types that force undefined checks all over the place, makes the code tidier (`secp256k1`, `secp256k1Uncompressed`, `ECDSACompactWithRecovery`)
* New key types are generated with `SignedPrivateKey.generate(signer: KeySigner)` as the signed keys must be created signed from the beginning. `WalletSigner` should be used to sign a key with a wallet. This deprecates `PublicKey.signWithWallet()`
* `SignedPublicKey.walletSignatureAddress()` now throws for keys that weren't signed by a wallet since we now know if that's the case. `SignedPublicKey.signerKey()` is a new more generic version of it that works for pre-keys as well.
* All the special handling of wallet signatures is now in `WalletSigner` (with some left overs in PublicKey as I didn't want to increase amount of changes there)

The new and old types aren't expected to mix, i.e. there's no support to sign a PublicKey with SignedPrivateKey and such. Many of the methods only refer to types of the compatible type.

There is some cleanup that wasn't strictly necessary, but seemed worthwhile while I was going through this code, e.g. tightening the imports (e.g. making ethers imports more specific).

There is a bit more explicit branching based on the types of things than I'd like, primarily on the Signature that now needs to accommodate two signature types. There's probably a way to map the union on the protocol side to a class hierarchy, but I didn't want to make things even messier than it already is. We may want to re-evaluate though as the number of types grows in the future.

### DIFF
--- a/src/crypto/Signature.ts
+++ b/src/crypto/Signature.ts
@@ -1,30 +1,85 @@
 import { signature } from '@xmtp/proto'
 import Long from 'long'
 import * as secp from '@noble/secp256k1'
-import PublicKey from './PublicKey'
+import { PublicKey, UnsignedPublicKey, SignedPublicKey } from './PublicKey'
+import { SignedPrivateKey } from './PrivateKey'
+import { Signer, utils } from 'ethers'
+import { bytesToHex, equalBytes, hexToBytes } from './utils'
 
-// Signature represents an ECDSA signature with recovery bit.
+// ECDSA signature with recovery bit.
+export type ECDSACompactWithRecovery = {
+  bytes: Uint8Array // compact representation [ R || S ], 64 bytes
+  recovery: number // recovery bit
+}
+
+// Validate signature.
+function ecdsaCheck(sig: ECDSACompactWithRecovery): void {
+  if (sig.bytes.length !== 64) {
+    throw new Error(`invalid signature length: ${sig.bytes.length}`)
+  }
+  if (sig.recovery !== 0 && sig.recovery !== 1) {
+    throw new Error(`invalid recovery bit: ${sig.recovery}`)
+  }
+}
+
+// Compare signatures.
+function ecdsaEqual(
+  a: ECDSACompactWithRecovery,
+  b: ECDSACompactWithRecovery
+): boolean {
+  return a.recovery === b.recovery && equalBytes(a.bytes, b.bytes)
+}
+
+// Derive public key of the signer from the digest and the signature.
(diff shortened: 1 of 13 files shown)


src/crypto/Signature.ts
```diff
@@ -51,3 +117,60 @@ export default class Signature implements signature.Signature {
...
   }
 }
+
+// A signer that can be used to sign public keys.
+export interface KeySigner {
+  signKey(key: UnsignedPublicKey): Promise<SignedPublicKey>
+}
+
+// A wallet based KeySigner.
+export class WalletSigner implements KeySigner {
```
neekolas, line 127: This new class is a nice improvement. Always was a little awkward having the signature text be included in the key class.

src/crypto/Signature.ts
```diff
@@ -1,30 +1,85 @@
...
 import Long from 'long'
 import * as secp from '@noble/secp256k1'
-import PublicKey from './PublicKey'
+import { PublicKey, UnsignedPublicKey, SignedPublicKey } from './PublicKey'
+import { SignedPrivateKey } from './PrivateKey'
+import { Signer, utils } from 'ethers'
+import { bytesToHex, equalBytes, hexToBytes } from './utils'
+
+// ECDSA signature with recovery bit.
+export type ECDSACompactWithRecovery = {
```
jazzz, line 10: Love that we are making this Type explicit!

jazzz (approved):
Love the cleanup to how keys are handled

neekolas (approved):

PR #160, opened 2022-09-16
Change default address in docs

jazzz:
Removing Vitalik's address from the default examples. 

New address points to a dedicated "hello" wallet for this purpose


neekolas:
We don't want to just point to Peter or Shane?

neekolas (approved):

PR #162, opened 2022-09-20
Key manager

neekolas:
## Summary
I wanted to break out this discrete piece of work for negotiated topics, since it can be safely merged to `main`.

It's a simple and narrowly scoped manager class for `TopicKey`s. I initially planned a broader `KeyManager` interface that would handle all types of keys, including `PrivateKeyBundle`s but ultimately couldn't find the synergy of having them all in one place. I think the right solution is probably to have general encrypt/decrypt functions on the client that make use of this class amongst others to get the right key and encrypt/decrypt a given message using the appropriate protocol for the message.

The functionality of this module:
1. Store `TopicKey`s, enforcing the "one key per topic" rule
2. Retrieve the `TopicKey` for a given topic to decrypt a received message
3. Retrieve the latest `TopicKey` for a given wallet to encrypt a message before sending

All encryption/decryption will be handled by the caller.

## Links
- [RFC](https://github.com/xmtp-labs/hq/blob/main/eng/RFCs/20220819-key-passing.md)

### DIFF
--- a/src/TopicKeyManager.ts
+++ b/src/TopicKeyManager.ts
@@ -1,23 +1,26 @@
-import PrivateKey from './crypto/PrivateKey'
-import PrivateKeyBundle from './crypto/PrivateKeyBundle'
 import PublicKeyBundle from './crypto/PublicKeyBundle'
 
-enum EncryptionAlgorithm {
+export enum EncryptionAlgorithm {
   AES_256_GCM_HKDF_SHA_256,
 }
 
-type TopicKeyRecord = {
+// TopicKeyRecord encapsulates the key, algorithm, and a list of allowed signers
+export type TopicKeyRecord = {
   keyMaterial: Uint8Array
   encryptionAlgorithm: EncryptionAlgorithm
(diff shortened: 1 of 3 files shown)


src/TopicKeyManager.ts
```diff
@@ -0,0 +1,136 @@
...
+    return {
+      topicKey,
+      contentTopic,
+    }
+  }
+
+  /**
+   *  Used to know which topic/key to use to send to a given wallet address
+   */
+  getLatestByWalletAddress(walletAddress: string): TopicResult | undefined {
```
mkobetic, line 116: IIUC this is to support the case where a 1:1 conversation was split into multiple topics, correct? This will break when group topics are introduced, won't it?

neekolas: I'm imagining group topics are going to be handled by an entirely different key management system (certainly would with ART)

mkobetic (approved):
I suspect this will evolve as we figure things out, but this is a good start 👍

PR #163, opened 2022-09-21
Include client version header with API requests

snormore:
This PR updates publish/subscribe/query API requests to include a `X-Client-Version` header with the current package version in the form `xmtp-js/1.0.0`

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     ]
   },
   "dependencies": {
+    "@jsbits/get-package-version": "^1.0.3",
     "@noble/secp256k1": "^1.5.2",
     "@stardazed/streams-polyfill": "^2.4.0",
     "@xmtp/proto": "^3.1.0",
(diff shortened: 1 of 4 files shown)


mkobetic (approved):
This looks good. I'm wondering do we want to allow devs to augment the value so that we can track the various apps as well? It could also be a separate header field so that we don't have to parse the value. Maybe that is out of scope here.

elisealix22 (approved):
Looks great to me! Might be nice to expose the string to client apps in an API so they can render it easily, but that can absolutely be a follow-up.

elisealix22 (approved):
💚  @bhavya2611 heads up that after this, we'll have the client version string available straight from the SDK!

package.json
```diff
@@ -59,6 +59,7 @@
     ]
   },
   "dependencies": {
+    "@jsbits/get-package-version": "^1.0.3",
```
neekolas, line 62: @snormore Got to this one a bit late, but I think you can replace this external dependency with
`import { version } from '../package.json'` in Typescript, so long as the tsconfig has the `resolveJsonModule` turned on.

PR #164, opened 2022-09-21
Remove get-package-version

neekolas:
## Summary
Removes the dependency on `get-package-version` and just pulls the version using a Typescript import


elisealix22 (approved):
Amazing. Love the minus 14k loc as well 🔥

neekolas:
FWIW, I'm not actually sure if this is going to pull in the correct version in the published package. Because of the way semantic-release dynamically sets the version, there's a timing issue. If the typescript import happens before semantic-release has updated the package.json it will get `0.0.0-development` instead of the real version. That would be the expected behaviour in local dev.

snormore (approved):
🙏

PR #165, opened 2022-09-21
Remove version altogether

neekolas:
The version header is causing CORS issues. This is causing all requests to the nodes to fail.

image


elisealix22 (approved):

neekolas:
Merging this while we work out the header issue

PR #166, opened 2022-09-21
Signed Public Key Bundle

mkobetic:
Fixes https://github.com/xmtp-labs/hq/issues/686

* Adds new `SignedPublicKeyBundle` to eventually replace `PublicKeyBundle`
* Renames `ContactBundle` to `ContactBundleV1` since that's what it implements and adds `ContactBundleV2`
* Makes `ContactBundle` a union of all supported versions, this forces turning `static ContactBundle.decodeV1(bytes)` into a more generic `DecodeContactBundle(bytes)`
* We're still publishing bare `PublicKeyBundle` to the contact topic, so enhance tests to check that `DecodeContactBundle()` can indeed handle that
* Updates `Client.getUserContactFromNetwork()` to ignore `SignedPublicKeyBundles` (V2) for now. This should allow newer clients to publish V2 bundle without breaking older clients.

I'll do the private key bundle in separate PR.

Rolling out V2 will be messy as the clients will have to manage V2 and V1 contacts separately during the transition period. They cannot be used interchangeably.

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -24,8 +24,8 @@ import {
 } from './MessageContent'
 import { decompress, compress } from './Compression'
 import { xmtpEnvelope, messageApi, fetcher } from '@xmtp/proto'
-import ContactBundle from './ContactBundle'
-import ApiClient from './ApiClient'
+import { DecodeContactBundle } from './ContactBundle'
+import ApiClient, { SortDirection } from './ApiClient'
 import { Authenticator } from './authn'
 const { Compression } = xmtpEnvelope
 const { b64Decode } = fetcher
@@ -208,16 +208,17 @@ export default class Client {
     let recipientKey: PublicKeyBundle | null = null
     const stream = this.apiClient.queryIterator(
       { contentTopics: [buildUserContactTopic(peerAddress)] },
-      { pageSize: 5 }
+      { pageSize: 5, direction: SortDirection.SORT_DIRECTION_DESCENDING }
     )
 
     for await (const env of stream) {
       if (!env.message) continue
-      const bundle = ContactBundle.fromBytes(b64Decode(env.message.toString()))
+      const bundle = DecodeContactBundle(b64Decode(env.message.toString()))
       const keyBundle = bundle.keyBundle
 
       const address = keyBundle?.walletSignatureAddress()
-      if (address === peerAddress) {
+      // TODO: Ignore SignedPublicKeyBundles for now.
+      if (address === peerAddress && keyBundle instanceof PublicKeyBundle) {
         recipientKey = keyBundle
         break
       }
--- a/src/ContactBundle.ts
+++ b/src/ContactBundle.ts
@@ -2,6 +2,7 @@ import { contact, publicKey } from '@xmtp/proto'
 import { PublicKeyBundle, SignedPublicKeyBundle } from './crypto'
 
 // ContactBundle packages all the information which a client uses to advertise on the network.
+// V1 uses the legacy PublicKeyBundle.
 export class ContactBundleV1 implements contact.ContactBundleV1 {
   keyBundle: PublicKeyBundle
 
@@ -22,6 +23,8 @@ export class ContactBundleV1 implements contact.ContactBundleV1 {
   }
 }
 
+// ContactBundle packages all the information which a client uses to advertise on the network.
+// V2 uses the SignedPublicKeyBundle.
 export class ContactBundleV2 implements contact.ContactBundleV2 {
   keyBundle: SignedPublicKeyBundle
 
@@ -42,8 +45,10 @@ export class ContactBundleV2 implements contact.ContactBundleV2 {
   }
 }
 
+// This is the union of all supported bundle versions.
 export type ContactBundle = ContactBundleV1 | ContactBundleV2
 
+// This is the primary function for reading contact bundles off the wire.
 export function DecodeContactBundle(bytes: Uint8Array): ContactBundle {
   let cb: contact.ContactBundle
   try {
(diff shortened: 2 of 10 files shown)


src/Client.ts
```diff
@@ -208,16 +208,17 @@ export default class Client {
     let recipientKey: PublicKeyBundle | null = null
     const stream = this.apiClient.queryIterator(
       { contentTopics: [buildUserContactTopic(peerAddress)] },
-      { pageSize: 5 }
+      { pageSize: 5, direction: SortDirection.SORT_DIRECTION_DESCENDING }
```
mkobetic, line 211: I think this is a bug, we want the freshest bundles right?

src/ContactBundle.ts
```diff
@@ -21,44 +20,44 @@ export default class ContactBundle implements contact.ContactBundleV1 {
...
+      v1: undefined,
+      v2: {
+        keyBundle: this.keyBundle,
+      },
+    }).finish()
+  }
+}
+
+type ContactBundle = ContactBundleV1 | ContactBundleV2
+export default ContactBundle
```
mkobetic, line 46: I'm not 100% sure how this is going to work out, but it would be nice if it did I think.

src/Message.ts
```diff
@@ -164,14 +164,14 @@ export default class Message implements proto.MessageV1 {
...
     }
-    const recipient = new PublicKeyBundle(
-      new PublicKey(header.recipient.identityKey),
-      new PublicKey(header.recipient.preKey)
-    )
-    const sender = new PublicKeyBundle(
-      new PublicKey(header.sender.identityKey),
-      new PublicKey(header.sender.preKey)
-    )
+    const recipient = new PublicKeyBundle({
```
mkobetic, line 167: I've changed the constructor to take the proto, to save duplicate undefined checking. That's what we do everywhere else too.

src/crypto/PublicKeyBundle.ts
```diff
@@ -29,15 +75,6 @@ export default class PublicKeyBundle implements publicKey.PublicKeyBundle {
 
   static fromBytes(bytes: Uint8Array): PublicKeyBundle {
     const decoded = publicKey.PublicKeyBundle.decode(bytes)
-    if (!decoded.identityKey) {
```
mkobetic, line 32: Can drop these, the constructor will check that.

src/Client.ts
neekolas: IIRC @jazzz was concerned about people spamming the topic

neekolas: We should be able to mitigate that by filtering out bogus bundles

src/ContactBundle.ts
neekolas: That'll work for the type, but not the implementation classes (which is really what you care about)

src/Client.ts
jazzz: @mkobetic Early on I did not want the freshest bundle.
- Migration to ContactBundlesV1 ran into the issue that older clients could not read the new bundle format until they upgraded.
- Currently the ContactBundle only changes when the bundle format changes. Yet it is published on every client initialization.
- To simplify reading, and negate spam issues it was easiest to read the first bundle only, rather than sorting through pages to find the most recent compatible bundle. 
- This also lead to a consistent experience between different versioned clients.

As the bundles are substantively different in V2, you are correct we'll need to handle this properly via filtering or adding a retry on ContactBundle Decoding. 

I'd suggest now is the time for a model where the contactBundle is only published if needed.

neekolas: +1 on only publishing if needed. I must have hundreds in my topic

src/ContactBundle.ts
neekolas: In other words, your default export here is a type instead of a class, which won't be very helpful when someone tries to import it

mkobetic: This seems to me to be the most sensible definition of what `ContactBundle` should be. I don't see a sensible way to make it a class. Do you agree with that?

I have no argument why exporting this type is useful, but I note that we do have examples of that. Bunch of the proto things we import are interfaces, so exporting types shouldn't be completely useless. Would refrain from exporting it for now, until it proves it could be useful?

src/Client.ts
mkobetic: I agree with fixing the publishing, but I think it could be argued it's out of scope for this PR. Do you see it as a gating factor for this PR?

I can definitely put more effort into making the filtering more robust. Would that be satisfactory?

src/ContactBundle.ts
neekolas: Exporting it seems fine, but I think the default export from a module should be something concrete. It's kinda an implicit assumption developers have that a default export is something you can actually use.

src/Client.ts
neekolas: That seems reasonable for this PR. We can handle the publishing frequency when we switch to publish V2 bundles

src/ContactBundle.ts
neekolas: No default export seems fine

src/Client.ts
mkobetic: I double checked, we are [filtering out invalid signatures](https://github.com/xmtp/xmtp-js/pull/166/files#diff-b7961e132a9ada7fe5b4cdec69198fc7e4fe506e45b5a7fe29f399a01c45ff53R221), so I think we should be good to go in this regard.

WRT to spamming one's contact topic with junk, republishing valid bundle frequently is probably actually helpful with avoiding having to page through many pages of garbage to get to a valid bundle. When we go fix that we may need to look at checking the contact topic regularly to see what's on top, rather than just at client start time.

mkobetic:
@neekolas is this good to go?

neekolas (approved):

PR #167, opened 2022-09-21
Revert "Remove version altogether"

neekolas:
Whenever we have the `access-control-allow-headers` including the `x-client-version` in the dev env, and have tested with version 6.3.1 of the SDK, we can merge this.


snormore (approved):

neekolas:
I guess because I reverted through the web ui it didn't have commits matching the pattern

Issue #168, opened 2022-09-21
Paginated message retrieval


saulmc:
### Context

Currently, `xmtp-js` does not expose a way to paginate a user's messages when retrieving them from the network. Pagination is available at the network level and is already used to assemble the `conversation.messages()` result. However, this result is currently only returned after the final page is loaded.

### Problem

When a user has hundreds of messages, load time appears slower because all messages in a conversation must all be retrieved from the network before any of them can be displayed.

### Goal

Provide an interface for retrieving a user's messages page-by-page, so that recent messages can be displayed sooner while older messages are still being retrieved.

neekolas:
I'm gonna sneak this one into this sprint some time when I am blocked on something else. It feels very high ROI, given how simple it is.

PR #171, opened 2022-09-23
Add ability to read messages with pagination

neekolas:
## Summary
Adds ability to read paginated messages, resolving https://github.com/xmtp/xmtp-js/issues/168.

These are the options I have made available to control the query:
```ts
export type ListMessagesPaginatedOptions = {
  startTime?: Date
  endTime?: Date
  pageSize?: number
  direction?: messageApi.SortDirection
}
```

## Notes
- This returns an `AsyncGenerator` yielding one page of results at a time. While I could simply return an iterator of all messages instead of an iterator of pages and still use pagination under the hood, this is a little more performant (saves waiting until `process.nextTick()` for each `yield`/message). It also lets the caller process the results one page at a time, which may be advantageous for React apps where they would want to batch updates to the DOM. I could be convinced to go the other way.
- I have not implemented listing conversations in a paginated manner. Mostly because no one has ever asked for that. Maybe we want to offer both just for the sake of consistency (it's trivially easy), but for the moment I'm inclined to punt for the sake of simplicity.


src/Client.ts
```diff
@@ -50,6 +51,13 @@ export type ListMessagesOptions = {
   limit?: number
 }
 
+export type ListMessagesPaginatedOptions = {
+  startTime?: Date
+  endTime?: Date
+  pageSize?: number
+  direction?: messageApi.SortDirection
```
mkobetic, line 58: I guess we don't want limit here, because you can just quit the pagination loop whenever you want, correct?

mkobetic:
This looks good, but one thing I wonder is that we're using Streams for new messages and async generator for pages of historic messages. The two constructs seem kinda equivalent to me, are they? If so using both may make the API look a bit schizophrenic. Is that a valid concern?

src/Client.ts
neekolas: That's right

neekolas:
> The two constructs seem kinda equivalent to me, are they? If so using both may make the API look a bit schizophrenic. Is that a valid concern?

From the user's perspective, the Streams are really just async iterators. The whole Stream concept is mostly an internal class for us to manage filling the async iterator from callbacks. Async Iterators and Async Generators work the same from the caller's perspective, so I'm not sure most users will notice.

Having worked with Async Generators a bit on the `ApiClient`, I'm starting to think it would be a nice refactor to rewrite the Stream class as an Async Generator as well. The code is much [cleaner](https://github.com/xmtp/xmtp-js/blob/main/src/ApiClient.ts#L226) and it handles a lot of the semantics around returning from the stream for you natively.

mkobetic:
> Having worked with Async Generators a bit on the `ApiClient`, I'm starting to think it would be a nice refactor to rewrite the Stream class as an Async Generator as well. The code is much [cleaner](https://github.com/xmtp/xmtp-js/blob/main/src/ApiClient.ts#L226) and it handles a lot of the semantics around returning from the stream for you natively.

Yeah, maybe the right thing to do is to get rid of the Stream in favour of the generator. Can we do that with reasonably backward compatible manner?

neekolas:
> Can we do that with reasonably backward compatible manner?

Ya, that's the nice thing. Callers wouldn't even notice the change. Only the types would be different, but you can use them the same way.

neekolas:
@mkobetic what do you think about merging this as-is and then I will make a backlog ticket for converting our Stream implementation to an Async Generator

mkobetic (approved):
Yes, :shipit:. Sorry I thought I already 👍 this.

Issue #169, opened 2022-09-22
Resolve react-scripts ^5.0.0 incompatibility


saulmc:
### Context

Due to [changes to polyfills in webpack v5](https://github.com/facebook/create-react-app/issues/11756), the `node-localstorage` dependency which relies on `fs` cannot be used with `react-scripts ^5.0.0`. The `xmtp-js` README therefore guides developers to use lower versions of `react-scripts`.

### Problem

Developers whose apps require `react-scripts ^5.0.0` cannot use `xmtp-js`. Although the requirement to use a lower version of this module usually isn't an issue when starting a new app, it can be prohibitively difficult to downgrade an existing app that was created using an incompatible `react-scripts` version.

### Goal

Remove dependencies that are incompatible with webpack v5 and `react-scripts ^5.0.0`.

#### Potential solution

Use a LocalStorage polyfill that doesn't require the filesystem and keeps the data in memory.

rygine:
check out our new docs on solving polyfill issues with webpack and react-scripts v5 here: https://xmtp.org/docs/faq#why-my-app-is-failing-saying-buffer-is-not-found

PR #173, opened 2022-09-26
Strip localstorage dependency from production

elisealix22:
Resolves https://github.com/xmtp/xmtp-js/issues/169

This PR strips our `node-localstorage` dependency from production bundles since it's only required for tests. This works because we set the mode to `production` in our webpack config.

https://github.com/xmtp/xmtp-js/blob/a1cc63336e5549d3f99c2cb3139ecd54a48b0cba/config/webpack.config.js#L5

Test cases:
- [x] Web app runs pointing to local version
- [x] React native app runs without local storage polyfill
- [x] Can import `xmtp-js` using React Scripts >= 5.0.0 (tested 5.0.1)


package.json
```diff
@@ -63,28 +63,28 @@
...
   },
   "devDependencies": {
     "@commitlint/cli": "^16.1.0",
     "@commitlint/config-conventional": "^16.0.0",
     "@types/bl": "^5.0.2",
     "@types/callback-to-async-iterator": "^1.1.4",
     "@types/jest": "^27.0.1",
-    "@typescript-eslint/eslint-plugin": "^4.31.1",
-    "@typescript-eslint/parser": "^4.31.1",
+    "@typescript-eslint/eslint-plugin": "^5.38.0",
```
elisealix22, line 74: I was [seeing](https://github.com/xmtp/xmtp-js/actions/runs/3128031452/jobs/5075358977) eslint resolution errors & upgrading resolved the issue.

README.md
```diff
@@ -42,16 +42,6 @@ npm install @xmtp/xmtp-js
 
 Additional configuration is required in React environments due to the removal of polyfills from Webpack 5.
 
-### Create React App
-
-Use `react-scripts` prior to version `5.0.0`. For example:
-
-```bash
-npx create-react-app my-app --scripts-version 4.0.2
```
elisealix22, line 50: I removed this entirely since it seemed like it was only added to detail the `--scripts-version` (which is no longer required). Lmk if you prefer we keep a create react app line here!

neekolas:
Have you tested to ensure that this works in node.js? Node-localstorage is not just used for tests, but also for Node apps who don't want to use the `NetworkTopicStore`.

package.json
neekolas: I ran into the same thing and made the same change in one of my PRs. Apologies for the merge conflicts.

src/store/LocalStorageStore.ts
```diff
@@ -16,11 +18,15 @@ export default class LocalStorageStore implements Store {
   }
 
   async get(key: string): Promise<Buffer | null> {
+    if (process.env.NODE_ENV === 'production') return null
```
neekolas, line 21: I think `process.env.NODE_ENV` will be `production` in the deployed version of our example chat (and certainly could be for other people's deployed apps. This will silently make the `LocalStorageStore` unusable in those cases.

elisealix22:
Closing this as I learned LocalStorage is also a feature for Node servers, not only used for tests. Moving forward we can either:

1. Deprecate LocalStorage as a feature and have people switch to export keys following [the guide in our README](https://github.com/xmtp/xmtp-js#manually-handling-private-key-storage). I think in this case we'll want to remove the `LocalStorageStore` entirely for production builds so it's not misleading.
2. Continue to support it as a feature and find a better workaround for react scripts.

neekolas:
My vote is for 1

PR #174, opened 2022-09-26
Check is a user is on the network without instantiating SDK

neekolas:
## Summary
- Adds the ability to check if a blockchain address has been registered to the XMTP network without instantiating a SDK instance. This can be helpful for cases where a developer wants to only show the signature prompt conditionally if another address is registered, or if the developer wants to show different educational text on the signature prompt depending on whether it was the user's first or subsequent time using XMTP.
- I was running into dependency conflicts after running `npm i` due to mismatched peer deps in eslint and its plugins. I've upgraded those to the latest versions.

## Links
https://github.com/xmtp-labs/hq/issues/655


src/Client.ts
```diff
@@ -561,3 +550,30 @@ function createApiClientFromOptions(options: ClientOptions): ApiClient {
...
+// retrieve a key bundle from given user's contact topic
+async function getUserContactFromNetwork(
+  apiClient: ApiClient,
+  peerAddress: string
+): Promise<PublicKeyBundle | undefined> {
+  // have to avoid undefined to not trip TS's strictNullChecks on recipientKey
+  let recipientKey: PublicKeyBundle | null = null
+  const stream = apiClient.queryIterator(
+    { contentTopics: [buildUserContactTopic(peerAddress)] },
+    { pageSize: 5, direction: SortDirection.SORT_DIRECTION_DESCENDING }
```
elisealix22, line 563: I'm not sure I fully understand the page size here. Is it necessary to guarantee a PublicKeyBundle will return if they are on the network?

elisealix22 (approved):
Just one Q on reading the contact bundle, otherwise looks great to me!

mkobetic:
This works, but I wonder if this is the right direction. I'm assuming the issue is with our client creation hardwiring the private key handling into it. So presumably another option is to separate the two. That would be a breaking API change, but would provide access to any capabilities that don't require the private key, instead of just whatever is extracted out.

src/Client.ts
neekolas: So, it's based on the assumption that a winning key bundle will likely be in the first 5 results. It will potentially pull all results down from the network, but will do it 5 at a time, and aborts once it finds a bundle that validates.

neekolas:
> This works, but I wonder if this is the right direction. I'm assuming the issue is with our client creation hardwiring the private key handling into it. So presumably another option is to separate the two. That would be a breaking API change, but would provide access to any capabilities that don't require the private key, instead of just whatever is extracted out.

That's true, although most of the SDK functionality requires access to private keys. This is one of very few exceptions that does not. If we start to have more cases like this, I'm open to a refactor. I'm just not sure how many there are going to be.

mkobetic:
> > This works, but I wonder if this is the right direction. I'm assuming the issue is with our client creation hardwiring the private key handling into it. So presumably another option is to separate the two. That would be a breaking API change, but would provide access to any capabilities that don't require the private key, instead of just whatever is extracted out.
> 
> That's true, although most of the SDK functionality requires access to private keys. This is one of very few exceptions that does not. If we start to have more cases like this, I'm open to a refactor. I'm just not sure how many there are going to be.

That is fair, we can do this later too. 👍

mkobetic (approved):

test/helpers.ts
```diff
@@ -43,7 +43,7 @@ export async function waitForUserContact(
 ): Promise<PublicKeyBundle | undefined> {
   return pollFor(
     async () => {
-      const contact = await c1.getUserContactFromNetwork(c2.address)
+      const contact = await c1.getUserContact(c2.address)
```
jazzz, line 46: [pebble] This is a substantial change to the ContactBundle tests as we are potentially using the cached value, rather than the freshest bundle on the network. 

While the this will not cause a change in behavior today, when Key Rotation is added there will be a meaningful difference between `GetUserContact` and `GetUserContactFromNetwork`

Will need future consideration

jazzz (approved):

test/helpers.ts
neekolas: I hadn't considered the implications around rotation, but we can always refactor to export the `getUserContactFromNetwork` function on the client and use that in tests if needed. I left it as-is, as it didn't seem like switching to the cached provider had an impact on the current test suite.

PR #175, opened 2022-09-26
Signed private key bundle

mkobetic:
This PR adds `PrivateKeyBundleV2`; it is largely a clone of `PrivateKeyBundleV1` (used to be called just `PrivateKeyBundle`) with bunch of tweaks to adjust for the V2 properties. 

To contain the fallout of the V2 changes, I had to explicitly restrict bunch of things in the main package to `PrivateKeyBundleV1` explicitly. Many of these places will be adjusted in follow up PRs to deal with the two versions. In this PR most things still operate with the V1 assumption only.

To avoid duplicating the wallet encryption logic I moved all of that to the EncryptedKeyStore (I also renamed it to make clear the distinction between Store and KeyStore) and also tied it to V1 only for now.  I also want to rename the file, but git doesn't recognize it as a rename due to the amount of changes, so I left it be for now.

The PrivateKeyBundle.ts diff is terrible, I mostly just removed all the encryption related bits, then duplicated the V1 bundle and tweaked the first copy to be V2. The V1 is largely unmodified modulo the encryption removal.

I'll squash/rebase after review with proper commit message.


src/ContactBundle.ts
```diff
@@ -43,6 +43,12 @@ export class ContactBundleV2 implements contact.ContactBundleV2 {
       },
     }).finish()
   }
+
+  static fromLegacyBundle(bundle: ContactBundleV1): ContactBundleV2 {
```
mkobetic, line 47: forgot this one in #176

src/crypto/PrivateKeyBundle.ts
```diff
@@ -98,159 +100,142 @@ export default class PrivateKeyBundle implements proto.PrivateKeyBundleV1 {
...
-      e instanceof RangeError ||
-      (e instanceof Error && e.message.startsWith('invalid wire type'))
-    ) {
-      // Adds a default fallback for older versions of the proto
-      return [proto.PrivateKeyBundleV1.decode(bytes), true]
-    }
-    throw new Error("Couldn't decode private bundle:" + e)
+export type PrivateKeyBundle = PrivateKeyBundleV1 | PrivateKeyBundleV2
+
+export function DecodePrivateKeyBundle(bytes: Uint8Array): PrivateKeyBundle {
```
neekolas, line 232: One thing I only noticed after review is the casing of `DecodeContactBundle`. Conventionally, Javascript functions are camelCased and not PascalCased (with Pascal Case reserved for classes and constructors). Think we should stick with that convention

src/store/EncryptedStore.ts
```diff
@@ -30,39 +34,119 @@ export default class EncryptedStore implements KeyStore {
...
+    return [proto.EncryptedPrivateKeyBundleV1.decode(bytes), true]
+  }
+  throw new Error('unrecognized encrypted private key bundle version')
+}
+
+// getPrivateV1Bundle returns the decoded bundle from the provided bytes. If there is an error decoding the bundle it attempts
+// to decode the bundle as a legacy bundle. Additionally return whether the bundle is in the expected format.
+function getPrivateBundle(bytes: Uint8Array): [PrivateKeyBundleV1, boolean] {
+  try {
+    const b = DecodePrivateKeyBundle(bytes) as PrivateKeyBundleV1
```
neekolas, line 132: This type assertion seems dangerous, given the function can return either V1 or V2 bundles. Today, we only write V1 so it's safe in practice. But we'll need to be careful that we update this logic to actually check the return value once we have the possibility of V1 or V2 bundles being stored on the network.

test/TestClient.ts
```diff
@@ -1,41 +0,0 @@
-import { Signer } from 'ethers'
```
neekolas, line 1: Good catch. Think this was being ignored because it didn't have the `.test.ts` suffix

test/crypto/PrivateKeyBundle.test.ts
```diff
@@ -1,24 +1,46 @@
...
 describe('Crypto', function () {
   describe('PrivateKeyBundle', function () {
+    it('v2 generate/encode/decode', async function () {
+      const wallet = newWallet()
+      // generate key bundle
+      const bundle = await PrivateKeyBundleV2.generate(wallet)
+      const bytes = bundle.encode()
+      const bundle2 = DecodePrivateKeyBundle(bytes)
+      expect(bundle2).toBeInstanceOf(PrivateKeyBundleV2)
+      assert.ok(bundle.equals(bundle2 as PrivateKeyBundleV2))
```
neekolas, line 26: We might be able to use a discriminated union type (by having an explicit `version` field on each class), that would let us check for which type of bundle we have without needing a type assertion. The type checker should treat the value as a `v2` if we include a line like `if (bundle.version === 1) { throw new Error }` ahead of usage

neekolas (approved):
Couple nits, but I think this looks good

src/crypto/PrivateKeyBundle.ts
mkobetic: Thanks for calling this out, I have no clue about the conventions, so please do so whenever you see something like this.

src/store/EncryptedStore.ts
mkobetic: All these assertions are temporary and should be removed as we roll out V2. I can mark them with TODO comments if that helps.

test/crypto/PrivateKeyBundle.test.ts
neekolas: I've never actually done it with classes, but it works great with objects. 🤞

mkobetic: I added the version property but it still won't let me omit the type assertion on line 26, even if I explicitly throw when version is not 2. 🤷 I'll leave the property as it will be useful for runtime checks I think.

PR #176, opened 2022-09-28
Key version conversion

mkobetic:
This PR adds the ability to convert the legacy style V1 key types to V2 key types. This require the ability to transfer the public key signature to the new key as is and consequently the unsigned key bytes to be identical between the two keys.

This is complicated by the fact that we've changed the timestamp resolution between the versions from ms to ns. To accommodate this discrepancy this PR allows storing ms value in `UnsignedPublicKey.createdNs` field and adjusts the `timestamp` and `created` accessors to distinguish between the two cases based on the magnitude of the value stored in `createdNs` field. Consequently this means clients should refrain from accessing `createdNs` directly.

The reason why I want to propose this ugliness is that the ability to convert V1 keys to V2 keys will simplify the transition greatly. Otherwise we'll have to regenerate (or at least resign) all user keys to enable V2 functionality. We will likely need to preserve the ability to read V1 public keys to be able to read old messages, but we should be able to drop the private keys and just converting the V1 public keys to V2 and use them with V2 private keys (assuming they are the same as the old V1 keys). This will allow preserving the key store APIs which are set up for handling single key bundle only.

So in short, this is not great, but I think it's worth it.

### DIFF
--- a/src/crypto/PublicKey.ts
+++ b/src/crypto/PublicKey.ts
@@ -22,9 +22,15 @@ function secp256k1UncompressedCheck(key: secp256k1Uncompressed): void {
   }
 }
 
+const MS_NS_TIMESTAMP_THRESHOLD = new Long(10 ** 9).mul(10 ** 9)
+
 // Basic public key without a signature.
 export class UnsignedPublicKey implements publicKey.UnsignedPublicKey {
-  createdNs: Long // time the key was generated, ns since epoch
+  // time the key was generated, normally ns since epoch, however
+  // to allow transparent conversion of pre-existing signed PublicKey to SignedPublicKey
+  // it can also be ms since epoch; use MS_NS_TIMESTAMP_THRESHOLD to distinguish
+  // the two cases.
+  createdNs: Long
   secp256k1Uncompressed: secp256k1Uncompressed // eslint-disable-line camelcase
 
   constructor(obj: publicKey.UnsignedPublicKey) {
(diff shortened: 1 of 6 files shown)


src/crypto/PublicKey.ts
```diff
@@ -22,9 +22,15 @@ function secp256k1UncompressedCheck(key: secp256k1Uncompressed): void {
   }
 }
 
+const MS_NS_TIMESTAMP_THRESHOLD = new Long(10 ** 9).mul(10 ** 9)
```
neekolas, line 25: I think this is too big. I get `1,000,000,000,000,000,000` (19 digits), but when I run `new Long(new Date().valueOf()).mul(1_000_000).toString()` I get a 16 digit number

neekolas: Hmm. Although the test below seems to work. Maybe I'm missing something.

mkobetic: Yes, ms timestamps are well below that value and ns timestamps are just above.

mkobetic: IOW ms timestamps are just over 10**12 and ns timestamps just over 10**18

mkobetic: I think you need to do `Long.fromNumber()` in your expression. I think the constructor truncates the first param to 32 bits.

neekolas (approved):

src/crypto/PublicKey.ts
neekolas: That did the trick

PR #177, opened 2022-09-29
Negotiated topic invitation classes

neekolas:
## Summary
- Creates classes for all Invitation related protos
- Enables decryption of encrypted bundles
- Adds test suite for Invitation classes

## Notes
- Does not include anything to do with tagged topics. That's for another day
- The diff should get a lot smaller once https://github.com/xmtp/xmtp-js/pull/175 goes in


src/crypto/PrivateKey.ts
```diff
@@ -237,7 +237,7 @@ export class PrivateKey implements privateKey.PrivateKey {
 
   // derive shared secret from peer's PublicKey;
   // the peer can derive the same secret using their PrivateKey and our PublicKey
-  sharedSecret(peer: PublicKey): Uint8Array {
+  sharedSecret(peer: PublicKey | SignedPublicKey): Uint8Array {
```
mkobetic, line 240: Do you actually need this? I was hoping to keep V1 and V2 keys completely separate. Invitations should deal strictly with V2 keys only.

neekolas: We don't strictly need it, but in the case of a shared secret they are interoperable and safe to mix and match. So, I don't see the harm to have that flexibility.

mkobetic: My concerns my not be completely warranted, the keys are indeed fundamentally the same after all. But it seems prudent to me to avoid potential complications, if we don't really need to mix the two versions. I don't feel that strongly about this, but my preference would be to not do this unless we need to.

neekolas: The typechecker should protect us if we try and access a property only available on one type.

src/Invitation.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  static fromBytes(bytes: Uint8Array): SealedInvitation {
+    return new SealedInvitation(invitation.SealedInvitation.decode(bytes))
+  }
+
+  /**
+   * Create a SealedInvitation with a SealedInvitationV1 payload
+   * Will encrypt all contents and validate inputs
+   */
+  static async createV1({
```
mkobetic, line 173: This is just a nitpick, but would this make more sense as `SealedInvitationV1.create()` ?

src/Invitation.ts
```diff
@@ -0,0 +1,201 @@
...
+  }: {
+    sender: PrivateKeyBundleV2
+    recipient: SignedPublicKeyBundle
+    created: Date
+    invitation: InvitationV1
+  }): Promise<SealedInvitation> {
+    const headerBytes = new SealedInvitationHeaderV1({
+      sender: sender.getPublicKeyBundle(),
+      recipient,
+      createdNs: Long.fromNumber(created.valueOf()).multiply(1_000_000),
```
mkobetic, line 187: We should probably create a helper for this, `nsNow()` or something, and hunt down all the places we do these conversions.

mkobetic (approved):
Few nitpicks, but otherwise very nice 👍

src/Invitation.ts
neekolas: Agreed. I can fold that in

src/Invitation.ts
neekolas: My thinking is that I don't want all consumers have to have to do `new SealedInvitation({ v1: await SealedInvitationV1.create(...) })`. Feels like no one wants the V1 invitation, and everyone will want the actual proto that is going over the wire.

PR #180, opened 2022-10-03
Update output locations

neekolas:
## Summary
In a recent change we started importing a file outside of the src directory, which leads to this [interesting behaviour](https://stackoverflow.com/a/58941798). It changes the structure of the output directories. 

This updates our package.json to reflect the new structure.


elisealix22 (approved):
Amazing find. Thank you! Confirmed this resolves the Lenster import locally as well.

PR #181, opened 2022-10-04
fix: pass message timestamp to the api envelope

mkobetic:
The API envelope timestamps can be slightly off compared to the message timestamps. In https://github.com/xmtp-labs/hq/issues/679 we've agreed that they should be equivalent. The reason is that the message timestamp is not being used to construct the envelope when publishing the message. This PR fixes that.


src/Client.ts
```diff
@@ -282,14 +282,19 @@ export default class Client {
...
     )
 
     return this.decodeMessage(msgBytes, topics[0])
   }
 
-  async publishEnvelope(env: messageApi.Envelope): Promise<void> {
+  async publishEnvelope(env: {
+    contentTopic: string
+    message: Uint8Array
+    timestamp?: Date
```
mkobetic, line 296: I dropped the messageApi.Envelope here so that I don't have to flip back and forth between Date and the string encoding of the ns timestamp.

neekolas (approved):
❤️

PR #182, opened 2022-10-05
feat: simplify contact bundle api

mkobetic:
The ContactBundleV1/V2 classes aren't really doing much that isn't already covered by PublicKeyBundle and SignedPublicKeyBundle.


neekolas (approved):
This seems like a good simplification

PR #183, opened 2022-10-05
Add sorting of messages to list endpoint

neekolas:
## Summary

One small feature I noticed was missing was the ability to sort the direction of messages from the `listMessages` endpoints. This is particularly useful to support "give me the 3 most recent messages" without having to use the paginated API. We should totally be doing this to load the conversation list more quickly in the example app (we currently load all messages, when we really just need the first one).

Surprised more developers aren't complaining about this missing feature. Always good to be eating your own dogfood I guess.

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -49,6 +49,7 @@ export type ListMessagesOptions = {
   startTime?: Date
   endTime?: Date
   limit?: number
+  direction?: messageApi.SortDirection
 }
 
 export type ListMessagesPaginatedOptions = {
@@ -491,7 +492,8 @@ export default class Client {
     const res = await this.apiClient.query(
       { contentTopics: [topic], startTime, endTime },
       {
-        direction: messageApi.SortDirection.SORT_DIRECTION_ASCENDING,
+        direction:
+          opts.direction || messageApi.SortDirection.SORT_DIRECTION_ASCENDING,
         limit,
       }
     )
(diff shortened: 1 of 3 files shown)


src/Client.ts
```diff
@@ -49,6 +49,7 @@ export type ListMessagesOptions = {
   startTime?: Date
   endTime?: Date
   limit?: number
+  direction?: messageApi.SortDirection
```
mkobetic, line 52: Do we want to expose the `messageApi.SortDirection` type? It feels a bit cumbersome, given that this could basically be a boolean.

neekolas: We already expose it on the paginated endpoint

mkobetic: Hm, worth fixing that too?

mkobetic (approved):

src/Client.ts
neekolas: That would be a breaking API change and require a major version update. We certainly can if we feel strongly about it

neekolas: Could also make a backlog item and roll it in to the next major version update

mkobetic: I don't feel strongly about it, if you agree then sounds good, if not I'm fine leaving it as is.

neekolas: I can see how it's a nicer API. I'd be fine rolling it into the next major version update.

PR #184, opened 2022-10-07
WIP - Invite Loading

neekolas:
## Summary

Sharing this PR as a very-unfinished draft, just so we can have a conversation about the approach taken before I get any further. It also can't be merged until we have support for reading V2 messages.

I've made notes on the parts of the code worth looking at. Ignore the rest. There's no tests, there are probably bugs, I'll probably refactor some of this, etc etc.

## Invite Loading Constraints

With the move to negotiated topics selecting the right pattern used to read each account’s invite topic is critical. Before sending or reading any DM, the client needs to have loaded the invite related to that topic.

Since there is only one invite topic per blockchain address, and users will have at least one invite for each ongoing conversation, the topic has the potential to get quite large.

There are a limited number of tools we can use to selectively read from this topic, given that the contents are either encrypted or obscured inside protobuf headers.

- We can sort by timestamp in either ascending or descending order
- We can query for a time range
- We can limit the size of each page of results

## Use Cases
These are the main use cases that will require data from the invite topic.

1. Loading all conversations (`client.conversations.list()`)
2. Loading a single conversation (`client.conversations.findOrCreate()`) in order to send a message. This may or may not include a `conversationId` to narrow the scope.
3. Streaming new conversations as they come in (`client.conversations.stream()`)

## Optimizations Possible
I don’t see a universal solution that is the perfect for all three cases. Many optimizations that would help one use-case would disadvantage another.

In many ways, the problem mirrors the access patterns required in a limited database with a very slow disk. Sometimes you need to do a full table scan, sometimes a partial scan, and sometimes you can rely on indexes.

### Use Case 1
By definition, loading all conversations will involve reading all invites. No optimizations possible for the first time run in a client session. Full table scan is desired.

If we were to load this list in ascending order we do have the possibility of keeping a marker of the timestamp of the newest invite found. On subsequent calls, we could limit the results to only invites newer than that timestamp (minus ~30 seconds to account for race conditions), making subsequent calls very cheap. It is reasonable to assume that most invites will be older than the last `list()` call in a session and excluded from the results.

### Use Case 2
It is possible to optimize this use-case. The client could iterate through a paginated list of invitations and stop when it had found an invite matching the `address/conversationId` combination desired. This is true with or without conversation tagging. Topic rotations would complicate this approach, since there would be more than one invite for an `address/conversationId` combo. To solve for that case we would need to include a marker on all but the first invitation marking that it was a continuation of an existing invite. We would then continue iterating through the list until we found the first invite.

We could even include the timestamp of the previous invite in the marker, which would allow the client to skip ahead to the right starting point to find the next invite. This would turn the invites into more of a linked list, resulting in a greater number of queries but less results returned. The number of serial round trips to the server in that scenario may work out to be more costly than simply requesting more invites than are needed and hoping we find more than one matching invite in a page of results. This would depend on the average number of rotations.

Any approach that involves stopping early will be a substantial improvement for finding recent conversations, but may be less optimal in other scenarios.

- Loading the oldest conversation would require loading all invites, since  this approach needs to happen in descending order (at least in a world with topic rotation)
- Loading multiple conversations at different times. By stopping early, we prevent ourselves from building a complete cache of older records that can be re-used for subsequent queries. We can still store invites that were collected incidentally (we were looking for conversations with address A, but came across conversations with B, C, and D along the way), but the cache will not be complete until we happen upon a query that takes us to the last page of results. An invite not already seen may be earlier than the earliest result of the last run, or later than the latest.
- Page size matters a lot here. Too small a page size and you will have a lot of requests to find the invite you need. Too large a page size and you will read many invites unrelated to your query (including potentially invites the client has seen before). You could get clever with using range queries based on previously seen results to omit certain ranges that have already been scanned, but that can get very complicated very fast.

	We will have to handle the uncertainty of having an incomplete cache by reading some of the same invites multiple times to satisfy different queries. At the very least, the first page of results will need to be read a lot.

### Use Case 3
Streaming in a world where we have a single invite per `address/conversationId` combination is very straightforward. We get a new invite and can use it as-is. No optimization needed.

If we were to have topic rotation, things get more complicated. If the newly streamed invite was a continuation of a previous conversation we would have to go through the same process as use-case 2 to join the new invite with older ones. Even if we had already cached some previous invites for that `address/conversationId` combination, we would still need to check and ensure that new invites had not been sent in the time between when the cache was loaded and when the stream had started.

## Conclusion
If I had to pick one optimization, my vote is for the optimizations for use-case 1. We know it is a very common pattern in the SDK to list all conversations and then send messages to those convos. If the average application has to load the full invite list at any point, it is the most efficient solution. And in many cases it is the most efficient approach even when the client never lists all conversations.

By loading conversations in ascending order, we only have to go through the pain of a “full table scan” once in a client session. All subsequent calls, including the ones in use-case 2 and 3, would only have to query for results after the newest discovered invite. This will almost always be a very small result set.

A hybrid of the optimizations for use-case 1 and use-case 2 might be possible, but are pretty complicated and feel like a premature optimization at this point. Open to ideas on how we can do this more cleanly.

## Drawbacks
- Given that we don’t have a concept of trusted timestamps, it would be possible to send a backdated invite that would not be picked up in a subsequent loading of the invite topic. This would resolve itself on a browser refresh or a new client instantiation
- This will be slower than other solutions for apps that only send to a single address (for example, a customer support app)

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -1,10 +1,14 @@
+import { PrivateKeyBundleV2 } from './crypto/PrivateKeyBundle'
+import { SealedInvitation } from './Invitation'
 import { PublicKeyBundle, PrivateKeyBundleV1 } from './crypto'
 import Message from './Message'
 import {
   buildDirectMessageTopic,
   buildUserContactTopic,
   buildUserIntroTopic,
+  buildUserInviteTopic,
   mapPaginatedStream,
+  nsLongToDate,
 } from './utils'
 import Stream, { MessageFilter, noTransformation } from './Stream'
 import { Signer } from 'ethers'
@@ -28,6 +32,7 @@ import { xmtpEnvelope, messageApi, fetcher } from '@xmtp/proto'
 import { decodeContactBundle } from './ContactBundle'
 import ApiClient, { SortDirection } from './ApiClient'
 import { Authenticator } from './authn'
+import TopicKeyManager, { EncryptionAlgorithm } from './TopicKeyManager'
 const { Compression } = xmtpEnvelope
 const { b64Decode } = fetcher
 
@@ -134,8 +139,10 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
 export default class Client {
   address: string
   keys: PrivateKeyBundleV1
+  keysV2: PrivateKeyBundleV2
   apiClient: ApiClient
   private contacts: Set<string> // address which we have connected to
+  private topicKeyManager: TopicKeyManager
   private knownPublicKeyBundles: Map<string, PublicKeyBundle> // addresses and key bundles that we have witnessed
   private _conversations: Conversations
   private _codecs: Map<string, ContentCodec<any>>
@@ -145,11 +152,13 @@ export default class Client {
     this.contacts = new Set<string>()
     this.knownPublicKeyBundles = new Map<string, PublicKeyBundle>()
     this.keys = keys
+    this.keysV2 = PrivateKeyBundleV2.fromLegacyBundle(keys)
     this.address = keys.identityKey.publicKey.walletSignatureAddress()
     this._conversations = new Conversations(this)
     this._codecs = new Map()
     this._maxContentSize = MaxContentSize
     this.apiClient = apiClient
+    this.topicKeyManager = new TopicKeyManager()
   }
 
   /**
@@ -507,6 +516,57 @@ export default class Client {
       msgs = msgs.filter(filterForTopics([topic]))
     }
     return msgs
+  }
+
+  async listInvites(): Promise<SealedInvitation[]> {
+    const envelopes = await this.apiClient.query(
+      {
+        contentTopics: [buildUserInviteTopic(this.address)],
+      },
+      {}
+    )
+    const bundle = PrivateKeyBundleV2.fromLegacyBundle(this.keys)
+    const invites: SealedInvitation[] = []
+    for (const { message } of envelopes) {
+      if (!message) {
+        continue
+      }
+      try {
+        const invite = SealedInvitation.fromBytes(
+          b64Decode(message as unknown as string)
+        )
+        // Get the invitation (and cache the value) to ensure decryption works
+        invite.v1.getInvitation(bundle)
+        invites.push(invite)
+      } catch (e) {
+        console.warn('Error decoding invite', e)
+      }
+    }
+
+    return invites
+  }
+
+  async saveInvite(invite: SealedInvitation) {
+    const unsealed = await invite.v1.getInvitation(this.keysV2)
+    const header = invite.v1.header
+    const myBundle = this.keysV2.getPublicKeyBundle()
+    const isSender = header.recipient.equals(myBundle)
+    const isRecipient = header.recipient.equals(myBundle)
+    if (!isSender && !isRecipient) {
+      throw new Error(
+        'Current user is neither the sender nor the recipient of this bundle'
+      )
+    }
+    this.topicKeyManager.addDirectMessageTopic(
+      unsealed.topic,
+      {
+        keyMaterial: unsealed.aes256GcmHkdfSha256.keyMaterial,
+        encryptionAlgorithm: EncryptionAlgorithm.AES_256_GCM_HKDF_SHA_256,
+        allowedSigners: [header.recipient, header.sender],
+      },
+      isSender ? header.recipient : header.sender,
+      nsLongToDate(header.createdNs)
+    )
   }
 }
 
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -1,5 +1,5 @@
 import { PrivateKeyBundleV2 } from './crypto/PrivateKeyBundle'
-import { SealedInvitation } from './Invitation'
+import { InvitationContext, SealedInvitation } from './Invitation'
 import { PublicKeyBundle, PrivateKeyBundleV1 } from './crypto'
 import Message from './Message'
 import {
@@ -32,7 +32,11 @@ import { xmtpEnvelope, messageApi, fetcher } from '@xmtp/proto'
 import { decodeContactBundle } from './ContactBundle'
 import ApiClient, { SortDirection } from './ApiClient'
 import { Authenticator } from './authn'
-import TopicKeyManager, { EncryptionAlgorithm } from './TopicKeyManager'
+import TopicKeyManager, {
+  DuplicateTopicError,
+  EncryptionAlgorithm,
+} from './TopicKeyManager'
+import Long from 'long'
 const { Compression } = xmtpEnvelope
 const { b64Decode } = fetcher
 
@@ -40,7 +44,8 @@ const { b64Decode } = fetcher
 // eslint-disable @typescript-eslint/no-explicit-any
 
 // Default maximum allowed content size
-const MaxContentSize = 100 * 1024 * 1024 // 100M
+const MAX_CONTENT_SIZE = 100 * 1024 * 1024 // 100M
+const INVITE_LOAD_OFFSET = 30
 
 export const ApiUrls = {
   local: 'http://localhost:5555',
@@ -121,7 +126,7 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
     env: 'dev',
     apiUrl: undefined,
     codecs: [new TextCodec()],
-    maxContentSize: MaxContentSize,
+    maxContentSize: MAX_CONTENT_SIZE,
   }
   if (opts?.codecs) {
     opts.codecs = _defaultOptions.codecs.concat(opts.codecs)
@@ -144,6 +149,7 @@ export default class Client {
   private contacts: Set<string> // address which we have connected to
   private topicKeyManager: TopicKeyManager
   private knownPublicKeyBundles: Map<string, PublicKeyBundle> // addresses and key bundles that we have witnessed
+  private latestInviteNs: Long
   private _conversations: Conversations
   private _codecs: Map<string, ContentCodec<any>>
   private _maxContentSize: number
@@ -156,9 +162,10 @@ export default class Client {
     this.address = keys.identityKey.publicKey.walletSignatureAddress()
     this._conversations = new Conversations(this)
     this._codecs = new Map()
-    this._maxContentSize = MaxContentSize
+    this._maxContentSize = MAX_CONTENT_SIZE
     this.apiClient = apiClient
     this.topicKeyManager = new TopicKeyManager()
+    this.latestInviteNs = new Long(0)
   }
 
   /**
@@ -211,10 +218,12 @@ export default class Client {
   // publish the key bundle into the contact topic
   private async publishUserContact(): Promise<void> {
     const pub = this.keys.getPublicKeyBundle()
-    await this.publishEnvelope({
-      contentTopic: buildUserContactTopic(this.address),
-      message: pub.toBytes(),
-    })
+    await this.publishEnvelopes([
+      {
+        contentTopic: buildUserContactTopic(this.address),
+        message: pub.toBytes(),
+      },
+    ])
   }
 
   /**
@@ -296,17 +305,35 @@ export default class Client {
 
     await Promise.all(
       topics.map(async (topic) => {
-        return this.publishEnvelope({
-          contentTopic: topic,
-          message: msgBytes,
-        })
+        return this.publishEnvelopes([
+          {
+            contentTopic: topic,
+            message: msgBytes,
+          },
+        ])
       })
     )
 
     return this.decodeMessage(msgBytes, topics[0])
   }
 
-  async publishEnvelope(env: messageApi.Envelope): Promise<void> {
+  async publishEnvelopes(envelopes: messageApi.Envelope[]): Promise<void> {
+    for (const env of envelopes) {
+      this.validateEnvelope(env)
+    }
+    try {
+      await this.apiClient.publish(
+        envelopes.map((env) => ({
+          contentTopic: env.contentTopic!, // eslint-disable-line @typescript-eslint/no-non-null-assertion
+          message: env.message!, // eslint-disable-line @typescript-eslint/no-non-null-assertion
+        }))
+      )
+    } catch (err) {
+      console.log(err)
+    }
+  }
+
+  private validateEnvelope(env: messageApi.Envelope): void {
     const bytes = env.message
     if (!env.contentTopic) {
       throw new Error('Missing content topic')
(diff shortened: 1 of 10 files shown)


src/Client.ts
```diff
@@ -135,9 +147,12 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
 export default class Client {
   address: string
   keys: PrivateKeyBundleV1
+  signedKeys: PrivateKeyBundleV2
```
neekolas, line 150: I don't love this, but it feels preferable to doing the `fromLegacyBundle` conversion in every method

src/Client.ts
```diff
@@ -288,22 +308,31 @@ export default class Client {
...
 
     return this.decodeMessage(msgBytes, topics[0])
   }
 
-  async publishEnvelope(env: {
-    contentTopic: string
-    message: Uint8Array
-    timestamp?: Date
-  }): Promise<void> {
+  async publishEnvelopes(envelopes: PublishParams[]): Promise<void> {
```
neekolas, line 324: Converted this to always accept an array of envelopes so we can send the invite to both the sender and recipient topics in one API request.

src/conversations/Conversation.ts
```diff
@@ -13,11 +13,20 @@ import Message from '../Message'
...
-  constructor(client: Client, address: string) {
+  constructor(
+    client: Client,
+    address: string,
+    topics: string[],
+    context?: InvitationContext
+  ) {
     this.peerAddress = address
     this.client = client
+    this.topics = topics
```
neekolas, line 28: Conversations have the topics explicitly defined now. Can be multiple topics.

Haven't gotten around to rewriting the message reading/writing functions to use the topic array yet, but that should be pretty straightforward.

src/conversations/Conversations.ts
```diff
@@ -24,6 +24,43 @@ export default class Conversations {
...
    */
   async list(): Promise<Conversation[]> {
+    const v1Addresses = await this.getV1Addresses()
+    const topicResults = this.client.allTopics()
+    const out: Conversation[] = []
+    for (const result of topicResults) {
+      // If no context, attempt to join with existing V1 intro
+      const topics = [result.contentTopic]
+      if (!result.context) {
+        if (v1Addresses.has(result.peerAddress)) {
```
neekolas, line 34: In cases where there is both a V1 and a V2 invite for the same address, and the V2 invite has no context, merge them into a single conversation. This is the migration path for people to seamlessly be able to continue an existing conversation started from before the release of this feature.

I definitely prefer doing this merging for developers, rather than having both an old and a new conversation returned in the list and having the developer have to somehow know to merge them themselves, and also be aware that they should only write to the V2 conversation.

galligan:
@neekolas just for curiosity's sake: would it be possible to have a persistent file that could be stored for a user that might be more expressive and have more options to query? That is, as a replacement to the intro topic being the only place where the conversation list is fetched?

mkobetic:
I have yet to go through this in more detail but it seems we're working on similar things. Let me share what I've got at this point, very early WIP, but maybe showing both can help us get on the same page #185. I've mostly just spent some time thinking about how to weave the new stuff into the conversation API, made the same thing re private key bundles, just called the old one legacyKeys for now, etc. You are much further ahead on the invitation front, I wonder how we can collaborate better on this.

src/Client.ts
```diff
@@ -516,6 +533,116 @@ export default class Client {
...
+    )
+  }
+
+  async sendInvite(peerAddress: string, context?: InvitationContext) {
+    const recipientContact = await this.getUserContact(peerAddress)
+    if (!recipientContact) {
+      throw new Error(`Recipient with address ${peerAddress} cannot be found`)
+    }
+    const created = new Date()
+    const recipient = SignedPublicKeyBundle.fromLegacyBundle(recipientContact)
```
mkobetic, line 621: I was planning to use published V2 contact bundle as a trigger for using V2 behavior.

src/conversations/Conversation.ts
mkobetic: So this is for stitching conversations from multiple invitations? Do we want to go there at this point?

src/conversations/Conversations.ts
mkobetic: I think we'll need to figure out what is the behavior of `newConversation` in this world. Are we stitching all invites with no context into one conversation?

mkobetic:
This is a very solid step forward. I do have some conceptual concerns but we can iron those out.

mkobetic:
So I think our main point of contention is conversation stitching. I'm worried about the complexity and potential edge cases we might be running into. You think whatever complexity we run into is worth the developer convenience. Let me just summarize some of the questions raised by the added complexity that I see at this point.

Let's assume that the stitching is governed by relatively simple criteria: Conversations are stitched based strictly on the conversation_id (in the invitation context). Invitations with the same conversation_id belong to the same conversation. That includes missing conversation_id. V1 conversation is the same as missing conversation_id. Missing context is the same as missing conversation_id.

* Client.newConversation - I guess the most sensible setup is that it takes the invitation context as parameter. If you want to create an invitation that continues previous one, you pass the same context into newConversation (or maybe we move this to the Conversation class?). If you want a new conversation you pass Context with conversation_id set to null? If you want a conversation with missing conversation_id you set it to ???
 
* Conversations.stream() - this will now have to have a side-effect of updating preexisting conversations with new invitations, without returning anything. 

* Conversation.streamMessages() - This stream will now have to subscribe to invitations too, in order to pick up new invitations for this conversation.

* Conversation.list() - do we sort the messages purely by timestamp or by topic first then timestamp?

* Conversation.send() - we write to the topic of the most recent invitation, if that invitation is V1, we will transparently create a continuing V2 invitation and write there.

It's possible that we can create reasonably sound model here, but the added complexity is non-trivial. How many use cases will actually want this behavior. What if we put aside the transient benefit of being able to transparently stitch v1 and v2 conversation into one?

mkobetic:
This can get even messier with group conversation context. If you don't have a list of all the participants, how do you send out an invitation to start a new segment of the conversation? If not everyone has the new invitation, what if some keep writing to older segments of the conversation?

mkobetic:
There's also a race condition between one party sending a new invitation for a pre-existing conversation, while the other party writes to the old topic still. So we'll need to handle messages showing up in previous topic as well.

mkobetic:
This might be worst of both, but could it make sense to make stitching optional somehow. E.g. no conversation_id means no stitching. We could make V1 conversation to be interpreted as conversation_id = 'legacy' or something. So if you wanted to continue that you'd use that ID. We could also say that stitching doesn't work for group conversations. Would that make anything better?

Alternatively, is it possible to implement stitching as a layer above the simpler invitation == conversation model?

src/conversations/Conversation.ts
neekolas: I've been really designing this around stitching from 1 invitation and one intro message, which is a good place to start. Considering we don't have topic rotation, that will be the only use-case we support to start

neekolas:
> @galligan would it be possible to have a persistent file that could be stored for a user that might be more expressive and have more options to query? That is, as a replacement to the intro topic being the only place where the conversation list is fetched?

So, the trick to the invite topic is that it needs to have the same data for both participants so they can communicate with the same keys on the same topic. So, a file on its own wouldn't work. If we had some sort of out-of-band protocol to do invites (maybe a QR code) both sides could store the same file.

neekolas:
This is exactly the conversation I was hoping to start @mkobetic 

> So I think our main point of contention is conversation stitching. 

FWIW, this PR isn't really proposing any stitching beyond stitching together the legacy conversations and new ones. In practice, what I'm proposing would have at-most 2 topics per conversation. Maybe we also want topic rotation, but I think that's a whole other discussion.

> Conversations.stream() - this will now have to have a side-effect of updating preexisting conversations with new invitations, without returning anything.

So, that's a choice. We could architect it so that we check for an existing conversation, and don't emit anything when it's an update

> Conversation.streamMessages() - This stream will now have to subscribe to invitations too, in order to pick up new invitations for this conversation. Does it subscribe to all topics of the conversation or only the last one?

One of the things I've been thinking about is whether or not we just want to be constantly streaming invites once you load up the SDK. It would remove a lot of the complicated race conditions here, although we'd have to do some work to make it reliable (for example, filling in the gaps when a connection is lost).

If we didn't do that, we would have to either listen to the invite topic while streaming messages or be OK with sending to an old topic

> Conversation.list() - do we sort the messages purely by timestamp or by topic first then timestamp?

Purely by timestamp. The user shouldn't care what topic a message was sent on

> Conversation.send() - we write to the topic of the most recent invitation, if that invitation is V1, we will transparently create a continuing V2 invitation and write there.

I think that's the right developer experience

> Do we add new API to observe Invitations rather than Conversations?

Maybe. It requires a little more explaining to developers about when you should use one versus the other. On the other hand, it means that we don't have to be loading/listening to legacy topics until the end of time.

> There's also a race condition between one party sending a new invitation for a pre-existing conversation, while the other party writes to the old topic still. So we'll need to handle messages showing up in previous topic as well.

I think we'll want to accept messages from all topics in a convo. 

> is it possible to implement stitching as a layer above the simpler invitation == conversation model?

I have the same worry about this as I do about doing nothing and leaving it to developers. If conversations are scoped to one topic, list results are going to have to get stitched together. That has its own complications
- More requests to our API
- Breaks pagination and limit queries
- Have to re-sort the messages once merged, since each conversation will be sorted independently

mkobetic:
> This is exactly the conversation I was hoping to start @mkobetic

❤️ 

> > So I think our main point of contention is conversation stitching.
> 
> FWIW, this PR isn't really proposing any stitching beyond stitching together the legacy conversations and new ones. In practice, what I'm proposing would have at-most 2 topics per conversation. Maybe we also want topic rotation, but I think that's a whole other discussion.

So what would the exact stitching rule be in this smaller case? If there's an invite with no convo id and an introduction, then stitch them together? What if there is more than one such invite?

I think that even if we intend to implement just the simple legacy stitching for now, the underlying model `topics[]` indicates something more than that.

> > Conversation.streamMessages() - This stream will now have to subscribe to invitations too, in order to pick up new invitations for this conversation. Does it subscribe to all topics of the conversation or only the last one?
> 
> One of the things I've been thinking about is whether or not we just want to be constantly streaming invites once you load up the SDK. It would remove a lot of the complicated race conditions here, although we'd have to do some work to make it reliable (for example, filling in the gaps when a connection is lost).

Yup, this is probably the way to go if we wanted to support stitching properly.  Again more complexity in the SDK.

> > is it possible to implement stitching as a layer above the simpler invitation == conversation model?
> 
> I have the same worry about this as I do about doing nothing and leaving it to developers. If conversations are scoped to one topic, list results are going to have to get stitched together. That has its own complications
> 
> * More requests to our API
> * Breaks pagination and limit queries
> * Have to re-sort the messages once merged, since each conversation will be sorted independently

I'm not sure I quite follow the argument here, but this sounds like an argument to do the full stitching from the get go, not just the simpler legacy stuff.

My main concern with doing the stitching at the SDK level assumes that there is one right way to do it and everyone will be happy with it. Is there such a model? What exactly is it? I do think we'll have to think it through beyond just the simple legacy thing and also in the context of group chats. We don't want to change the semantics after we ship this.

src/TopicKeyManager.ts
```diff
@@ -1,38 +1,44 @@
...
  * TopicResult is the public interface for receiving a TopicKey
  */
 export type TopicResult = {
-  topicKey: TopicKeyRecord
+  topicKey: TopicKey
   contentTopic: string
+  // Callers should validate that the signature comes from the list of allowed signers
+  // Not strictly necessary, but it prevents against compromised topic keys being
+  // used by third parties who would sign the message with a different key
+  allowedSigners: SignedPublicKeyBundle[]
```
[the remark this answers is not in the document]
mkobetic: Knowing full membership will be tricky. We have no support for that in the model now. If we wanted that, then we should probably include full membership list in the invite. If someone wants to add to the group, they will have send the invite to everyone in that list.

neekolas:
@mkobetic 

> So what would the exact stitching rule be in this smaller case? If there's an invite with no convo id and an introduction, then stitch them together? What if there is more than one such invite?

I think so. We would try and do everything we can to avoid more than one invite, but we can't prevent that entirely. I think in the case that someone (either through a race condition or misusing the SDK) had multiple valid invites with no convo id we would include that too, and use the most recent invite for sending.

> I'm not sure I quite follow the argument here, but this sounds like an argument to do the full stitching from the get go, not just the simpler legacy stuff.

What I think the minimum viable amount of stitching is legacy + invite, and no topic rotation. In that case we are stitching at most 2 topics.

> My main concern with doing the stitching at the SDK level is that it assumes that there is one right way to do it and everyone will be happy with it. Is there such a model? What exactly is it? I do think we'll have to think it through beyond just the simple legacy thing and also in the context of group chats. We don't want to change the semantics after we ship this.

My feeling is that in a world where we aren't doing topic rotation or mutable metadata, everyone is going to be happy with the legacy + invite flow, since it's just what you would expect. I want to talk directly to a wallet and I get all the messages in the history of the conversation when I query for `messages()`. That feels pretty intuitive. If we aren't doing topic rotation to start, we avoid some of the hairier use-cases. We would just strong-arm developers to use the find-or-create method of continuing conversations to avoid an explosion of invites.

mkobetic:
> > So what would the exact stitching rule be in this smaller case? If there's an invite with no convo id and an introduction, then stitch them together? What if there is more than one such invite?
> 
> I think so. We would try and do everything we can to avoid more than one invite, but we can't prevent that entirely. In the case that someone (either through a race condition or misusing the SDK) had multiple valid invites with no convo id we would include that too, and use the most recent invite for sending.

So we would create a no-id invite transparently when they want to send to the old convo and otherwise force a new convo id on all user created invites?

neekolas:
> So we would create a no-id invite transparently when they want to send to the old convo

I think so. That way we can migrate users to the new format and one day deprecate intros. That requires all conversations to move over to the invite topic. 

> otherwise force a new convo id on all user created invites?

The ID field would be optional or have a default value. I still don't know what makes the most sense there.

mkobetic:
> > otherwise force a new convo id on all user created invites?
> 
> The ID field would be optional or have a default value. I still don't know what makes the most sense there.

I'm thinking let's not give user direct control over the ids at all, instead `newConversation` would generate a new ID, whereas `conversation.renew(): Conversation` would copy the ID from the old one. This way users can create conversation chains linked with the same ID and do whatever they want with that.

neekolas:
I'm not sure I follow. How would automatic IDs work with use-cases like Lens Profile <> Profile messaging?

neekolas:
> This way users can create conversation chains linked with the same ID and do whatever they want with that.

What do you imagine them doing with the conversation chains other than querying all of them and merging the results together? If you want the results to be used separately, why would the conversations share an ID?

PR #185, opened 2022-10-07
V2 Conversations

mkobetic:
## TODO
* [ ] acceptable test coverage of new features
* [x] replace unions with proper interfaces
* [x] make sure conversation_ids are scoped to peerAddress
* [x] squash commits and properly mark as BREAKING CHANGE to trigger major version increment
* [x] rebase on 'beta' branch

## Changes

* Client maintains both V1 and V2 private key bundles (`keys` vs `legacyKeys`), still publishing V1 only
* Client supports  reading both V1 and V2 peer contact bundles
* Client keeps publishing V1 contact bundle by default (tests can make it publish V2 contact)
* MessageV1/V2 classes to support header differences
* Both Message classes support `senderAddress()` & `sent`; `recipientAddress` not supported by V2 and should be abandoned (we could make it work in 1:1 convos if we link messages back to their conversations, but it won't work in group chat context)
* Stream generalized to support both message versions (streamAllMessages yet to be fixed in a follow up)
* ConversationV1/V2 classes to support different semantics
* ConversationV2 implements MessageV2 encoding/decoding (Client retains V1 version of that)
* listing/streaming conversations reads both the intro and invite topics and creates both kinds of conversations
* creating V2 conversation sends invites to both the sender and the recipient
* see below for the expanded behavior of `newConversation`
 
## Breaking changes

* `Message` is now `MessageV1 | MessageV2`
* `Conversation` is now `ConversationV1 | ConversationV2`
* Many of the `Client` methods are now restricted to `MessageV1` only as they are only pertinent to the V1 protocol; analogous V2 functionality can be found on `Conversation(s)`
* `Client.publishEnvelope()` is replaced by `Client.publishEnvelopes()` (not meant for general use)

## V1 vs V2 conversation rules

1.  V1 conversations will continue to be supported in read/write mode.
2. `newConversation` behavior
  a. if conversation_id defined => V2 conversation
  b. if peer contact v1 => V1 conversation - (to support old peer clients)
  c. if pre-existing v1 convo => V1 conversation - (to avoid splitting pre-existing conversations) 
  d. otherwise V2 conversation, but
    1. if v2 convo with the same conversation_id exists => continue that convo (no new invites)
    2. otherwise create a new V2 convo with the new ID

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -1,12 +1,19 @@
-import { PublicKeyBundle, PrivateKeyBundleV1 } from './crypto'
-import Message from './Message'
+import {
+  PublicKeyBundle,
+  SignedPublicKeyBundle,
+  PrivateKeyBundleV1,
+  PrivateKeyBundleV2,
+} from './crypto'
+import { MessageV1, MessageV2 } from './Message'
 import {
   buildDirectMessageTopic,
   buildUserContactTopic,
   buildUserIntroTopic,
   mapPaginatedStream,
+  EnvelopeMapper,
+  buildUserInviteTopic,
 } from './utils'
-import Stream, { MessageFilter, noTransformation } from './Stream'
+import Stream from './Stream'
 import { Signer } from 'ethers'
 import {
   EncryptedKeyStore,
@@ -25,9 +32,10 @@ import {
 } from './MessageContent'
 import { decompress, compress } from './Compression'
 import { xmtpEnvelope, messageApi, fetcher } from '@xmtp/proto'
-import { decodeContactBundle } from './ContactBundle'
-import ApiClient, { SortDirection } from './ApiClient'
+import { decodeContactBundle, encodeContactBundle } from './ContactBundle'
+import ApiClient, { PublishParams, SortDirection } from './ApiClient'
 import { Authenticator } from './authn'
+import { SealedInvitation } from './Invitation'
 const { Compression } = xmtpEnvelope
 const { b64Decode } = fetcher
 
@@ -86,6 +94,7 @@ type NetworkOptions = {
 
 type ContentOptions = {
   // Allow configuring codecs for additional content types
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   codecs: ContentCodec<any>[]
 
   // Set the maximum content size in bytes that is allowed by the Client.
@@ -134,18 +143,28 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
  */
 export default class Client {
   address: string
-  keys: PrivateKeyBundleV1
+  legacyKeys: PrivateKeyBundleV1
+  keys: PrivateKeyBundleV2
   apiClient: ApiClient
   private contacts: Set<string> // address which we have connected to
-  private knownPublicKeyBundles: Map<string, PublicKeyBundle> // addresses and key bundles that we have witnessed
+  private knownPublicKeyBundles: Map<
+    string,
+    PublicKeyBundle | SignedPublicKeyBundle
+  > // addresses and key bundles that we have witnessed
+
   private _conversations: Conversations
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private _codecs: Map<string, ContentCodec<any>>
   private _maxContentSize: number
 
   constructor(keys: PrivateKeyBundleV1, apiClient: ApiClient) {
     this.contacts = new Set<string>()
-    this.knownPublicKeyBundles = new Map<string, PublicKeyBundle>()
-    this.keys = keys
+    this.knownPublicKeyBundles = new Map<
+      string,
+      PublicKeyBundle | SignedPublicKeyBundle
+    >()
+    this.legacyKeys = keys
+    this.keys = PrivateKeyBundleV2.fromLegacyBundle(keys)
     this.address = keys.identityKey.publicKey.walletSignatureAddress()
     this._conversations = new Conversations(this)
     this._codecs = new Map()
@@ -184,7 +203,7 @@ export default class Client {
     opts?: Partial<ClientOptions>
   ): Promise<Uint8Array> {
     const client = await Client.create(wallet, opts)
-    return client.keys.encode()
+    return client.legacyKeys.encode()
   }
 
   async init(options: ClientOptions): Promise<void> {
@@ -201,12 +220,16 @@ export default class Client {
   }
 
   // publish the key bundle into the contact topic
-  private async publishUserContact(): Promise<void> {
-    const pub = this.keys.getPublicKeyBundle()
-    await this.publishEnvelope({
-      contentTopic: buildUserContactTopic(this.address),
-      message: pub.toBytes(),
-    })
+  // WARNING: temporarily public to allow testing negotiated topics
+  // TODO: make private again asap
+  async publishUserContact(legacy = true): Promise<void> {
+    const keys = legacy ? this.legacyKeys : this.keys
+    await this.publishEnvelopes([
+      {
+        contentTopic: buildUserContactTopic(this.address),
+        message: encodeContactBundle(keys.getPublicKeyBundle()),
+      },
+    ])
   }
 
   /**
@@ -216,7 +239,7 @@ export default class Client {
 
   async getUserContact(
     peerAddress: string
-  ): Promise<PublicKeyBundle | undefined> {
+  ): Promise<PublicKeyBundle | SignedPublicKeyBundle | undefined> {
     const existingBundle = this.knownPublicKeyBundles.get(peerAddress)
 
     if (existingBundle) {
@@ -233,6 +256,10 @@ export default class Client {
     }
 
     return newBundle
+  }
+
+  forgetContact(peerAddress: string) {
+    this.knownPublicKeyBundles.delete(peerAddress)
   }
 
   /**
@@ -261,13 +288,17 @@ export default class Client {
    */
   async sendMessage(
     peerAddress: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     content: any,
     options?: SendOptions
-  ): Promise<Message> {
+  ): Promise<MessageV1> {
     let topics: string[]
     const recipient = await this.getUserContact(peerAddress)
     if (!recipient) {
       throw new Error(`recipient ${peerAddress} is not registered`)
+    }
+    if (!(recipient instanceof PublicKeyBundle)) {
+      throw new Error(`recipient bundle is not legacy bundle`)
     }
 
     if (!this.contacts.has(peerAddress)) {
@@ -288,22 +319,20 @@ export default class Client {
 
     await Promise.all(
       topics.map(async (topic) => {
-        return this.publishEnvelope({
-          contentTopic: topic,
-          message: msgBytes,
-          timestamp,
-        })
+        return this.publishEnvelopes([
+          {
+            contentTopic: topic,
+            message: msgBytes,
+            timestamp,
+          },
+        ])
       })
     )
 
     return this.decodeMessage(msgBytes, topics[0])
   }
 
-  async publishEnvelope(env: {
-    contentTopic: string
-    message: Uint8Array
-    timestamp?: Date
-  }): Promise<void> {
+  private validateEnvelope(env: PublishParams): void {
     const bytes = env.message
     if (!env.contentTopic) {
       throw new Error('Missing content topic')
@@ -312,26 +341,27 @@ export default class Client {
     if (!bytes || !bytes.length) {
       throw new Error('Cannot publish empty message')
     }
+  }
 
+  async publishEnvelopes(envelopes: PublishParams[]): Promise<void> {
+    for (const env of envelopes) {
+      this.validateEnvelope(env)
+    }
     try {
-      await this.apiClient.publish([
-        {
-          contentTopic: env.contentTopic,
-          message: bytes,
-          timestamp: env.timestamp,
-        },
-      ])
+      await this.apiClient.publish(envelopes)
     } catch (err) {
       console.log(err)
     }
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   registerCodec(codec: ContentCodec<any>): void {
     const id = codec.contentType
     const key = `${id.authorityId}/${id.typeId}`
     this._codecs.set(key, codec)
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   codecFor(contentType: ContentTypeId): ContentCodec<any> | undefined {
     const key = `${contentType.authorityId}/${contentType.typeId}`
     const codec = this._codecs.get(key)
@@ -344,12 +374,11 @@ export default class Client {
     return codec
   }
 
-  async encodeMessage(
-    recipient: PublicKeyBundle,
-    timestamp: Date,
+  async encodeContent(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     content: any,
     options?: SendOptions
-  ): Promise<Message> {
+  ): Promise<Uint8Array> {
     const contentType = options?.contentType || ContentTypeText
     const codec = this.codecFor(contentType)
     if (!codec) {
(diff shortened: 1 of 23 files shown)


src/Client.ts
```diff
@@ -134,18 +140,27 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
  */
 export default class Client {
   address: string
-  keys: PrivateKeyBundleV1
+  legacyKeys: PrivateKeyBundleV1
```
mkobetic, line 143: keys -> legacyKeys, add keys as the V2 bundle

src/Client.ts
```diff
@@ -134,18 +140,27 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
...
+  keys: PrivateKeyBundleV2
   apiClient: ApiClient
   private contacts: Set<string> // address which we have connected to
-  private knownPublicKeyBundles: Map<string, PublicKeyBundle> // addresses and key bundles that we have witnessed
+  private knownPublicKeyBundles: Map<
+    string,
+    PublicKeyBundle | SignedPublicKeyBundle
+  > // addresses and key bundles that we have witnessed
```
mkobetic, line 150: peer keys now can be V1 or V2, this will be used as a signal to conduct v1 or v2 conversation.

src/Client.ts
```diff
@@ -201,11 +216,13 @@ export default class Client {
...
-  private async publishUserContact(): Promise<void> {
-    const pub = this.keys.getPublicKeyBundle()
+  // WARNING: temporarily public to allow testing negotiated topics
+  // TODO: make private again asap
+  async publishUserContact(legacy = true): Promise<void> {
+    const keys = legacy ? this.legacyKeys : this.keys
     await this.publishEnvelope({
       contentTopic: buildUserContactTopic(this.address),
-      message: pub.toBytes(),
+      message: encodeContactBundle(keys.getPublicKeyBundle()),
```
mkobetic, line 225: Apparently we've still been publishing bare key bundle instead of contact bundle.

src/Client.ts
```diff
@@ -263,12 +280,15 @@ export default class Client {
     peerAddress: string,
     content: any,
     options?: SendOptions
-  ): Promise<Message> {
+  ): Promise<MessageV1> {
```
mkobetic, line 283: The client API remains V1, the V2 api is moving to Conversation.

src/Client.ts
```diff
@@ -344,12 +364,10 @@ export default class Client {
     return codec
   }
 
-  async encodeMessage(
-    recipient: PublicKeyBundle,
-    timestamp: Date,
+  async encodeContent(
```
mkobetic, line 367: extracting the content encoding bits to standalone methods so that they can be reused for V2.

src/Client.ts
```diff
@@ -455,66 +500,88 @@ export default class Client {
   listMessagesPaginated(
     contentTopics: string[],
     opts?: ListMessagesPaginatedOptions
-  ): AsyncGenerator<Message[]> {
+  ): AsyncGenerator<MessageV1[]> {
     const topicFilter = filterForTopics(contentTopics)
-    return mapPaginatedStream(
```
mkobetic, line 460: extracting listing and paginated listing of envelopes, so that it can be reused for V2.

src/Invitation.ts
```diff
@@ -11,9 +11,14 @@ import { dateToNs } from './utils'
  */
 export class InvitationV1 implements invitation.InvitationV1 {
   topic: string
+  context: invitation.InvitationV1_Context | undefined
   aes256GcmHkdfSha256: invitation.InvitationV1_Aes256gcmHkdfsha256 // eslint-disable-line camelcase
 
-  constructor({ topic, aes256GcmHkdfSha256 }: invitation.InvitationV1) {
+  constructor({
+    topic,
+    context,
```
mkobetic, line 19: Need context to be able to use proto@v3.4.0

src/Message.ts
```diff
@@ -5,26 +5,32 @@ import {
...
+  }
+  throw new Error('unknown message version')
 }
 
 // Message is basic unit of communication on the network.
-// Message header carries the sender and recipient keys used to protect message.
 // Message timestamp is set by the sender.
-export default class Message implements proto.MessageV1 {
-  header: proto.MessageHeaderV1 // eslint-disable-line camelcase
+class Message {
```
mkobetic, line 33: Message is now abstract superclass of the two versions of message.

src/Message.ts
```diff
@@ -42,45 +48,50 @@ export default class Message implements proto.MessageV1 {
...
     header: proto.MessageHeaderV1,
     bytes: Uint8Array
-  ): Promise<Message> {
+  ): Promise<MessageV1> {
     const id = bytesToHex(await sha256(bytes))
-    return new Message(id, bytes, obj, header)
+    return new MessageV1(id, bytes, obj, header)
   }
 
-  static async fromBytes(bytes: Uint8Array): Promise<Message> {
```
mkobetic, line 75: This seems unused, so ripping it out.

src/Message.ts
```diff
@@ -103,13 +114,13 @@ export default class Message implements proto.MessageV1 {
     ).walletSignatureAddress()
   }
 
-  // encrypt and serialize the message
+  // LEGACY: encrypt and serialize the message
```
mkobetic, line 117: This is used by the Client API so it's only on MessageV1

src/Message.ts
```diff
@@ -193,7 +200,48 @@ export default class Message implements proto.MessageV1 {
...
+    const id = bytesToHex(await sha256(bytes))
+    return new MessageV2(id, bytes, obj, header, signed)
+  }
+
+  get sent(): Date {
+    return nsToDate(this.header.createdNs)
+  }
+
+  // wallet address derived from the signature of the message sender
+  get senderAddress(): Promise<string> | undefined {
```
mkobetic, line 239: MessageV2 doesn't have recipientAddress(), we'll need to figure out how to do without.

src/Stream.ts
```diff
@@ -1,23 +1,25 @@
 import { UnsubscribeFn } from './ApiClient'
-import Message from './Message'
+import { MessageV1, MessageV2 } from './Message'
 import Client from './Client'
 import { messageApi, fetcher } from '@xmtp/proto'
 
-export type MessageTransformer<T> = (msg: Message) => T
+export type MessageDecoder<M> = (env: messageApi.Envelope) => Promise<M>
```
mkobetic, line 6: MessageDecoder is to abstract out the V1 specific env->msg decoding.
Consequently the other types need to take message type parameter to support both versions.

src/Stream.ts
```diff
@@ -1,23 +1,25 @@
...
-export type MessageFilter = (msg: Message) => boolean
+export type MessageTransformer<M, T> = (msg: M) => T
 
-export type ContentTopicUpdater = (msg: Message) => string[] | undefined
+export type MessageFilter<M> = (msg: M) => boolean
 
-export const noTransformation = (msg: Message): Message => {
+export type ContentTopicUpdater<M> = (msg: M) => string[] | undefined
+
+export const noTransformation = (msg: any) => {
```
mkobetic, line 14: Not sure how to avoid the any here, to provide an identity function for both versions.

src/Message.ts
neekolas: Seems like we could work it out if we had a reference to the invitation headers. Whichever party was not the signer must be the recipient.

src/conversations/Conversations.ts
```diff
@@ -36,34 +43,45 @@ export default class Conversations {
...
 
   /**
    * Returns a stream of any newly created conversations.
    * Will dedupe to not return the same conversation twice in the same stream.
    * Does not dedupe any other previously seen conversations
    */
-  stream(): Promise<Stream<Conversation>> {
-    const messageTransformer: MessageTransformer<Conversation> = (
-      msg: Message
+  async stream(): Promise<JoinStream<ConversationV1, ConversationV2>> {
```
neekolas, line 74: It seems like this would be more efficient to just stream from both topics, and have a switch in the decoding step that would either create a `ConversationV1` or a `ConversationV2` depending on the topic, rather than keeping 2 HTTP connections open

mkobetic: That sounds a lot better. This is quite a mess at the moment 👍

src/conversations/Conversation.ts
```diff
@@ -1,43 +1,62 @@
...
+  private _peerAddress: string
+  sent: Date
   private client: Client
 
-  constructor(client: Client, address: string) {
-    this.peerAddress = address
+  constructor(client: Client, address: string, sent: Date) {
+    this._peerAddress = address
     this.client = client
+    this.sent = sent
```
neekolas, line 32: What is the `sent` field used for?

mkobetic: To sort Conversations. Currently we sort by peerAddress, that's not going to be sufficient going forward.

mkobetic: https://github.com/xmtp/xmtp-js/blob/main/src/conversations/Conversations.ts#L43-L47

src/conversations/Conversations.ts
mkobetic: This turned out to be soooo much simpler! ❤️ 🙇 🙏

src/Message.ts
mkobetic: It won't work in the context of group conversations, so I was thinking of trying to stop using the concept altogether.

src/crypto/PublicKeyBundle.ts
```diff
@@ -41,7 +41,9 @@ export class SignedPublicKeyBundle implements publicKey.SignedPublicKeyBundle {
 
   static fromLegacyBundle(bundle: PublicKeyBundle): SignedPublicKeyBundle {
     return new SignedPublicKeyBundle({
-      identityKey: SignedPublicKey.fromLegacyKey(bundle.identityKey),
+      // Note: I am assuming all PublicKeyBundles passed into this have had their identity keys signed by a wallet
+      // Maybe that is not universally true in the future
+      identityKey: SignedPublicKey.fromLegacyKey(bundle.identityKey, true),
```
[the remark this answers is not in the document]
mkobetic: That's definitely how it should have been, my bad. I don't expect legacy bundles without wallet signatures.

test/conversations/Conversations.test.ts
```diff
@@ -148,4 +152,121 @@ describe('conversations', () => {
...
+    })
+
+    it('continues to use v1 conversation even after upgrading bundle', async () => {
+      const aliceConvo = await alice.conversations.newConversation(bob.address)
+      await aliceConvo.send('gm')
+      expect(aliceConvo instanceof ConversationV1).toBeTruthy()
+      await bob.publishUserContact(false)
+      await sleep(50)
+
+      const aliceConvo2 = await alice.conversations.newConversation(bob.address)
```
mkobetic, line 172: I think if you want to make sure alice picks up the new bundle for bob, you have to make her forget the old one (at least I did in my test). It might be good to add assertions for that.

test/conversations/Conversations.test.ts
```diff
@@ -148,4 +152,121 @@ describe('conversations', () => {
...
+
+    it('creates a v2 conversation when conversation ID is present', async () => {
+      const conversationId = 'xmtp.org/foo'
+      const aliceConvo = await alice.conversations.newConversation(
+        bob.address,
+        { conversationId, metadata: { foo: 'bar' } }
+      )
+      await sleep(50)
+
+      expect(aliceConvo instanceof ConversationV2)
```
mkobetic, line 192: Does this actually check the truthiness, or do you need to do invoke `toBeTruthy()` on it?

test/conversations/Conversations.test.ts
```diff
@@ -148,4 +152,121 @@ describe('conversations', () => {
...
+    })
+
+    it('creates a v2 conversation when conversation ID is present', async () => {
+      const conversationId = 'xmtp.org/foo'
+      const aliceConvo = await alice.conversations.newConversation(
+        bob.address,
+        { conversationId, metadata: { foo: 'bar' } }
+      )
+      await sleep(50)
```
mkobetic, line 190: I find 50 is often not enough when I run locally. 100 seems to do it, but it's really annoying having to pepper these tests with random sleeps. But I'm not sure how we could write the code that would wait for the message to be committed on the node side.

test/conversations/Conversations.test.ts
mkobetic: https://github.com/xmtp/xmtp-js/blob/dcafb40098b5ae82519b67f3924aa97c98463f72/test/conversations/Conversation.test.ts#L153-L159

test/conversations/Conversations.test.ts
neekolas: Oops. Fixed

neekolas:
Before merging this to the `beta` branch we should add the `BREAKING CHANGE:` commit, so that when we get the beta release it is on the 7.x version and not the 6.x

mkobetic:
Yup I definitely want to clean up the commits before merging. It's going to be easier to rebase squashed as well.

src/Message.ts
```diff
@@ -42,45 +47,43 @@ export default class Message implements proto.MessageV1 {
...
-    const msg = proto.Message.decode(bytes)
-    const innerMessage = extractV1Message(msg)
-    const header = proto.MessageHeaderV1.decode(innerMessage.headerBytes)
-    return Message.create(msg, header, bytes)
+    return new MessageV1(id, bytes, obj, header)
   }
 
-  get sent(): Date | undefined {
-    return this.header ? new Date(this.header?.timestamp.toNumber()) : undefined
+  get sent(): Date {
```
neekolas, line 85: Definitely a positive change here. That `Date | undefined` was always annoying as an application developer.

src/conversations/Conversation.ts
```diff
@@ -4,40 +4,61 @@ import Client, {
...
+  }
+
+  async peerAddress(): Promise<string> {
```
neekolas, line 37: Making this async is going to be an annoying breaking change for developers. This field gets used a lot, often in synchronous code (like rendering a UI). I'd suggest we make this a concrete property and do the async work for the ConversationV2 before the class is initialized.

src/conversations/Conversation.ts
```diff
@@ -46,7 +67,177 @@ export default class Conversation {
...
+    ) {
+      throw new Error('invalid signature')
+    }
+    const message = await MessageV2.create(msg, header, signed, env.message)
+    await this.client.decodeContent(message)
+    return message
   }
 }
+
+export type Conversation = ConversationV1 | ConversationV2
```
neekolas, line 236: I looked at the output type of this, and it's exactly what I would have written as an interface. Think it's fine to leave as-is

src/conversations/Conversation.ts
neekolas: Makes sense for the ordering.

Maybe it's just the name that's bothering me. I could imagine `conversation.sent` could be confusing for developers (is it the time the last message was sent? The time the first message was sent?). In reality, it's neither. For V1 it is the time that the first message was sent in the most recent session of either user, which is a really arbitrary and useless timestamp.

`createdAt` would be better, but it's not accurate for V1 since we are actually using the latest `intro` and not the first.

neekolas: What if we use the earliest seen timestamp instead of the latest seen timestamp in `getIntroductionPeers`. That would let us call it something like `createdAt` and would be a more meaningful value.

src/Message.ts
neekolas: Fair enough

src/conversations/Conversations.ts
```diff
@@ -24,50 +29,35 @@ export default class Conversations {
...
-      const peerAddress = this.getPeerAddress(message)
+    const conversations: Conversation[] = []
+    seenPeers.forEach((sent, peerAddress) =>
+      conversations.push(new ConversationV1(this.client, peerAddress, sent))
+    )
 
-      if (peerAddress) {
-        seenPeers.add(peerAddress)
-      }
+    for (const sealed of await this.client.listInvitations()) {
```
neekolas, line 39: We could use a `Promise.all()` to get the invitations and the intros in parallel. Would speed up this method a bit.

src/conversations/Conversation.ts
mkobetic: I think that makes sense. Do you want to make that change?

neekolas: Can do

src/conversations/Conversations.ts
mkobetic: Hm, not quite sure what you had in mind here. Do you mean something like this at the top?
```
  const [seenPeers, invitations] = await Promise.all([
      this.getIntroductionPeers(),
      this.client.listInvitations(),
    ])
```

neekolas (approved):

src/conversations/Conversations.ts
neekolas: Just implemented it

Issue #172, opened 2022-09-26
Handle addresses not in checksum format


saulmc:
### Context

ETH addresses use a special pattern of casing as a [checksum](https://coincodex.com/article/2078/ethereum-address-checksum-explained/) against typos.

`xmtp-js` expects addresses in this format. If you want to initiate a conversation with an address that has been stored in lowercase or uppercase format, you can use ethers' [`getAddress`](https://docs.ethers.io/v5/api/utils/address/#utils-getAddress) utility function to convert the address to checksum format.

### Problem

When developers attempt to initiate conversations with lowercased/uppercased addresses, `xmtp-js` fails to do so and doesn't explain how to correct the issue. This can be confusing for developers who haven't previously had to convert addresses to checksum format.

### Goal

Make it possible to initiate conversations with lowercased/uppercased addresses without having prior knowledge about converting addresses to checksum format.

### Potential Solutions

1. Run `getAddress` on addresses as a convenience to developers (preferred)
2. Return an informative error message when a developer supplies an address that isn't in checksum format
3. Specify `xmtp-js`'s checksum format requirement in the README

### Notes

- This issue has been noted by developers while using `conversations.newConversation` and `Client.canMessage`.

neekolas:
I'm in favour of #1. The performance impact should be negligible, and it's the nicest.

PR #187, opened 2022-10-12
Make get user contact case-safe

elisealix22:
Resolves https://github.com/xmtp/xmtp-js/issues/172

This PR updates `Client#getUserContact` to be case-safe so we can use `canMessage()` and `newConversation()` passing a lowercase address for example.


elisealix22:
Ugh, when will I remember to prefix my commits! Closing for a sec while I clean this up lol.

PR #188, opened 2022-10-12
feat: make peeraddress case-safe

elisealix22:
Resolves https://github.com/xmtp/xmtp-js/issues/172

This PR updates `Client#getUserContact` to be case-safe so we can use `canMessage()` and `newConversation()` passing a lowercase address for example.

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -7,7 +7,7 @@ import {
   mapPaginatedStream,
 } from './utils'
 import Stream, { MessageFilter, noTransformation } from './Stream'
-import { Signer } from 'ethers'
+import { ethers, Signer } from 'ethers'
 import {
   EncryptedKeyStore,
   KeyStore,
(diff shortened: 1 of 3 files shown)


src/Client.ts
```diff
@@ -213,23 +213,23 @@ export default class Client {
    * Returns the cached PublicKeyBundle if one is known for the given address or fetches
    * one from the network
    */
-
   async getUserContact(
     peerAddress: string
   ): Promise<PublicKeyBundle | undefined> {
-    const existingBundle = this.knownPublicKeyBundles.get(peerAddress)
+    const safeAddress = ethers.utils.getAddress(peerAddress)
```
neekolas, line 219: Maybe we want to move this to `utils.ts`

neekolas (approved):

elisealix22:
This is kinda frozen right now as we didn't end up needing it for Lenster and one of the test failures is legitimate in that if ethers errors, I will have to provide a fallback. Going to keep it open to revisit next week and handle that edge case / fix tests.

elisealix22:
Closing because it just makes more sense to start fresh on 7.0 on my next, more-focused attempt at this.

PR #190, opened 2022-10-19
Pass bytes as Uint8Array

neekolas:
## Summary

I noticed an issue where we were passing in the message bytes from the envelope as a string instead of a Uint8Array. This wasn't caught by the type checker, because the generated types for GRPC Gateway are always wrong for bytes fields. It also wasn't caught by the tests for some reason, even though the `v2 conversation` test does exercise this code. Maybe it's some quirk between server and client side `SubtleCrypto`?

But I have confirmed in the browser that this immediately fixes the error by using `yarn link`

### DIFF
--- a/src/conversations/Conversation.ts
+++ b/src/conversations/Conversation.ts
@@ -184,7 +184,8 @@ export class ConversationV2 {
     if (!env.message) {
       throw new Error('empty envelope')
     }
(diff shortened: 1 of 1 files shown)


elisealix22 (approved):
Nice catch!

mkobetic (approved):
This mismatch is so annoying, I spent good bit of time debugging this before I realized that the bytes from the decoded envelope is a b64 encoded string.

PR #191, opened 2022-10-19
streamAllMessages Fixes

neekolas:
## Summary

One of the missing pieces around negotiated topics is the ability to stream all messages, whether they are coming from V1 or V2 conversations.

The existing implementation would only stream V1 messages. Getting it to work with both had some interesting wrinkles.
- When an invite is received, it is already too late to update the topic list and stream the first message. So, in that case I query the newly found topic and yield any messages found. 
- Decoding V2 messages requires keeping track of the conversation it came from, since that is the only place with the key material. Cannot process with just the message alone like we can with V1
- When a message is found on the intro topic, need to both yield the message to the listener and also use it to update the topic list

I decided to switch this over to return an async generator, which no one should notice but us. Usage is unchanged. This lets us filter the `Stream` and ignore the `Invitation` messages, while still updating the topic list.

### DIFF
--- a/src/conversations/Conversation.ts
+++ b/src/conversations/Conversation.ts
@@ -14,7 +14,7 @@ import { messageApi, xmtpEnvelope, fetcher } from '@xmtp/proto'
 import { encrypt, decrypt, SignedPublicKey, Signature } from '../crypto'
 import Ciphertext from '../crypto/Ciphertext'
 import { sha256 } from '../crypto/encryption'
-import { dateToNs, nsToDate } from '../utils'
+import { buildDirectMessageTopic, dateToNs, nsToDate } from '../utils'
 const { b64Decode } = fetcher
 
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
@@ -39,6 +39,10 @@ export class ConversationV1 {
    */
   async messages(opts?: ListMessagesOptions): Promise<MessageV1[]> {
     return this.client.listConversationMessages(this.peerAddress, opts)
+  }
+
+  get topic(): string {
+    return buildDirectMessageTopic(this.peerAddress, this.client.address)
   }
 
   messagesPaginated(
@@ -180,8 +184,8 @@ export class ConversationV2 {
     return msg
   }
 
-  private async decodeMessage(env: messageApi.Envelope): Promise<MessageV2> {
-    if (!env.message) {
+  async decodeMessage(env: messageApi.Envelope): Promise<MessageV2> {
+    if (!env.message || !env.contentTopic) {
       throw new Error('empty envelope')
     }
     const messageBytes = b64Decode(env.message.toString())
@@ -210,6 +214,7 @@ export class ConversationV2 {
     ) {
       throw new Error('incomplete signed content')
     }
+
     const digest = await sha256(concat(msgv2.headerBytes, signed.payload))
     if (
       !new SignedPublicKey(signed.sender?.preKey).verify(
@@ -220,6 +225,7 @@ export class ConversationV2 {
       throw new Error('invalid signature')
     }
     const message = await MessageV2.create(msg, header, signed, messageBytes)
+    message.contentTopic = env.contentTopic
     await this.client.decodeContent(message)
     return message
   }
--- a/src/conversations/Conversations.ts
+++ b/src/conversations/Conversations.ts
@@ -169,22 +169,15 @@ export default class Conversations {
             : (msg.recipientAddress as string),
           msg.sent
         )
-
         const isNew = addConvo(convo.topic, convo)
-        if (!isNew) {
-          return undefined
-        }
 
-        return Array.from(topics.values())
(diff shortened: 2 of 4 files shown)


src/conversations/Conversation.ts
```diff
@@ -41,6 +41,10 @@ export class ConversationV1 {
     return this.client.listConversationMessages(this.peerAddress, opts)
   }
 
+  get topic(): string {
+    return buildDirectMessageTopic(this.peerAddress, this.client.address)
```
neekolas, line 45: Exposing the `topic` field on the V1 conversation, mostly for consistency's sake.

src/conversations/Conversations.ts
```diff
@@ -97,56 +97,115 @@ export default class Conversations {
...
+      decodeMessage,
       contentTopicUpdater
     )
+
+    return (async function* generate() {
+      for await (const val of str) {
+        if (val instanceof MessageV1 || val instanceof MessageV2) {
+          yield val
+        }
+        // For conversation V2, we know we have already missed the message when we get the invite
```
mkobetic, line 200: I think this statement is a bit oversimplified. With V2 the invite is actually sent earlier than with V1, we send invites from `newConversation` (V1 does nothing here), the first message is sent with `send` after that. That of course doesn't mean that we didn't miss the message by the time we subscribe to the topic. Same could easily be true with V1 convos, by the time we subscribe in response to the first message, the second, third, etc message could have already gone through.

I'm not quite sure what sort of guarantees is reasonable to offer here. But it's definitely more complicated than what we admit here.

src/conversations/Conversations.ts
```diff
@@ -97,56 +97,115 @@ export default class Conversations {
...
 
-    return Stream.create<MessageV1>(
+    const str = await Stream.create<Message | Conversation | null>(
       this.client,
-      topics,
-      async (env) => {
-        const msg = await this.client.decodeEnvelope(env)
-        return filter(msg) ? msg : undefined
-      },
+      Array.from(topics.values()),
```
mkobetic, line 190: I'm not sure how to make this better, but want to point out that, by the time we subscribe to all these topics, we could have missed new intros or invites. It might be possible to subscribe to invites/intros first, then list all pre-existing, and we may have to make sure we don't subscribe to the same topic twice. Sounds messy and I really like how clean this is. Is it worth making it more complicated?

src/conversations/Conversations.ts
```diff
@@ -97,56 +97,115 @@ export default class Conversations {
...
+
+        convoMap.set(convo.topic, convo)
+        topics.add(convo.topic)
+
+        return Array.from(topics.values())
       }
-      return true
+
+      if (msg instanceof ConversationV2) {
+        if (topics.has(msg.topic)) {
```
mkobetic, line 175: This is just a nitpick but 175-182 is the same as 164-171, is it worth combining?

mkobetic (approved):
I have few comments that we may or may not want to address here. Might be worth to at least add some of that as comments for future reference. But I really like how clean this is. I think going with the generator is the right approach.

src/conversations/Conversations.ts
neekolas: That's fair. We don't know for sure if we have missed anything, but we do know that if there is anything in the topic that we will likely not have captured it at this point.

I can clarify the comment

src/conversations/Conversations.ts
neekolas: I refactored to simplify

src/conversations/Conversations.ts
neekolas: This whole thing is full of possible race conditions.

I don't love having to do a resubscribe immediately after setting up the stream (every time we resubscribe we open up a new potential gap in the streams or potential for duplicate messages). Seems problematic no matter what.

neekolas:
I added some more comments to clarify the limitations.

src/conversations/Conversations.ts
```diff
@@ -97,6 +97,13 @@ export default class Conversations {
     )
   }
 
+  /**
+   * Streams messages from all conversations.
+   *
+   * When a new conversation is initiated with the client's address, this function will automatically register it and add it to the list of conversations to watch.
+   * Callers should be aware that in some cases the first messages in a newly created legacy conversation may not be picked up by this method.
+   * This will be resolved in a future release where all new conversations will be `ConversationV2`
```
mkobetic, line 105: There's still going to be the resubscription holes and the race at the beginning that could result in a conversation to be missed entirely. Do we want to make such a strong claim of this being fixed?

mkobetic: I'd be inclined to just call this best effort or something and generally try to steer people away from this.

neekolas: I updated to couch the language even further. 

The only reason I'm hesitant to steer people away entirely is because they don't have any tools to do a better job than us trying to implement the same thing.

Issue #178, opened 2022-09-29
V2 rollout


mkobetic:
V2 refers to a planned switch to "negotiated topics" that will enhance message and conversation privacy and allow carrying multiple independent conversations between parties.

# V2 Implementation
This just adds the support for all these things without actually employing them anywhere
* [x] signed keys and bundles #159, #166, #175, #176
* [x] invitations #177
* [x] message v2 #185

At this point there are no changes on the network, everything runs in v1 mode

# V1/V2 Transition
* [x] dual v1/v2 private key bundle support in the Client (based on V1 published bundles) #185
* [x] dual v1/v2 messaging support applied based on the version of the recipient's contact bundle #185
* [x] 2022-11-02 release v7.0.0 #208
* [x] 2022-11-08 release v7.1.0 #204
* [x] 2023-01-01 deprecate pre-v7.0.0 SDK 
* [ ] convert all contacts to v2

At this point we'll start seeing v2 activity on the network, v1 should slowly decrease as clients upgrade and users engage.
We need to get all the users to republish their contact bundles as v2 so that majority of new activity is v2. We might be able to do this ourselves (when all clients are upgraded), since we don't need the wallet for the v1->v2 contact bundle conversion. Some level of V1 activity will remain as pre-existing V1 conversations will continue, unless the participants explicitly start a new V2 conversation.

# V1 Deprecation
* [ ] start publishing V2 private key bundles

We will have to be careful here, users new to the platform that don't have any pre-existing conversations should be able to get by with native V2 keys only. However users that have pre-existing V1 conversation will have to keep using "V1 compatible" key bundles (i.e. bundles generated as V1 and converted to V2). Otherwise they won't be able to resume their pre-existing V1 conversations (see #222).

# V1 Cleanup

It's not clear at this point whether there will be anything to clean up. At this point we intend to maintain pre-existing V1 conversations going forward.

mkobetic:
In order to proceed with the rollout plan we need to get clients to version at least v6.4.0 in order for clients to be able to tolerate/ignore v2 contact bundles being published into contact topics (#166 has the fix).

We have a pending deprecation related to sunsetting the waku/libp2p based client interface on the node (in favor of the new GRPC based interface). The deadline we set is November 8th.

> “On November 8th, we plan to sunset the legacy APIs that power versions of the SDK < 6.0.0.“

[...]

PR #193, opened 2022-10-20
fix: add deprecation schedule

mkobetic:
We should have a more permanent record of our deprecation schedule, than just the Discord announcements. The README is one of the more prominent places where we could surface this. The idea here is that we'd list all deadlines in this table and remove them as they expire.

I'd be also inclined to establish an automatic deprecation of versions that are too old (a year maybe) that wouldn't be listed explicitly in this table, but the rule should be mentioned above the table here as well if we agree on something like that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,10 @@ Older versions of the SDK will eventually become deprecated, which means:
 
 Following table shows the deprecation schedule.
 
-| Announced | Effective | Minimum Version | Rationale
-| --- | --- | --- | ---
-| 2022-08-18 | 2022-11-08 | v6.0.0 | XMTP network will stop supporting the Waku/libp2p based client interface in favor of the new GRPC based interface
-| 2022-11-21 | 2022-11-08 | v6.4.0 | clients will start publishing V2 contact bundles to the network, pre-v6.4.0 clients may not be able to tolerate this activity ([details](https://github.com/xmtp/xmtp-js/issues/178#issuecomment-1285527263))
+| Announced  | Effective  | Minimum Version | Rationale                                                                                                                                                                                                     |
+| ---------- | ---------- | --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


mkobetic:
Here is how an automatic deprecation rule could look like:

`A major version of the SDK is automatically deprecated when there is a more recent major version of the SDK available for more than a year.`

neekolas:
Love including this in the README. Will help a lot.

I'm not sure how useful automatic deprecation is going to be in a world where we already are listing explicit versions and moving this quickly on the protocol. I think in practice, for the next year at least we are going to require more frequent updates through explicit migrations like the November 8th one. So automatic deprecation may give developers a false sense of security that they don't have to worry for a year (or whatever the time frame is).

Once the product matures a bit, and we have a period of relative stability of our protocol, automatic deprecation starts to make sense.

mkobetic:
> I'm not sure how useful automatic deprecation is going to be in a world where we already are listing explicit versions and moving this quickly on the protocol. I think in practice, for the next year at least we are going to require more frequent updates through explicit migrations like the November 8th one. So automatic deprecation may give developers a false sense of security that they don't have to worry for a year (or whatever the time frame is).
> 
> Once the product matures a bit, and we have a period of relative stability of our protocol, automatic deprecation starts to make sense.

That is a good point, as long as the table has dates on it, automatic deprecation is less important. I was focusing on the opposite case when the table empties out, don't give users false sense of being able to sit on old versions for a while.

I guess we should also indicate somewhere what the current minimum version is, can probably just put it right above the table.

mkobetic:
We've agreed to revise the existing 11/8 deadline to call for v6.4.0. We will make a new deprecation announcement in Discord today.

@jhaaaa/@saul can we put the new deprecation announcement somewhere where I can link to it from the table here? 🙏

neekolas (approved):

mkobetic:
In light of https://github.com/xmtp/xmtp-js/issues/178#issuecomment-1287000325, I dropped the v6.4.0 deadline from the schedule. No other changes.

PR #195, opened 2022-10-20
Add example of conversation filtering

neekolas:
## Summary

Updates README to include examples using conversation filtering. Goes through the use-case of creating multiple conversations, filtering the list, and accessing conversation metadata.


jhaaaa (approved):
This looks great, @neekolas - thank you for taking the time to add this info to the README!

PR #196, opened 2022-10-21
Add conversation reference to messages

neekolas:
## Summary

In many places, we may want a reference to the conversation the message originated from attached to the message for convenience. But the most important is in `streamAllMessages` where there may be multiple sources for a given message, and no obvious reference to point back to (the conversation may not even be loaded by the caller of `streamAllMessages`).

I've only added the reference to V2 messages for now. Implementing in V1 would be a much more involved process, and would require a big refactor.

### DIFF
--- a/src/conversations/Conversation.ts
+++ b/src/conversations/Conversation.ts
@@ -146,7 +146,7 @@ export class ConversationV2 {
     options?: SendOptions
   ): Promise<MessageV2> {
     const msg = await this.encodeMessage(message, options)
-    this.client.publishEnvelopes([
+    await this.client.publishEnvelopes([
       {
         contentTopic: this.topic,
         message: msg.toBytes(),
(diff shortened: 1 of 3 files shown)


mkobetic:
This makes total sense. 👍

src/conversations/Conversation.ts
```diff
@@ -146,7 +146,7 @@ export class ConversationV2 {
     options?: SendOptions
   ): Promise<MessageV2> {
     const msg = await this.encodeMessage(message, options)
-    this.client.publishEnvelopes([
+    await this.client.publishEnvelopes([
```
neekolas, line 149: @mkobetic I also noticed that we weren't awaiting the sending of V2 messages.

This might explain why we needed to sleep for longer than expected to deal with race conditions in tests.

mkobetic: Ah good catch. I wonder how many of these landmines have I buried in here. :/

mkobetic (approved):
❤️

PR #197, opened 2022-10-24
Public API Cleanup

neekolas:
## Summary
I opened this PR to clean up some of our public APIs in the SDK and fully embrace the changes that come with V7. 

## The Problem
1. With the initial move to the V7 SDK, there are a number of public methods on the `Client` class that are not really meant for use by the public and are specific to `MessageV1`. For example, `listIntroductionMessages`, `encodeMessage`, `sendMessage`, `decodeEnvelope`, `streamIntroductionMessages`, `streamConversationMessages`, `listConversationMessagesPaginated`, `listMessages`. These methods are not meant to be used directly now that we have two types of conversations.
2. Our `Message` types have become overloaded. They are DTOs functioning as wrappers of our wire format, contain intermediate data used for validation/encryption/decryption of messages, and are public APIs meant to be used in client applications. These differing uses makes them difficult to reason about as a developer.
3. Because the `Message` type deals with content before and after decryption, there are many fields that are treated as mutable or optional (`content`, `senderAddress`) which should be treated as required from the perspective of client applications. We have also had to bolt on fields that come from outside the message itself (`contentTopic`, `conversation`). That makes the type checker not very useful when dealing with messages, and it would be easy for future developers to accidentally omit those fields.

## Solutions
1. Move all functions specific to `MessageV1` or `ConversationV1` out of the `Client` and into the `MessageV1` and `ConversationV1` classes or standalone utility functions.
2. Introduce a new class, `DecodedMessage` with much stricter types that represent the expected shape of a message after validation/decoding/decryption.
3. Make `MessageV1` and `MessageV2` immutable. All operations that previously would mutate the existing object (`decodeMessage`, `decodeContent`), are now are handled by pure functions. This should make the code easier to reason about.

## Todo
- ~Port over all the tests that I broke in the process~
- ~Finalize the `DecodedMessage` type (do we actually want some reference to the underlying `MessageV1`/`MessageV2` somewhere?)~

## Notes
- I no longer treat `NoMatchingPreKeyError` as a special case that will be swallowed and return a message with no content. Instead, those errors will bubble up and the message will be ignored.
- Messages with a content type not supported by the SDK and no fallback will still have an undefined `content` field. Technically this is valid, since `content` has an `any` type, but it doesn't feel great.


test/Client.test.ts
```diff
@@ -74,300 +61,13 @@ describe('Client', () => {
         assert.deepEqual(alice.legacyKeys.getPublicKeyBundle(), alicePublic)
       })
 
-      it('send, stream and list messages', async () => {
```
neekolas, line 77: I've moved most of these to the `Conversation.test.ts`

src/conversations/Conversations.ts
```diff
@@ -117,19 +115,33 @@ export default class Conversations {
...
         }
+        // Decrypt the message to ensure it hasn't been spoofed
+        await msg.decrypt(this.client.legacyKeys)
+        const peerAddress =
+          msg.senderAddress === this.client.address
+            ? msg.recipientAddress
+            : msg.senderAddress
+
+        // Temporarily create a convo to decrypt the message
+        const convo = new ConversationV1(
```
neekolas, line 138: I don't love having to create this throwaway conversation instance, but it feels like a necessary consequence of making all decoding tied to the conversation.

src/Stream.ts
```diff
@@ -47,26 +47,30 @@ export default class Stream<T> {
...
-      // decoder can return undefined to signal a message to ignore/skip.
-      if (!msg) {
-        return
-      }
-      // Check to see if we should update the stream's content topic subscription
-      if (contentTopicUpdater) {
-        const topics = contentTopicUpdater(msg)
-        if (topics) {
-          this.resubscribeToTopics(topics)
+      try {
```
neekolas, line 50: Needed to wrap this in a try/catch since otherwise the errors would bubble up if an invalid message was found in the stream.

test/conversations/Conversation.test.ts
```diff
@@ -18,182 +29,365 @@ describe('conversation', () => {
...
-    })) {
-      for (const msg of page) {
-        if (lastMessage && lastMessage.sent) {
-          expect(msg.sent?.valueOf()).toBeLessThanOrEqual(
-            lastMessage.sent?.valueOf()
-          )
+      expect(numMessages).toBe(1)
+    })
+
+    it('works for messaging yourself', async () => {
```
neekolas, line 128: We had this test before in the SDK. I wonder if we should log a warning when someone does this

src/conversations/Conversation.ts
```diff
@@ -47,25 +64,114 @@ export class ConversationV1 {
...
+    ) {
+      throw new Error('Headers do not match intended recipient')
+    }
+    const decrypted = await decoded.decrypt(this.client.legacyKeys)
+    const { content, contentType, error } = decodeContent(
+      decrypted,
+      this.client
+    )
+
+    return new DecodedMessage({
```
mkobetic, line 109: Why isn't this calling DecodeMessage.fromV1Message()?

src/conversations/Conversations.ts
mkobetic: Yeah, I think this is just something we need to accept. Nice to see that separating Invitation from Message avoids this.

src/index.ts
```diff
@@ -1,56 +1,30 @@
...
-import {
-  Composite,
-  CompositeCodec,
-  ContentTypeComposite,
-} from './codecs/Composite'
-import { SortDirection } from './ApiClient'
-
+export { TextCodec, ContentTypeText } from './codecs/Text'
 export {
-  Client,
```
mkobetic, line 32: ❤️ Really like seeing this trimmed down. What was the criteria you used to clean this up?

test/conversations/Conversation.test.ts
mkobetic: Do you want to disallow this?

mkobetic (approved):
I really like these changes, for all the reasons you've listed. I think this is a major face lift for the API. ❤️ 👍 💯

src/Message.ts
```diff
@@ -254,3 +242,126 @@ export class MessageV2 extends MessageBase implements proto.MessageV2 {
...
 export type Message = MessageV1 | MessageV2
+
+export interface DecodedMessage {
+  id: string
+  messageVersion: 'v1' | 'v2'
+  senderAddress: string
+  conversation: Conversation
+  contentType: ContentTypeId
+  sent: Date
+  contentTopic: string
```
mkobetic, line 253: contentTopic seems to be specific to V1, do we want to make it part of the interface?

src/Message.ts
```diff
@@ -254,3 +242,126 @@ export class MessageV2 extends MessageBase implements proto.MessageV2 {
...
+  recipientAddress?: string
```
mkobetic, line 254: I have doubts about recipientAddress too, but it may not be easy to get rid of it.

src/Message.ts
```diff
@@ -254,3 +242,126 @@ export class MessageV2 extends MessageBase implements proto.MessageV2 {
...
+  error?: Error
+  content: any // eslint-disable-line @typescript-eslint/no-explicit-any
+}
+
+export class DecodedMessage {
```
mkobetic, line 259: How come you can export two things with the same name?

src/conversations/Conversation.ts
neekolas: That's a good question. Mostly because I wrote `fromV1Message` later. Fixed

src/Message.ts
neekolas: So, I think in the context of V2 messages it is also relevant (especially in a world where we have stitching and the reference to the Conversation is not sufficient to know whether the message came from the latest topic).

It's also strictly necessary to figure out if a message came from the DM or Intro topic in `streamAllMessages`.

mkobetic: Good point about stitching, but we could introduce it later when we get to it.

The streamAll aspect is what I meant that it's a V1 only concern. So I still feel it would be better to leave this out for now.

src/Message.ts
neekolas: Interfaces allow it (they become overloads), although I realized I don't really need the interface at all anymore.

src/Message.ts
neekolas: It definitely has more of a chance of breaking downstream apps, but we do all our checks based on recipient address before we convert to a `DecodedMessage` (I think, possibly not true in tests) so it's safe to do on our end.

test/conversations/Conversation.test.ts
neekolas: I don't want to disallow it. But I suspect it's being done by mistake if any actual user is doing it. Or they are using the SDK as some sort of private storage library, which it really wasn't designed for.

src/Message.ts
mkobetic: OK, fair enough. It's not really that harmful to have it. It was more about trying to keep the interface as tidy as possible.

test/conversations/Conversation.test.ts
mkobetic: I see, I'm ok either way. I just don't find warnings a good way to guide users, it's more of a last resort to surface something if nothing better is available.

PR #198, opened 2022-10-25
Remove LocalStorageStore

neekolas:
## Summary
In one more "take-out-the-trash" change I'd like to handle in V7, this PR gets rid of the LocalStorageStore entirely. It seems to be pretty much unused, is problematic whenever it is used (how do keys get synced between instances?), and causes all kinds of compatibility problems in different environments.

This change just removes it altogether. Good riddance.

## Bonus

I realized that we do not need to be using the `--runInBand` flag for our tests, and can safely run them in parallel. That speeds up test performance _a lot_.


mkobetic (approved):
👍

PR #200, opened 2022-10-26
fix: include mismatched pre-key in error

mkobetic:
This should make it easier to debug causes of the error if it happens.

This PR just dumps the key bytes in hex, but maybe we want a more user-friendly fingerprint?


neekolas (approved):
Seems safe enough. These are all public

PR #201, opened 2022-10-27
Catch errors with no matching preKey

neekolas:
## Summary

Helps mitigate INC-2, the No Matching PreKey error incident

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -369,7 +369,7 @@ export default class Client {
         const res = await mapper(env)
         results.push(res)
       } catch (e) {
-        console.log('Error in listEnvelopes mapper', e)
+        console.warn('Error in listEnvelopes mapper', e)
       }
     }
     return results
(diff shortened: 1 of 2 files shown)


src/Client.ts
```diff
@@ -369,7 +369,7 @@ export default class Client {
         const res = await mapper(env)
         results.push(res)
       } catch (e) {
-        console.log(e)
+        console.log('Error in listEnvelopes mapper', e)
```
mkobetic, line 372: Should we use `warn` here too?

elisealix22 (approved):

mkobetic:
Do we already suppress errors decoding conversation messages? The same error will pop up there for the same message.

mkobetic:
Also we need the same handling on the streaming side too.

mkobetic:
> Do we already suppress errors decoding conversation messages? The same error will pop up there for the same message.

I'm thinking V1 here which doesn't apply to lenster, but I think the point still stands.

neekolas:
We already catch all errors in the mapper functions for streams and `listMessages`. This was the one exception, where the decoding happened separately from the mapper.

mkobetic (approved):

PR #203, opened 2022-10-31
Verify encryption signature

snormore:
There's a bug in the interaction with LedgerLive for iOS where the previous signature is returned as the response of a consecutive signature request, and so this PR verifies that signature, and retries if it's not what's expected, which seems to solve the issue when interacting with LedgerLive.

### DIFF
--- a/src/store/EncryptedStore.ts
+++ b/src/store/EncryptedStore.ts
@@ -1,5 +1,5 @@
 import { Store } from './Store'
-import { Signer } from 'ethers'
+import { Signer, utils } from 'ethers'
 import {
   PrivateKeyBundleV1,
   decodePrivateKeyBundle,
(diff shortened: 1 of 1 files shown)


mkobetic:
> The PR also updates the decryption logic to recreate the improperly encrypted key bundle if it fails to decrypt, so that existing accounts that may have encountered this bug can continue through their currently broken initialization workflow.

This bit worries, are we sure that by recreating the bundle we won't lose previously used keys? That may be similarly catastrophic for a user.

src/store/EncryptedStore.ts
```diff
@@ -73,9 +73,31 @@ export default class EncryptedKeyStore implements KeyStore {
...
+    let address = utils.recoverAddress(utils.hashMessage(input), sig)
+    if (address !== walletAddr) {
+      sig = await wallet.signMessage(input)
+      console.log('invalid signature, retrying', address, walletAddr)
+    }
+
+    address = utils.recoverAddress(utils.hashMessage(input), sig)
+    if (address !== walletAddr) {
+      throw new Error('invalid signature')
+    }
```
jazzz, line 94: In this case I don't think lower level recoverAddress is needed. Verify message should work

```
const recoveredAddr = utils.verifyMessage(input, sig)
if ((await wallet.getAddress()) === recoveredAddr) {
  throw new Error('invalid storage signature')
}
```

jazzz (approved):

src/store/EncryptedStore.ts
```diff
@@ -73,9 +73,27 @@ export default class EncryptedKeyStore implements KeyStore {
...
+
+    // Check that the signature is correct, was created using the expected
+    // input, and retry if not. This mitigates a bug in interacting with
+    // LedgerLive for iOS, where the previous signature response is
+    // returned in some cases.
+    let address = utils.verifyMessage(input, sig)
+    if (address !== walletAddr) {
+      sig = await wallet.signMessage(input)
+      console.log('invalid signature, retrying')
+    }
```
mkobetic, line 89: Nice, not for this PR, but we should probably use `verifyMessage()` in `WalletSigner.signerKey()` as well to simplify the code there.

mkobetic (approved):

snormore:
Does this need to be merged into the `beta` branch too?

neekolas:
> Does this need to be merged into the `beta` branch too?

Yes please

PR #204, opened 2022-10-31
Publish v2 contact by default

mkobetic:
Switching the default published contact bundle to V2. This will start spilling network traffic over to the V2 style as more and more clients adopt v7.0.0.

It also fixes the constant contact republishing on every client initialization. Instead it reads the current contact from the network and republishes it if it's different from what the private key bundle implies. This should reduce the churn on the contact topic, while still allow clients to fix up the current contact if it's stale or if there's garbage in the contact topic somehow.

In order to keep most of the test suite operating with V1 mode as before, I'm introducing a legacy client option `publishLegacyContact?: boolean` that allow overriding the default V2 behavior. This allows me to configure the client in the test suite to preserve the old behavior. It also makes this explicit in the test suite definition, so it will remind us that some tests are still running with the V1 behavior.

I'm also adding bunch of sleeps to the tests as the test were constantly failing for me running locally. There were quite a few races between sends and lists and also between streams and lists. The latter is interesting, because one would assume that if a message shows up on a stream it should show up in a list as well, but the thing to remember that streams operate from the relay whereas list operates from the store, so it is possible for the relay to deliver a message before it is actually committed to the store.

More comments inline...

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -224,7 +224,13 @@ export default class Client {
     ) {
       return
     }
-    this.publishUserContact(legacy)
+    // TEMPORARY: publish V1 contact to make sure there is one in the topic
+    // in order to preserve compatibility with pre-v7 clients.
+    // Remove when pre-v7 clients are deprecated
+    this.publishUserContact(true)
+    if (!legacy) {
+      this.publishUserContact(legacy)
+    }
   }
 
   // PRIVATE: publish the key bundle into the contact topic
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -97,11 +97,18 @@ type KeyStoreOptions = {
   privateKeyOverride?: Uint8Array
 }
 
+type LegacyOptions = {
+  publishLegacyContact?: boolean
+}
+
 /**
  * Aggregate type for client options. Optional properties are used when the default value is calculated on invocation, and are computed
  * as needed by each function. All other defaults are specified in defaultOptions.
  */
-export type ClientOptions = NetworkOptions & KeyStoreOptions & ContentOptions
+export type ClientOptions = NetworkOptions &
+  KeyStoreOptions &
+  ContentOptions &
+  LegacyOptions
 
 /**
  * Provide a default client configuration. These settings can be used on their own, or as a starting point for custom configurations
@@ -195,12 +202,12 @@ export default class Client {
     return client.legacyKeys.encode()
   }
 
-  async init(options: ClientOptions): Promise<void> {
+  private async init(options: ClientOptions): Promise<void> {
     options.codecs.forEach((codec) => {
       this.registerCodec(codec)
     })
     this._maxContentSize = options.maxContentSize
-    await this.publishUserContact()
+    await this.ensureUserContactPublished(options.publishLegacyContact)
   }
 
   // gracefully shut down the client
@@ -208,10 +215,21 @@ export default class Client {
     return undefined
   }
 
-  // publish the key bundle into the contact topic
-  // WARNING: temporarily public to allow testing negotiated topics
-  // TODO: make private again asap
-  async publishUserContact(legacy = true): Promise<void> {
+  private async ensureUserContactPublished(legacy = false): Promise<void> {
+    const bundle = await getUserContactFromNetwork(this.apiClient, this.address)
+    if (
+      bundle &&
+      bundle instanceof SignedPublicKeyBundle &&
+      this.keys.getPublicKeyBundle().equals(bundle)
+    ) {
+      return
+    }
+    this.publishUserContact(legacy)
+  }
+
+  // PRIVATE: publish the key bundle into the contact topic
+  // left public for testing purposes
+  async publishUserContact(legacy = false): Promise<void> {
     const keys = legacy ? this.legacyKeys : this.keys
     await this.publishEnvelopes([
       {
@@ -247,6 +265,9 @@ export default class Client {
     return newBundle
   }
 
+  /**
+   * Used to force getUserContact fetch contact from the network.
+   */
   forgetContact(peerAddress: string) {
     this.knownPublicKeyBundles.delete(peerAddress)
   }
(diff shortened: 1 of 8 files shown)


src/Client.ts
```diff
@@ -195,23 +202,34 @@ export default class Client {
...
+      this.keys.getPublicKeyBundle().equals(bundle)
+    ) {
+      return
+    }
+    this.publishUserContact(legacy)
+  }
+
+  // PRIVATE: publish the key bundle into the contact topic
+  // left public for testing purposes
+  async publishUserContact(legacy = false): Promise<void> {
```
mkobetic, line 232: The mixed v1/v2 test cases still need to be able to republish the contact, so I have to leave this open.

test/conversations/Conversation.test.ts
```diff
@@ -22,14 +22,14 @@ describe('conversation', () => {
   let alice: Client
   let bob: Client
 
-  beforeEach(async () => {
```
mkobetic, line 25: moving this down to the respective v1/v2 suites so that the client can be initialized accordingly.

test/conversations/Conversation.test.ts
```diff
@@ -331,18 +334,17 @@ describe('conversation', () => {
...
   describe('v2', () => {
+    beforeEach(async () => {
+      alice = await newLocalHostClient()
+      bob = await newLocalHostClient()
+      await waitForUserContact(alice, alice)
+      await waitForUserContact(bob, bob)
+    })
+
     it('v2 conversation', async () => {
-      // publish new contact bundles
```
mkobetic, line 335: This isn't necessary anymore since the correct version of the contact is published by default.

test/conversations/Conversation.test.ts
```diff
@@ -385,134 +387,140 @@ describe('conversation', () => {
       for (let i = 0; i < 5; i++) {
         await bobConvo.send('hi')
       }
+      await sleep(100)
       const messages = await bobConvo.messages({ limit: 2 })
       expect(messages).toHaveLength(2)
     })
-  })
 
-  it('conversation filtering', async () => {
```
mkobetic, line 393: From here down the code isn't changing, it's just moving into the v2 suite scope. It was at the top level before which had it effectively running with v1 behavior, but these are largely v2 tests, so I think they should run in v2 scope.

test/conversations/Conversations.test.ts
```diff
@@ -40,13 +40,11 @@ describe('conversations', () => {
 
     const aliceConversationsAfterMessage = await alice.conversations.list()
     expect(aliceConversationsAfterMessage).toHaveLength(1)
-    expect(await aliceConversationsAfterMessage[0].peerAddress).toBe(
```
mkobetic, line 43: removing few unnecessary awaits (earlier we've changed peerAddress to be sync)

test/helpers.ts
```diff
@@ -119,5 +119,8 @@ export const newLocalHostClient = (
   })
 
 // client running against the dev cluster in AWS
-export const newDevClient = (): Promise<Client> =>
-  Client.create(newWallet(), { env: 'dev' })
+export const newDevClient = (opts?: Partial<ClientOptions>): Promise<Client> =>
```
mkobetic, line 122: making `newDevClient()` respect client options.

neekolas:
I thought we wanted to wait to roll this change out until more developers (any, other than us) were already on V7?

I'm not sure what the rush is.

mkobetic:
> I thought we wanted to wait to roll this change out until more developers (any, other than us) were already on V7?
> 
> I'm not sure what the rush is.

I thought this would be part of the v7.0.0 release. Without it it's not quite a V2 switch. I'm open to discuss this though. What timeline would you propose? I just updated https://github.com/xmtp/xmtp-js/issues/178 with what I assumed.

neekolas:
@mkobetic I would suggest we release V7 as-is, given that we have already tested it thoroughly and have confidence it interoperates with existing XMTP apps well.

Then we set a date for the V2 contact rollout maybe 3-4 weeks out. At least gives developers a chance to upgrade to a version that is interoperable with V2 bundles before we start publishing them.

jazzz:
> Then we set a date for the V2 contact rollout maybe 3-4 weeks out. 

I'm not opposed. I can see the appeal of reducing entropy for v7

> At least gives developers a chance to upgrade to a version that is interoperable with V2 bundles before we start publishing them.

Thought I would say as @mkobetic martin noted here(https://github.com/xmtp/xmtp-js/issues/178#issuecomment-1287000325) there is no blocker on the V2 Bundles roll out. 

Waiting 3-4 weeks means developers will need to update again later to get the V2 bundles by default.

neekolas:
> Given that this should also be a fully backward compatible change, what are we worried about exactly

It's not a backward compatible change if the first time you use XMTP is from a 7.x client, and then someone wants to message you using a 6.x client. They won't have a contact bundle the 6.x client understands.

mkobetic:
> > Given that this should also be a fully backward compatible change, what are we worried about exactly
> 
> It's not a backward compatible change if the first time you use XMTP is from a 7.x client publishing a new style bundle, and then someone wants to message you using a 6.x client. They won't have a contact bundle the 6.x client understands.

Good point. Would publishing both v1 and v2 contact address this concern?

neekolas:
> Good point. Would publishing both v1 and v2 contact address this concern?

I don't love it, but it would solve this problem. The only client versions that would fail are the same ones we are deprecating on the 8th.

mkobetic:
> > Good point. Would publishing both v1 and v2 contact address this concern?
> 
> I don't love it, but it would solve this problem. The only client versions that would fail are the same ones we are deprecating on the 8th.

This would presumably be temporary, until we deprecate pre-v7.0.0 clients. I'll make that change.

neekolas:
That would make me lose interest in waiting a long time to merge. I do still think we should wait until next week to respect the code freeze we have going for Lenster

mkobetic:
Next week sounds good to me. :+1:

mkobetic:
@saulmc this PR will disable your session metrics, we will stop republishing contacts every time the client initializes, we will only republish if the contact changes.

saulmc:
@mkobetic that's OK :) we knew that visibility would be short-lived. thanks for the heads up!

neekolas (approved):

Issue #202, opened 2022-10-28
Add app identifier arg to Client.create


saulmc:
From @snormore

We should consider adding a required arg (or even an optional one) to Client.create for an app identifier like `lenster/1.0.0` so we can get telemetry on users of the client/sdk

PR #207, opened 2022-11-01
Add client app version on api requests

snormore:
Adds optional client app version identifier that's included with API requests.

https://github.com/xmtp/xmtp-js/issues/202


saulmc:
this is awesome!!!

@jhaaaa let's figure out how best to document to maximize adoption. I think we can be pretty assertive in the README.

mkobetic (approved):
👍

PR #210, opened 2022-11-02
Friendlier log level when resubscribing

snormore:
Reduce log level from `warn` to `info` to avoid the intimidating highlight + error expansion

Screen Shot 2022-11-02 at 4 38 37 PM


neekolas (approved):
Seems reasonable

PR #213, opened 2022-11-07
docs: emphasize `env` default is `dev`

jhaaaa:
Update README to emphasize the ability to set a client environment parameter and that the default value is `dev`


mkobetic:
I wonder if we should emphasize somehow that the networks are not interchangeable and that there is no way to move accounts, messages etc between them.

README.md
```diff
@@ -43,9 +43,11 @@ Issues and PRs are welcome in accordance with our [contribution guidelines](http
 
 XMTP provides both `production` and `dev` network environments to support the development phases of your project.
 
-The `production` network is configured to store messages indefinitely. XMTP may occasionally delete messages and keys from the `dev` network, and will provide advance notice in the [XMTP Discord community](https://discord.gg/xmtp).
+The `production` and `dev` networks are not interchangeable. For example, an account (XMTP identity) and messages created on the `dev` network can't be accessed from or moved to the `production` network, and vice versa.
```
jhaaaa, line 46: @mkobetic I added this text above to call out that the networks are not interchangeable per your excellent suggestion! Any thoughts? =) cc @saulmc

mkobetic: I think the change capture what I had in mind, but I would also preserve the description of the differences between the two network that was there.

jhaaaa: Thank you, @mkobetic! Ah yes, I kept that info, just moved it below. =) What do you think of the new sequence of info? I was thinking this sequence expresses the correct priority of info - but perhaps not? =)

mkobetic: My bad, I missed that @jhaaaa, it's all there indeed.

mkobetic:
In light of @elisealix22's #215 I think we may want to elaborate on the `dev`/`prod` question even more. What is the purpose, why should users want to use one over the other and when?

mkobetic (approved):
I think this is worth shipping even if we want to elaborate more on this later.

jhaaaa:
@mkobetic - Thank you so much for your approval! 

I made a few updates based on your meaningful feedback to my questions in https://github.com/xmtp/xmtp-js/pull/215. 

Might you be willing to take another look just to make sure I haven't introduced anything that would make you rescind your approval? ;-) Thank you again!

mkobetic:
@jhaaaa those are great updates. Thank you!

PR #215, opened 2022-11-08
RFC: Require an XMTP env when creating `Client`

elisealix22:
Opening this up as an RFC because this would be a **BREAKING CHANGE** for everyone.

## Context

In the past few weeks, we've seen confusion around which environment we are pointing to causing bugs (and missing messages) internally and externally. Even us employees at XMTP Labs sometimes forget to set the `production` env when calling `Client.canMessage()` or `Clinet.getKeys()` since it is optional right now. We saw this with Lenster and have heard this is a friction point from developers in Discord and at Hackathons.

## Goal

I think requiring `env` when creating a Client or using any of the static methods on Client will push developers to consider the env and likely make a constant for it to be used everywhere. This is what we did for Lenster [here](https://github.com/lensterxyz/lenster/blob/fe30651fe0b56706f7b07b896027b6ac5c679b8c/src/constants.ts#L25). I think this is an instance where a little bit of friction will help with education and best practices and hopefully make it like bowling with gutter guards to do the right thing.

Ideally, we will see less valuable messages on the dev network that want to get ported over to production as well as fewer `No matching pre-key` errors with people accidentally using dev keys on the production env via `Client.getKeys()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ import { Wallet } from 'ethers'
 // You'll want to replace this with a wallet from your application
 const wallet = Wallet.createRandom()
 // Create the client with your wallet. This will connect to the XMTP development network by default
-const xmtp = await Client.create(wallet)
+const xmtp = await Client.create(wallet, 'dev')
 // Start a conversation with XMTP
 const conversation = await xmtp.conversations.newConversation(
   '0x3F11b27F323b62B159D2642964fa27C46C841897'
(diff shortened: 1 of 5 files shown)


mkobetic:
You put forward some very good arguments @elisealix22, but I think that before we decide whether this should be optional or required parameter, we need to clarify the purpose of the two networks and communicate that better (seems we need this for ourselves too). Why should a developer care about dev network? Why is it useful to them? Why not just push everything to production? If we fail there we may as well end up with most people just mindlessly setting it to production. In that case even better solution is to keep it optional and default to production. If we don't explain the differences and purpose well enough, people won't make the right choices for themselves.

jazzz:
> we need to clarify the purpose of the two networks and communicate that better

Absolutely agree. From working with teams at ethSF it was clear the major tension and source of bugs was that developers were not aware that there were two different networks (let alone being able to distinguish which network their app is connecting to)

 The next follow up question was always.... "if production is free, why would I use dev?"

I think adding some education and visibility around the networks is needed

elisealix22:
I've learned so much today on this! I think ultimately we do not feel this is the correct direction as it does add friction to getting started. Some alternatives we can explore that came from this RFC (in Slack and in today's eng weekly meeting):

1. Do we even need to expose two networks?
    - Seems like yes because we want to be able to distinguish between hackathon/testing traffic and real traffic. Also want to make sure people have a place to test where they aren't necessarily bloating their production addresses.
2. Can we use a global env variable?
   - This seems promising! If we set a global env variable we'd still be able to ensure the static methods are configured properly for the given environment. It also would only have to be set once instead of the original proposed solution where we pass it in every call. This would also be a breaking change though.
3. Can we spam the logs when you're on the dev network?
    - This seems like a very quick win and it's not a breaking change! The idea is for every time client is instantiated or one of the static Client methods is called, we would output a big log to indicate you are on the dev network. I'm not sure if everyone will catch it, but it should hopefully help with the education piece on there being multiple networks.

jhaaaa:
Thank you so much for writing this up and facilitating this conversation, @elisealix22! 

I will do all I can to ensure that the documentation reflects best practices for setting the environment based on the outcomes of this RFC.

FWIW, I assumed that the parameter described in [Configuring the Client](https://github.com/xmtp/xmtp-js#configuring-the-client) in the README was a global value because it is the only place we explicitly talk about setting the value. 

I also noticed that this section in the README lists a `local` value - is this a third and supported `env` value?

saulmc:
@elisealix22 I have no idea! Absent data suggesting meaningful usage, I would be in favor of deprecation given the complexities rooted in this option.

neekolas:
The local env is primarily used for running our test suite

mkobetic:
It's also useful for node development. I often run the node locally and run the test suite against it to generate some client traffic on the node. I'd prefer to not lose that.

neekolas:
Great point @mkobetic. I do the same thing

jhaaaa:
Per @mkobetic's suggestion, I thought I'd take a pass at defining the impact and benefits of setting `env` to each of the supported values. Just a starting point - but might this help us define a path forward? =) 

- `dev` (default): Use to have a client communicate with the `dev` network. As a best practice, set `env` to `dev` while testing your app. Following this best practice can help isolate test messages to `dev` XMTP inboxes.

- `production`: Use to have a client communicate with the `production` network. As a best practice, set `env` to `production` when your app is ready to go live. Following this best practice can reduce the number of test messages displayed in `production` XMTP inboxes.

- `local`: Use to have a client communicate with an XMTP node you are running locally. For example, an XMTP node developer can set `env` to `local` to generate client traffic to test a node running locally.

**Open questions:**

- Do we want to mention that `dev` and `production` networks are identical?
- I'm unsure about the specifics around when a developer should switch between `dev` and `production` ... 
- DUH question - but I have to ask - if you use `local` - nothing is saved to the XMTP network, correct? Is this obvious - or do we need to say it?

mkobetic:
An argument agains making the env setting global occurred to me: if we want to write scripts that migrate data between networks, we would probably want to be able to instantiate two clients each connected to a different network. I'm not sure that we will ever want to do this, but I think it's worth considering.

If we do think this usecase is worth supporting, that still doesn't kill the global setting idea. We can still take the default environment from a global setting and allow overriding it explicitly when instantiating the client. Recommend using the global setting wherever possible, and only use the instantiation parameter as a last resort. Anyway, just thought I'd throw this out there.

mkobetic:
I think your definitions look pretty good.

> **Open questions:**
> 
> * Do we want to mention that `dev` and `production` networks are identical?

I wouldn't say they are identical, at the very least the retention guarantees and possibly other quality of service guarantees are different. I think the critical aspect to highlight is that the two networks are **isolated**, meaning that for a given wallet, its identity on dev network is completely distinct from its identity on the production network, as are the conversations associated with these identities. That's what ensures that test messages in one network do not pollute the inbox on the other network.

> * I'm unsure about the specifics around when a developer should switch between `dev` and `production` ...

I think you got the gist there. Use `dev` for testing and development, use production for real user service only. Keep the two activities separate and isolated.

> * DUH question - but I have to ask - if you use `local` - nothing is saved to the XMTP network, correct? Is this obvious - or do we need to say it?

Good question, it should be completely detached from any "real" network. In theory a local node could potentially be connected to a wider network, but that's not really possible today. I wouldn't worry about it. I think what you wrote is good enough.

elisealix22:
Alright, I truly appreciate everyone's input/discussion around this! We opened up the [PR](#221) to loudly log when connected to the dev network which feels to me like a positive direction. We'll keep an eye out if we want to move to using a global variable if this is not enough along with the updated documentation in #213.

PR #219, opened 2022-11-08
docs: add documentation for streamAllMessages() to readme

### LINKED ISSUE #214: Add `streamAllMessages()` snippet to readme
Based on the conversation in `https://discord.com/channels/831836269558235136/993229186951028796/1039310518348951653`, I think it would be a good idea to add a snippet to the readme that illustrates use of `streamAllMessages()`.  I'm thinking the easiest way would be to update the instructions at `https://xmtp.org/docs/client-sdk/javascript/tutorials/quickstart#listen-for-new-messages-in-a-conversation`, adding at the bottom:

>To listen for any incoming messages from *all* conversations, use `streamAllMessages()`:
>
>```
>for await (const message of await xmtp.conversations.streamAllMessages()) {
>
>  if (message.senderAddress === xmtp.address) {
>    // This message was sent from me
>    continue
>  }
>  console.log(`New message from ${message.senderAddress}: ${message.content}`)
>}
>```

I'd be happy to take this on if you want to assign it to me.

joshuasorkin:
This is intended to resolve #214 by providing a code snippet for use of `streamAllMessages()` in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,13 +253,12 @@ for await (const message of await conversation.streamMessages()) {
 
 #### Listen for new messages in all conversations
 
-To listen for any new incoming/outgoing messages from *all* conversations, use `conversations.streamAllMessages()`.
+To listen for any new incoming/outgoing messages from _all_ conversations, use `conversations.streamAllMessages()`.
 
 **Note:** there is a chance this stream can miss messages if multiple conversations begin at nearly the same time while we are updating the stream to account for a new conversation.
 
 ```ts
 for await (const message of await xmtp.conversations.streamAllMessages()) {
-
   if (message.senderAddress === xmtp.address) {
     // This message was sent from me
     continue
--- a/README.md
+++ b/README.md
@@ -251,6 +251,21 @@ for await (const message of await conversation.streamMessages()) {
 }
 ```
 
+#### Listen for new messages in all conversations
+
+To listen for any new incoming/outgoing messages from *all* conversations, use `streamAllMessages()`:
+
+```ts
+for await (const message of await xmtp.conversations.streamAllMessages()) {
+
+  if (message.senderAddress === xmtp.address) {
+    // This message was sent from me
+    continue
+  }
+  console.log(`New message from ${message.senderAddress}: ${message.content}`)
+}
+```
+
 #### Checking if an address is on the network
 
 If you would like to check and see if a blockchain address is registered on the network before instantiating a client instance, you can use `Client.canMessage`.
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -251,6 +251,21 @@ for await (const message of await conversation.streamMessages()) {
 }
 ```
 
+#### Listen for new messages in all conversations
+
+To listen for any new incoming/outgoing messages from *all* conversations, use `streamAllMessages()`:
```
elisealix22, line 256:
```suggestion
To listen for any new messages from *all* conversations, use `converstaions.streamAllMessages()`.

Note: there is a chance this stream can miss messages if multiple conversations begin at nearly the same time while we are updating the stream to account for a new conversation.
```

elisealix22 (commented):
Love it! Thank you for opening this. I left a suggestion to note the possible edge case with multiple conversations beginning at the same time. Also adding our techincal writer @jhaaaa for a second pair of eyes.

README.md
jhaaaa: Thank you so much for taking such initiative in opening this PR, @joshuasorkin!

And thank you to @elisealix22 for your refinements and the very important note!

Just a couple of nit tweaks and a question! <3

- Change`converstaions` --> `conversations`

- Change `Note: there` -> `**Note:** There`

- Question: In the statement `we are updating` - I wonder if it might be possible to say what "we" is? What is updating the stream? The client?

elisealix22: @jhaaaa wow, thank you so much! I updated the suggestion to account for those notes.

@joshuasorkin it's actually due to the `contentTopicUpdater` which is called when a new message is received. So we start the stream listening for all existing conversations' content topics (where one content topic represents one existing conversation) as well as the invite topic to ensure we also receive new messages for new conversations as they are created.

The tricky part is once we receive an invite topic for a new conversation, we want to update the existing stream's content topics to include that new conversation's content topic so the stream can receive new messages for that conversation as well. To accomplish this, we use the `contentTopicUpdater` to update the stream's list of content topics as we receive messages indicating a new conversation has started. The potential race condition is that if we receive multiple new conversations in the time it takes to update the stream's content topics (I have yet to see this in practice for what it is worth), we could potentially drop a content topic while the stream is resubscribing to a different conversations' content topic.

https://github.com/xmtp/xmtp-js/blob/2868899477e1dc6afddc38e71214354c8e154b0b/src/conversations/Conversations.ts#L186

jhaaaa (approved):
Thank you for such a meaningful update, @elisealix22 and @joshuasorkin - LGTM!

README.md
joshuasorkin: @elisealix22 thank you for clarifying!  Is there a standard tag that I could drop into the comment to mark this as a theoretical bug?  I'd like it to show up on a search for issues that need to be addressed eventually but are not currently affecting performance.

elisealix22 (approved):
Great addition imo, @joshuasorkin ! Thank you.

PR #220, opened 2022-11-09
Updated with lower casing the peerAddress

edisonqu:
Lower cased the peerAddress for both of the functions, making sure that they are both standardized... 

Hopefully this works, please let me know if it does not, I want to know a fix-- especially if this is only a small part of the problem fixed


edisonqu:
no longer needed; [ethers.utils.getAddress ](https://docs.ethers.io/v5/api/utils/address/#utils-getAddress)works like a charm when standardizing the checkSum format.

PR #222, opened 2022-11-11
fix: maintain pre-existing legacy conversation

mkobetic:
Fixes error: `recipient bundle is not legacy bundle`

This turned out to be a bit messier than I thought. The issue was clear, `ConversationV1.send()` explicitly checked that the recipient's contact is a V1 bundle, if not it threw this error. So when we started publishing V2 contacts, users that had pre-existing V1 conversations started tripping over this check. The obvious solution is to convert V2 contacts back to V1 contacts for V1 conversations, which is largely what this PR does.

However this conversion is not always possible. This is due to the precision change of creation time between V1 and V2 keys. We moved from millisecond precision to nanosecond precision, which caused trouble from the beginning, but we worked around it by allowing V2 keys converted from V1 keys to stick with ms precision in order to be able to transplant the key signature during the conversion. This enabled V1->V2 conversion, but the inverse is only possible if the V2 key uses the ms precision, i.e. it was converted from V1 to begin with.

This is sufficient for us today, all the keys out there, including new ones being created, are generated as V1 keys and converted to V2. However, we will have to be careful when we start generating native V2 key bundles. Users with native V2 bundles will not be able to participate in V1 conversations. This should be OK as long as we can expect that

1. all users out there to have V2 contacts at that point AND that
2. the user with native V2 keys does not have any pre-existing V1 conversations with anyone.

That means we MUST NOT create V2 keys until we deprecate pre-v7.1 SDK, and ensure that all users have a V2 contact published.

### DIFF
--- a/src/conversations/Conversations.ts
+++ b/src/conversations/Conversations.ts
@@ -285,7 +285,21 @@ export default class Conversations {
       const intros = await this.getIntroductionPeers()
       const introSentTime = intros.get(peerAddress)
       // If intro already exists, return V1 conversation
+      // if both peers have V1 compatible key bundles
       if (introSentTime) {
+        if (!this.client.keys.getPublicKeyBundle().isFromLegacyBundle()) {
+          throw new Error(
+            'cannot resume pre-existing V1 conversation; client keys not compatible'
+          )
+        }
+        if (
+          !(contact instanceof PublicKeyBundle) &&
+          !contact.isFromLegacyBundle()
+        ) {
+          throw new Error(
+            'cannot resume pre-existing V1 conversation; peer keys not compatible'
+          )
+        }
         return new ConversationV1(this.client, peerAddress, introSentTime)
       }
     }
--- a/test/conversations/Conversations.test.ts
+++ b/test/conversations/Conversations.test.ts
@@ -176,6 +176,15 @@ describe('conversations', () => {
 
       const aliceConvo2 = await alice.conversations.newConversation(bob.address)
       expect(aliceConvo2 instanceof ConversationV1).toBeTruthy()
+      await aliceConvo2.send('hi')
+
+      await sleep(100)
+      const bobConvo = await bob.conversations.newConversation(alice.address)
+      expect(bobConvo instanceof ConversationV1).toBeTruthy()
+      const messages = await bobConvo.messages()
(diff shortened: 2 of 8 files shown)


src/conversations/Conversations.ts
```diff
@@ -285,7 +285,21 @@ export default class Conversations {
...
+          )
+        }
+        if (
+          !(contact instanceof PublicKeyBundle) &&
+          !contact.isFromLegacyBundle()
+        ) {
+          throw new Error(
+            'cannot resume pre-existing V1 conversation; peer keys not compatible'
+          )
+        }
```
mkobetic, line 302: These errors should never happen. If they do we messed up V2 private key bundle rollout.

test/conversations/Conversations.test.ts
```diff
@@ -176,6 +176,15 @@ describe('conversations', () => {
 
       const aliceConvo2 = await alice.conversations.newConversation(bob.address)
       expect(aliceConvo2 instanceof ConversationV1).toBeTruthy()
+      await aliceConvo2.send('hi')
```
mkobetic, line 179: This reproduces the reported error.

saulmc:
> Users with native V2 bundles will not be able to participate in V1 conversations

Based on @jhaaaa's [diagram](https://xmtp.org/docs/dev-concepts/architectural-overview#determining-whether-to-use-xmtp-v2-or-v1-topics), I read this as "Users with native v2 bundles will not be able to initiate conversations with users who have not upgraded to v2 bundles." Is that the extent of the limitation?

mkobetic:
> > Users with native V2 bundles will not be able to participate in V1 conversations
> 
> Based on @jhaaaa's [diagram](https://xmtp.org/docs/dev-concepts/architectural-overview#determining-whether-to-use-xmtp-v2-or-v1-topics), I read this as "Users with native v2 bundles will not be able to initiate conversations with users who have not upgraded to v2 bundles." Is that the extent of the limitation?

Basically yes. We have to be careful to make sure that users that get native V2 bundles do not have any pre-existing V1 conversations. There shouldn't be any V2 incapable users at that point, so there shouldn't be a need to create V1 conversations either.

neekolas (approved):

PR #225, opened 2022-11-17
docs: remove empty table row

jhaaaa:


mkobetic (dismissed):

mkobetic:
OK, I can now reproduce the failure locally against the dev network
```
TESTNET=true npm test -t NetworkStore
```

mkobetic:
And now also locally when I pulled the latest xmtp/node-go image from docker hub. plot thickens

mkobetic:
OK, I think I know what's going on, it's that we've finally managed to update the nodes after @neekolas fixed our TF deploy issues yesterday. The nodes now enforce the private topic publishing restrictions and the test is not written to comply with those
```
2022-11-18 13:48:35 2022-11-18T18:48:35.014Z    ERROR   xmtpd.api       api request     {"node": "16Uiu2HAmNCxLZCkXNbpVPBpSSnHj9iq4HZQj7fxRzw2kj1kKSHHA", "service": "xmtp.message_api.v1.MessageApi", "method": "Publish", "client": "xmtp-js", "client_version": "xmtp-js/0.0.0-development", "app": "", "app_version": "", "client_ip": "172.22.0.1", "error": "rpc error: code = PermissionDenied desc = publishing to restricted topic", "error_code": "PermissionDenied", "error_message": "publishing to restricted topic"}
```

mkobetic:
I'm inclined to delete the offending test case, the only way to make it work is to spin up 2 different wallets and stores. I don't see what that would buy us.

mkobetic:
One thing I wonder is why was the test case not throwing the PermissionDenied exception on publishing 🤔

neekolas:
I've been meaning to change it, but we are still swallowing publish errors for legacy reasons. 

We should totally stop doing that

mkobetic:
@neekolas, RU ok just dropping this test case for now here? I'm going to look into this deeper in #226

mkobetic (approved):
Most of this got resolved in #227

PR #233, opened 2022-12-05
feat: add method to determine msg origin

jazzz:
This PR adds some ergonomics to working with DecodedMessages.

## Problem

In order to determine if the message in a conversation was sent by a client, developers need to keep access to a `Client` instance. This adds needless friction. 

#### Add a `clientAddress()` function to `Conversation` 
- developers can now access the clientAddress of the message by `decodedMessage.conversation.clientAddress()`

#### Add `isOutboundMessage()` to `DecodedMessage`
- as a connivence developers can now determine if a message was sent from their account directly on the decoded message. e.g `decodedMessage.isOutgoingMessage()` 

## Notes

- 'isOutboundMessage' does not necessarily imply that the message was not addressed to the owning account. It's possible that the client sent a message to its self.


src/Message.ts
```diff
@@ -279,6 +279,13 @@ export class DecodedMessage {
     this.content = content
     this.contentTopic = contentTopic
   }
+}
+
+export class DecodedMessage extends DecodedMessageBase {
+  // @returns true if the message sender is the same as the client which received the message.
+  isOutboundMessage(): boolean {
```
jazzz, line 286: Really wish this name was better. Proposed alternatives 'senderIsMe'. 

Thoughts suggestions?

jazzz:
@killthebuddh4 any thoughts on if this would be helpful? Any changes you'd make?

src/Message.ts
```diff
@@ -279,6 +279,13 @@ export class DecodedMessage {
     this.content = content
     this.contentTopic = contentTopic
   }
+}
+
+export class DecodedMessage extends DecodedMessageBase {
```
mkobetic, line 284: Curious, why you decided to split off the base class?

mkobetic (approved):
I like the naming you came up with. 👍

src/Message.ts
neekolas: 100% agree.

src/Message.ts
neekolas: Ya. I would just put this into the `DecodedMessage` type. I don't see what extending the types is buying you here

jazzz: Wasn't my first choice. Perhaps theres an easier way...

The `DecodedMessage` constructor takes a `DecodedMessage` as a parameter (rather than a defined interface. :: https://github.com/xmtp/xmtp-js/blob/main/src/Message.ts#L259-L270

When adding a function to `DecodedMessage` the type is changed and all previous instantiations fail with:
`Argument of type '{ id: string; messageVersion: "v1"; senderAddress: string; recipientAddress: string | undefined; sent: Date; content: any; contentType: ContentTypeId; contentTopic: string; conversation: Conversation; error: Error | undefined; }' is not assignable to parameter of type 'DecodedMessage'.
  Property 'isOutboundMessage' is missing in type '{ id: string; messageVersion: "v1"; senderAddress: string; recipientAddress: string | undefined; sent: Date; content: any; contentType: ContentTypeId; contentTopic: string; conversation: Conversation; error: Error | undefined; }' but required in type 'DecodedMessage'.ts(2345)`
  
- Creating a `DecodedMessageParams` interface requires both the Interface and Class requires properties to be defined in both and kept in sync.
- There looked to be a work around with  class/interface merging but implementing it was messier.   
- Flattening the constructor would be a breaking change.

Anything I am missing?

neekolas: Think you can use some fancy typing to get around that problem. If you used something like `Omit<DecodedMessage, 'isOutboundMessage'>` in the constructor would work.

killthebuddh4:
> @killthebuddh4 any thoughts on if this would be helpful? Any changes you'd make?

It surely would! I really appreciate that you tagged me also, thanks! I have some very minor thoughts regarding naming, I'll add them to the existing thread about naming.

src/Message.ts
jazzz: image

src/Message.ts
killthebuddh4: My initial thought was that I agree that I prefer `isOutboundMessage` but I changed my mind as I typed this comment :).  My new opinion is that "outbound" sort of couples the name to the context in which the method is called. For example, if I have a React component that's just listening to a messages stream and rendering them, I think my mental model of the messages would be that they're all _inbound_ but some of them were _sent by me_.

So all that said my new thought is that maybe this helper should also just be called `clientAddress` and instead of returning the `boolean` it could just return `this.conversation.clientAddress` directly.

mkobetic: I wouldn't object to having `clientAddress` on Messages as well, it doesn't seem much more out of place than it is on the Conversation. If we did want to preserve the conditional as well, maybe spelling out what it does could remove the contextual dependency somewhat, `senderIsTheClient`  maybe?

killthebuddh4: > If the address is ultimately what is wanted, I am inclined to skip the helper function all together.

I think I am also inclined in that direction for what it's worth.

jazzz: Awesome. Sounds like adding `conversation.clientAddress` eases most of the developer friction. We can revisit the helper-function on `DecodedMessage` in the future if there is demand.

PR #236, opened 2022-12-09
Refactor utils to multiple files

neekolas:
## Summary

Small refactor to break up the utils into multiple files. Should be a non-breaking change since imports do not need to change.


mkobetic (approved):
👍

PR #237, opened 2022-12-09
Add v2 conversation cache

neekolas:
## Summary

We are seeing performance degradations for users attempting to load a large number of conversations. Decrypting an invitation takes approximately 15ms on a M1 mac, so decrypting 100 can easily take multiple seconds depending on the device.

This is particularly noticeable when there are multiple code paths calling `conversations.list()`. A common case is the use of `conversations.streamAllMessages()`, which has a call to `conversations.list()` in its startup but does not remove the need to list all conversations. Applications that were offline for a period of time may also wish to call `conversations.list()` to get their internal store of conversations current with the network.

I implemented this using a cache and a mutex. The mutex is required to ensure that a second request to list conversations is not fired before the first completes. The cache keeps track of the most recently seen conversation so that subsequent calls can use that date as a `startTime` and only fetch/decrypt invites since the last query.

## Further improvements

- Applications with access to persistent storage could take an export of the cache and load it on startup. To do this, Conversations would need to be serializable.

### DIFF
--- a/src/conversations/Conversations.ts
+++ b/src/conversations/Conversations.ts
@@ -12,9 +12,15 @@ import { messageApi, fetcher } from '@xmtp/proto'
 import { SortDirection } from '../ApiClient'
 const { b64Decode } = fetcher
 
+const CLOCK_SKEW_OFFSET_MS = 10000
+
 const messageHasHeaders = (msg: MessageV1): boolean => {
   return Boolean(msg.recipientAddress && msg.senderAddress)
 }
+
+type CacheLoader = (args: {
+  latestSeen: Date | undefined
+}) => Promise<Conversation[]>
 
 class ConversationCache {
   private conversations: Conversation[]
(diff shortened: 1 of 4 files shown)


src/conversations/Conversations.ts
```diff
@@ -8,49 +9,108 @@ import { buildUserIntroTopic, buildUserInviteTopic } from '../utils'
...
     conversations.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime())
     return conversations
   }
 
+  private async listV2Conversations(): Promise<Conversation[]> {
+    const { latestSeen, release, addConvos } = await this.v2Cache.getAndLock()
+    try {
+      const newConvos: Conversation[] = []
+      const invites = await this.client.listInvitations({
+        startTime: latestSeen,
```
mkobetic, line 92: You're already handling dupes in the cache, we may want to add some fudge factor to lastSeen, to accommodate reasonable clock differences.

neekolas: Sounds reasonable

src/conversations/Conversations.ts
```diff
@@ -8,49 +9,108 @@ import { buildUserIntroTopic, buildUserInviteTopic } from '../utils'
...
-        console.warn('Error decrypting invitation', e)
-      }
-    }
+    const conversations = v1Convos.concat(v2Convos)
 
     conversations.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime())
     return conversations
   }
 
+  private async listV2Conversations(): Promise<Conversation[]> {
```
mkobetic, line 87: I think this works as expected. Wondering why not hide this inside the cache to avoid potential mistake of forgetting to release the lock. Not feeling strongly about this, feel free to keep as is.

neekolas: I considered having the cache take a callback function to execute so that it could handle the releasing itself. I'm still playing around with the ergonomics and might do that.

I do want this to be extensible enough that I can do the same thing for V1 conversations using the same cache class, which that would allow

mkobetic (approved):
I think this is great step in the right direction. I think we may want an internally managed convo stream as well, but that can be done as separate PR.

mkobetic:
Oops, sorry for jumping the shark, I only now noticed this is still a draft 🤦

src/conversations/Conversations.ts
neekolas: I updated to use a callback and I think that's a lot tidier. Makes it impossible for callers to leave the cache locked.

neekolas:
Alright @mkobetic. I've switched the cache API to take a loader callback and added a bunch of tests. Now I'm ready for you.

PR #239, opened 2022-12-14
fix: subscribe reconnect loop

snormore:
Updates subscribe reconnect loop so that it reconnects on both error and 200 completion.


test/ApiClient.test.ts
```diff
@@ -230,7 +230,7 @@ describe('Publish authn', () => {
   })
 })
 
-describe('Subscribe', () => {
+describe.only('Subscribe', () => {
```
neekolas, line 233: You'll want to remove the only before merging

neekolas (approved):
Thanks for catching and fixing this. Solution looks good to me

test/ApiClient.test.ts
snormore: 😅  good catch

PR #242, opened 2022-12-19
JSON Serializable Conversations

neekolas:
## Summary

I started this a few months ago, but decided to put it off until I was sure there was a clear use-case given the added complexity. 

I've now found three cases where JSON serializable conversations would be a significant improvement.

1. Caching conversations for decryption of push notifications. We have a short window of time to handle a push notification, and battery budgets to think about in mobile devices. Not having to do a network request or decrypt conversations would be a big help. The previous caching work assumes that there is an old client instance available.
2. Preloading the cache from LocalStorage for web based clients (Lenster/example app). We currently only cache conversations within the life of a client instance. Refreshing the page forces all conversations to be reloaded into the cache. Ideally, we would store conversations in `LocalStorage` and load the cache immediately on startup.
3. Some of the work I'm doing on MetaMask SNAPs could greatly benefit from JSON serializable conversations. Even better, SNAPs has their own encrypted storage solution so we have a secure place to store them.

## Includes
- Exporting and loading of V1 and V2 conversations
- Exporting of entire conversation list
- Import of entire conversation list
- Caching of V1 conversation calls

## Notes
It may be desirable to actually include an initial import in the client create options and only use the import functionality internally.

```ts
await Client.create(wallet, { initialConversations: someExport })
```

I have not implemented this yet, curious what others think.

### DIFF
--- a/src/conversations/Conversation.ts
+++ b/src/conversations/Conversation.ts
@@ -32,6 +32,23 @@ const { b64Decode } = fetcher
 
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 
+type ConversationV1Export = {
+  version: 'v1'
+  peerAddress: string
+  createdAt: Date
+}
+
+type ConversationV2Export = {
+  version: 'v2'
+  topic: string
+  keyMaterial: string
+  createdAt: Date
+  peerAddress: string
+  context: InvitationContext | undefined
+}
+
+export type ConversationExport = ConversationV1Export | ConversationV2Export
+
 /**
  * Conversation class allows you to view, stream, and send messages to/from a peer address
  */
@@ -84,6 +101,21 @@ export class ConversationV1 {
       [this.topic],
       this.decodeMessage.bind(this)
     )
+  }
+
+  export(): ConversationV1Export {
+    return {
+      version: 'v1',
+      peerAddress: this.peerAddress,
+      createdAt: this.createdAt,
+    }
+  }
+
+  static fromExport(
+    client: Client,
+    data: ConversationV1Export
+  ): ConversationV1 {
+    return new ConversationV1(client, data.peerAddress, data.createdAt)
   }
 
   async decodeMessage({
@@ -180,23 +212,25 @@ export class ConversationV1 {
 
 export class ConversationV2 {
   topic: string
-  keyMaterial: Uint8Array // MUST be kept secret
+  private keyMaterial: Uint8Array // MUST be kept secret
   context?: InvitationContext
-  private header: SealedInvitationHeaderV1
   private client: Client
+  createdAt: Date
   peerAddress: string
 
   constructor(
     client: Client,
-    invitation: InvitationV1,
-    header: SealedInvitationHeaderV1,
-    peerAddress: string
+    topic: string,
+    keyMaterial: Uint8Array,
+    peerAddress: string,
+    createdAt: Date,
+    context: InvitationContext | undefined
   ) {
-    this.topic = invitation.topic
-    this.keyMaterial = invitation.aes256GcmHkdfSha256.keyMaterial
-    this.context = invitation.context
+    this.topic = topic
+    this.keyMaterial = keyMaterial
+    this.createdAt = createdAt
+    this.context = context
     this.client = client
-    this.header = header
     this.peerAddress = peerAddress
   }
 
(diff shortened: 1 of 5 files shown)


src/conversations/Conversations.ts
```diff
@@ -61,33 +67,46 @@ export class ConversationCache {
...
-    )
-
     const conversations = v1Convos.concat(v2Convos)
 
     conversations.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime())
     return conversations
   }
 
+  private async listV1Conversations(): Promise<Conversation[]> {
+    return this.v1Cache.load(async ({ latestSeen }) => {
```
michaelx11, line 95: Neat change! Is it correct to say these v1 convos were not cached previously but are cached with this change?

src/conversations/Conversations.ts
```diff
@@ -408,6 +430,37 @@ export default class Conversations {
     )
   }
 
+  async export(): Promise<ConversationExport[]> {
+    const conversations = await this.list()
+    return conversations.map((convo) => convo.export())
+  }
+
+  async import(convoExports: ConversationExport[]): Promise<number> {
```
michaelx11, line 438: another dumb q: convos are exported in sorted order (`createdAt`) implicitly but there is no such requirement on import right? Just trying to think of any implicit invariants that could be gotchas in the future e.g. relied upon some ordering in exported data, ordering changes, imports break etc

michaelx11 (commented):
I'm not up for review but looks good!

src/conversations/Conversations.ts
neekolas: That's right

src/conversations/Conversations.ts
neekolas: The reason for that being that we merge together `v1` and `v2` conversations and want the final results to have a consistent sort.

src/conversations/Conversation.ts
```diff
@@ -360,6 +397,31 @@ export class ConversationV2 {
       error
     )
   }
+
+  export(): ConversationV2Export {
+    return {
+      version: 'v2',
+      topic: this.topic,
+      keyMaterial: Buffer.from(this.keyMaterial).toString('base64'),
```
elisealix22, line 405: This is a part where my lack of knowledge scares me a bit. I'm assuming this is ok because clients would otherwise already have access to this keyMaterial when decrypting conversations locally? Should we emphasize Conversations MUST be stored in a secure location similar to the warning we have in the ConversationV2 constructor?

src/conversations/Conversations.ts
```diff
@@ -61,33 +67,46 @@ export class ConversationCache {
...
+
+      return Array.from(seenPeers).map(
+        ([peerAddress, sent]) =>
+          new ConversationV1(this.client, peerAddress, sent)
+      )
+    })
+  }
+
   private async listV2Conversations(): Promise<Conversation[]> {
     return this.v2Cache.load(async ({ latestSeen }) => {
```
elisealix22, line 111: Just thinking out loud here, I wonder if there's a way for someone to only want to cache a few conversations (ones that they saw come thru on pushes or something). Should we indicate/document that the cache must be a complete cache or else anything older than the most recent will be dropped?

elisealix22 (approved):
Looks awesome and nothing blocking on my end! I think when we document the `{ initialConversations: someExport }` we may want to note that only conversations after the most recently created convo will be appended to the export. I could see people thinking it would fill in the gaps but I like the way you implemented it for performance reasons!

src/conversations/Conversation.ts
neekolas: That's a good point. Definitely deserves a section in the README that explains safe usage.

src/conversations/Conversations.ts
neekolas: Yes. I'll write up a section and add to the README

neekolas: Added a section to the README and included some code comments

mkobetic (approved):
This looks pretty good. The only thought I had was whether it would be worth modelling the import/export interface around some sort of source/sink of exports, rather then passing the full collections around, but it certainly isn't a concern now and hard to say whether it ever will be in the future.

src/conversations/Conversations.ts
michaelx11: Awesome, thanks!

PR #243, opened 2022-12-19
Test issue with newer versions of Node.js

neekolas:
## Summary

I noticed some tests failing locally that succeeded in CI. I narrowed it down to a difference in error output on newer versions of Node.

The test seems to work on `16.16.0` (the node version in our `.nvmrc`), but not on `16.18.0` or `18.x`. Interestingly, it seems to work on Node 17. Go figure.

The issue is that the error message for decryption errors has changed to `DOMException [OperationError]: The operation failed for an operation-specific reason`.

I've updated the test to be agnostic to the actual message and only verify that _an_ error has thrown.


michaelx11 (approved):
lgtm

PR #244, opened 2022-12-19
fix: unsubscribe in ApiClient tests to avoid leaking subscription loops

michaelx11:
# Summary

Running unit tests would result in repeated warnings of the form (below). I believe this is caused by not unsubscribing during tests in ApiClient.test.ts

## Changes
- Explicitly invoke unsubscribe function (returned by `subscribe`) in ApiClient unit tests
- Add what appears to be a missing `await stream.return()` in a conversation unit test
- Add eslint suppression for an `any` typed catch error

Warnings seen before:
```  ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "Stream connection closed. Resubscribing TypeError: fetchImpl.apply is not a function".

      201 |             return
      202 |           }
    > 203 |           console.info(
          |                   ^
      204 |             'Stream connection closed. Resubscribing',
      205 |             err.toString()
      206 |           )

      at console.info (node_modules/@jest/console/build/BufferedConsole.js:195:10)
      at ApiClient.<anonymous> (src/ApiClient.ts:203:19)
          at Generator.throw (<anonymous>)
      at rejected (src/ApiClient.ts:6:65)
```

## Test Plan

Observe no more warnings of the above form during testing

note: there are still open handle warnings that I may investigate further. After running with `--detectOpenHandles` looks like the complains lead back to `query` in ApiClient

### DIFF
--- a/src/ApiClient.ts
+++ b/src/ApiClient.ts
@@ -196,6 +196,7 @@ export default class ApiClient {
           if (new Date().getTime() - startTime < 1000) {
             await sleep(1000)
           }
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
         } catch (err: any) {
           if (isAbortError(err) || abortController.signal.aborted) {
             return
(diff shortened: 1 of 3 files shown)


neekolas (approved):
Great first PR @michaelx11!

elisealix22 (approved):

PR #246, opened 2022-12-22
Make newConversation cache-aware

neekolas:
## Summary

The `newConversation` method was not making use of the conversation caches. I've updated to use the existing `listV1Conversations()` and `listV2Conversations()` methods, in lieu of the lower level methods it was previously using.

The previous implementation skipped decrypting invitations for Conversations that did not match the intended recipient as a performance optimization. This method will load all new invitations, although it will use the cache to avoid decrypting any that have been seen before. This does make the worst-case performance worse (someone loads the SDK solely to create a new conversation), although the best-case performance should be better (creating a new conversation before or after running `conversations.list()`).

Both the old and new implementation will load/decrypt all V1 conversations on a cold start, although this implementation will only load/decrypt new introductions on all subsequent calls.

## Notes
In order to make the update transactional, I had to do a bit of a refactor to allow for `newConversation` to read from the cache, update it with items written since the last cache load, and potentially update with a newly created conversation before releasing the mutex.

I did that by breaking out the callback from `listV2Conversations` into a separate function that could be called from inside the cache loader, allowing for all logic to be executed before the mutex was released. This avoids race conditions where a second asynchronous process might create a conversation after the first releases the mutex.

Also worth noting that this will not prevent cases where a multiple client instances create a conversation with the same peer/conversationId at the same time.

## Resolves
Will resolve https://github.com/xmtp/example-chat-react/issues/168 (although we should also figure out why that hook is firing multiple times)

### DIFF
--- a/src/conversations/Conversations.ts
+++ b/src/conversations/Conversations.ts
@@ -450,6 +450,8 @@ export default class Conversations {
       }
     )
 
+    // Keep the typechecker happy
+    // v2Convo should never actually be undefined. An error in the loader will bubble and halt execution
     if (!v2Convo) {
       throw new Error('Failed to create conversation')
     }
--- a/src/conversations/Conversations.ts
+++ b/src/conversations/Conversations.ts
@@ -365,12 +365,15 @@ export default class Conversations {
       return new ConversationV1(this.client, peerAddress, new Date())
     }
 
+    // If no conversationId, check and see if we have an existing V1 conversation
     if (!context?.conversationId) {
-      const intros = await this.getIntroductionPeers()
-      const introSentTime = intros.get(peerAddress)
+      const v1Convos = await this.listV1Conversations()
+      const matchingConvo = v1Convos.find(
+        (convo) => convo.peerAddress === peerAddress
+      )
       // If intro already exists, return V1 conversation
       // if both peers have V1 compatible key bundles
-      if (introSentTime) {
+      if (matchingConvo) {
         if (!this.client.keys.getPublicKeyBundle().isFromLegacyBundle()) {
           throw new Error(
             'cannot resume pre-existing V1 conversation; client keys not compatible'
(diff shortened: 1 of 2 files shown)


src/conversations/Conversations.ts
```diff
@@ -43,7 +44,10 @@ export class ConversationCache {
   async load(loader: CacheLoader) {
     const release = await this.mutex.acquire()
     try {
-      const newConvos = await loader({ latestSeen: this.latestSeen })
+      const newConvos = await loader({
+        latestSeen: this.latestSeen,
+        existing: this.conversations,
```
neekolas, line 49: I now provide the existing items in the cache to the loader function, so that `newConversation` can avoid doing unnecessary network requests if the existing conversation is already cached.

src/conversations/Conversations.ts
```diff
@@ -108,28 +112,35 @@ export default class Conversations {
...
-        try {
-          const unsealed = await sealed.v1.getInvitation(this.client.keys)
-          newConvos.push(
-            await ConversationV2.create(this.client, unsealed, sealed.v1.header)
-          )
-        } catch (e) {
-          console.warn('Error decrypting invitation', e)
-        }
+  // Callback called in listV2Conversations and in newConversation
+  private async v2ConversationLoader(
```
neekolas, line 121: I needed to break this into its own function to avoid a deadlock that would be caused by calling `listV2Conversations` from inside a conversation loader.

src/conversations/Conversations.ts
```diff
@@ -393,41 +407,53 @@ export default class Conversations {
...
+          this.client,
+          invitation,
+          sealedInvite.v1.header
+        )
+
+        return [v2Convo]
+      }
     )
+
+    if (!v2Convo) {
```
neekolas, line 453: This is just there to keep the type checker happy. Should be impossible, since an error inside the loader would bubble up and halt execution before it gets here.

mkobetic: I'd add this as a comment, it's not obvious.

mkobetic (approved):
Nice 👍

PR #247, opened 2022-12-22
fix: remove dependency on ethers from public interface

richardhuaaa:
Don't force client apps to take a direct dependency on ethers. Note we will still depend on ethers internally within the SDK for some address util methods.

I added a test, and manually verified sending still works.

Fixes [#887](https://github.com/xmtp-labs/hq/issues/887)

### DIFF
--- a/src/crypto/PrivateKeyBundle.ts
+++ b/src/crypto/PrivateKeyBundle.ts
@@ -3,6 +3,7 @@ import { PrivateKey, SignedPrivateKey } from './PrivateKey'
 import { WalletSigner } from './Signature'
 import { PublicKey, SignedPublicKey } from './PublicKey'
 import { PublicKeyBundle, SignedPublicKeyBundle } from './PublicKeyBundle'
+import { Signer } from '../types/Signer'
 import { NoMatchingPreKeyError } from './errors'
 
 // PrivateKeyBundle bundles the private keys corresponding to a PublicKeyBundle for convenience.
--- a/src/types/Signer.ts
+++ b/src/types/Signer.ts
@@ -1,4 +1,4 @@
-interface Signer {
+export interface Signer {
   getAddress(): Promise<string>
   signMessage(message: ArrayLike<number> | string): Promise<string>
 }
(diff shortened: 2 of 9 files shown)


saulmc:
Yessss our web3.js frens will be so happy /cc @yash-luna

neekolas:
How hard would it be to break out the address utils and remove the ethers dependency entirely? It should decrease our bundle size and remove complexity from our dependency tree if we can kill it outright.

jhaaaa:
Thank you for tagging me @richardhuaaa - I'm so grateful for your updates to the README! I will keep an eye on this PR and will plan to update the docs on xmtp.org accordingly.

mkobetic (approved):
❤️ We should have done this long time ago.

src/crypto/PrivateKeyBundle.ts
```diff
@@ -3,7 +3,6 @@ import { PrivateKey, SignedPrivateKey } from './PrivateKey'
 import { WalletSigner } from './Signature'
 import { PublicKey, SignedPublicKey } from './PublicKey'
 import { PublicKeyBundle, SignedPublicKeyBundle } from './PublicKeyBundle'
-import { Signer } from 'ethers'
```
neekolas, line 6: I don't love the fact that we are exporting this type globally. Feels safer and easier to understand to actually import the type upon usage. Right now it just appears like magic, which is going to cause developer confusion.

I also don't remember where we configured this magic, so that things from the `types` folder were available in every module. I think it makes sense for something like a polyfill, but less so for an application-specific type like this.

src/types/Signer.ts
```diff
@@ -0,0 +1,4 @@
+interface Signer {
```
neekolas, line 1: We are going to need to export this type from `index.ts` so that client applications can easily make use of it.

src/crypto/PrivateKeyBundle.ts
neekolas: > I don't think this is special to the types folder - if you don't export or import from within any ts file it becomes global.

How did I not know that?

neekolas (approved):

neekolas:
> We are actually using a bunch of methods from ethers, I don't think it's wise to copy paste them all.

That's more than I realized. Seems reasonable to keep the dependency in that case.

PR #249, opened 2022-12-23
test: clean unit test console warnings

richardhuaaa:
The unit tests are spitting out console warnings that look like errors:

Screen Shot 2022-12-22 at 10 44 36 AM

I'm a bit worried that this noise can cause us to overlook true errors, as well as this generally being confusing for newbies. As these logs are expected given the unit tests are validating failure cases, I added a change to have the unit tests explicitly verify that these logs are being emitted, but swallow them so that we don't have to see them.

There is still some console output left over after this fix - an import error, and some connection info logs, these can be addressed separately.

### DIFF
--- a/test/conversations/Conversation.test.ts
+++ b/test/conversations/Conversation.test.ts
@@ -104,6 +104,9 @@ describe('conversation', () => {
     })
 
     it('ignores failed decoding of messages', async () => {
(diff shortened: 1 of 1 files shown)


neekolas (approved):
This seems like a sensible way to reduce the noise. Love it.

PR #250, opened 2022-12-23
test: fix async networking issues in unit tests

richardhuaaa:
#249 and #244 helped a lot, but there's still a bunch of leftover noise coming from the unit tests. The last piece is fixing our networking mocks so that the tests run cleanly without any errors.

Before:
Screen Shot 2022-12-23 at 3 37 05 PM

After:
Screen Shot 2022-12-23 at 3 42 21 PM

### DIFF
--- a/test/ApiClient.test.ts
+++ b/test/ApiClient.test.ts
@@ -269,6 +269,9 @@ describe('Subscribe', () => {
           cb: NotifyStreamEntityArrival<messageApi.Envelope> | undefined,
           initReq?: InitReq
         ): Promise<void> => {
+          // We mock a connection stream that immediately errors the first time
+          // it is called. The second time it is called, it behaves as expected (the connection
+          // stays open, and two messages are received over the subscription)
           called++
           if (called == 1) {
             throw new Error('error')
@@ -311,6 +314,9 @@ describe('Subscribe', () => {
           cb: NotifyStreamEntityArrival<messageApi.Envelope> | undefined,
           initReq?: InitReq
         ): Promise<void> => {
+          // We mock a connection stream that immediately terminates the first time
+          // it is called. The second time it is called, it behaves as expected (the connection
+          // stays open, and two messages are received over the subscription)
           called++
           if (called == 1) {
             return
@@ -400,6 +406,9 @@ function createSubscribeMock(numMessages: number) {
     .mockImplementation(subscribeMockImplementation(numMessages))
 }
 
+// Subscribes to a connection stream that pushes down the number of messages specified.
+// The connection stream is expected to stay open until it is closed by an unsubscribe
+// request (which is reflected by the 'onabort' signal)
 function subscribeMockImplementation(numMessages: number) {
   let subscribe = async (
     req: messageApi.SubscribeRequest,


mkobetic (approved):
I can't say I grok the fix, but I'm definitely grateful for the resolution. ❤️

test/ApiClient.test.ts
```diff
@@ -257,25 +260,7 @@ describe('Subscribe', () => {
...
-    const subscribeMock = jest
-      .spyOn(MessageApi, 'Subscribe')
-      .mockImplementation(
-        async (
-          req: messageApi.SubscribeRequest,
-          cb: NotifyStreamEntityArrival<messageApi.Envelope> | undefined
-        ): Promise<void> => {
-          called++
-          if (called == 1) {
-            throw new Error('error')
```
michaelx11, line 270: dumb question: are we sure this isn't deleting an expected error? My initial read is this differs from `createSubscribeMock` in that it's supposed to error, then trigger a resubscribe. The noisy logs are annoying but might be expected

test/ApiClient.test.ts
```diff
@@ -296,25 +282,7 @@ describe('Subscribe', () => {
...
-    let called = 0
-    const subscribeMock = jest
-      .spyOn(MessageApi, 'Subscribe')
-      .mockImplementation(
-        async (
-          req: messageApi.SubscribeRequest,
-          cb: NotifyStreamEntityArrival<messageApi.Envelope> | undefined
-        ): Promise<void> => {
-          called++
-          if (called == 1) {
```
michaelx11, line 308: same question here: was the previous "return" behavior intended to differ from the normal `createSubscribeMock` behavior? @snormore worth getting your eyes before merge

test/ApiClient.test.ts
```diff
@@ -263,19 +268,18 @@ describe('Subscribe', () => {
...
           called++
           if (called == 1) {
             throw new Error('error')
           }
-          for (let i = 0; i < 2; i++) {
-            if (cb) {
-              cb(createEnvelope())
-            }
-          }
+          let nonErroringSubscribe = subscribeMockImplementation(2)
```
michaelx11, line 278: I don't fully understand what this change does. So first we start with a Subscribe mock that errors after one call, but before it gets there it'll instead return a non-erroring subscribe mock?

Does it still end up throwing the error and triggering a resubscribe?

michaelx11 (commented):
Appreciate the rework! I still don't fully understand the changes. Maybe some comments would be helpful for future readers. I add my questions inline.

test/ApiClient.test.ts
```diff
@@ -284,6 +286,9 @@ describe('Subscribe', () => {
     const unsubscribeFn = client.subscribe(req, cb)
     await sleep(1200)
     expect(numEnvelopes).toBe(2)
+    // Resubscribing triggers an info log
+    expect(consoleInfo).toBeCalledTimes(1)
```
michaelx11, line 290: Also somehow lost this comment. I don't think we should be relying on expecting console logs for a test. In the future, someone could very reasonably remove the log statement and add another one elsewhere, such that behavior changes but the tests still pass.

test/ApiClient.test.ts
michaelx11: Okay I think I understand better. On the first subscribe request, this mock will invoke `nonErroringSubscribe` which should call `cb` with 2 messages. The second time `subscribe` is invoked it'll throw an error.

michaelx11: This should match the original behavior, looks good.

test/ApiClient.test.ts
michaelx11: I see, it's not a test requirement but just meant to suppress the console logs. In that case looks good to me.

michaelx11 (approved):

PR #255, opened 2023-01-16
feat: enforce 2party dh

jazzz:
## Problem: 
Deriving shared keys for a conversation with "yourself" can lead to weak keys ought not be allowed in the SDK.

## Solution:
SDK ought to throw an error when attempting to derive keys using the public and private keys from the same keypair. 

## Notes:
This is a breaking change. Applications which allowed users to send messages to themself will no longer be able to derive keys to decrypt those messages. This behavior is explicitly banned in the XMTP Example apps however may exist in other implementations.

### DIFF
--- a/src/crypto/PrivateKey.ts
+++ b/src/crypto/PrivateKey.ts
@@ -23,6 +23,29 @@ function secp256k1Check(key: secp256k1): void {
   }
 }
 
+function deriveSharedSecret(
+  privateKey: SignedPrivateKey | PrivateKey,
+  publicKey: SignedPublicKey | UnsignedPublicKey
+): Uint8Array {
+  // Single party DHKE should be avoided. Throw error if publickey matches this keypair
+  if (
+    equalBytes(
+      privateKey.publicKey.secp256k1Uncompressed.bytes,
+      publicKey.secp256k1Uncompressed.bytes
+    )
+  ) {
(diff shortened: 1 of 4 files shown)


michaelx11 (approved):
Overall LGTM, can we add a short justification for why self-ECDH is problematic? Would be helpful for future readers

Hmac512:
A self ECDH is not mathematically problematic. 

It’s equivalent to squaring the private key. The resulting key’s public key is used to generate an AES key.

This is non-standard though.

For generating a key for your own use, usually you want to use ECIES. In which you generate a new private key, and its corresponding public key. Then you do ECDH to get a shared secret to encrypt your data. The public key is stored with the cipher-text so you can generate the key without keeping the ephemeral private key.

ECIES is used because you can generate many different encryption keys if you don’t want to use the same AES-GCM key.


The “self-ECDH” is equivalent to the squared computational diffie-hellman problem. In this paper they prove it’s the same security as the normal CDH assumption: https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.104.3007&rep=rep1&type=pdf


—-

Cryptographically it’s secure.

I feel like ECIES should be good to have in the SDK anyway.

rygine:
closing this PR due to inactivity. please reach out if you have any questions.

PR #260, opened 2023-01-19
docs: Fix compression docs in README

nakajima:
Using a string errors in Typescript:

Screen Shot 2023-01-19 at 10 47 42 AM

PR #261, opened 2023-01-19
fix: Fix deflate compression

nakajima:
Client.encodeContent was checking `options?.compression`, which in the case of `Compression.COMPRESSION_DEFLATE` equals `0`, which is falsey in JS.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,8 +2,6 @@
 
 If you're seeing this document, you are an early contributor to the development and success of XMTP. We welcome your questions, feedback, suggestions, and code contributions.
 
-> If you wish to contribute, please [apply for Early Access](https://xmtp.typeform.com/yojTJarb) so that we can provide you with access to our Discord.
-
 ## ❔ Questions
 
 Have a question? We welcome you to ask it in [Q&A discussions](https://github.com/orgs/xmtp/discussions/categories/q-a).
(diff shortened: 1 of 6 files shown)


mkobetic (approved):
Good catch. Thank you! ❤️

nakajima:
@mkobetic Thanks for the review! I think I found another issue that I pushed a fix for in https://github.com/xmtp/xmtp-js/pull/261/commits/859ae756b9a0ef7504075c3ac74b9a7d12d9d6b7 if you don't mind taking a look at that as well.

mkobetic:
> @mkobetic Thanks for the review! I think I found another issue that I pushed a fix for in [859ae75](https://github.com/xmtp/xmtp-js/commit/859ae756b9a0ef7504075c3ac74b9a7d12d9d6b7) if you don't mind taking a look at that as well.

Ouch, was this really completely broken for V2 messages? Seems that way 😞. Thank you so much for getting to the bottom of this. ❤️

mkobetic:
@neekolas do you remember what happened here? I can't believe this was broken from the beginning.

mkobetic:
OK, IIUC the decoding was broken for both versions, so the thing I'm puzzled by is how did the v1 compression test pass? Otherwise the fixes look good to me.

nakajima:
> so the thing I'm puzzled by is how did the v1 compression test pass? 

I was wondering about that too. I don't think we have an assertion that says that anything was actually compressed to begin with in that code path. I can add some additional checks in there to make sure things are happening the way we expect. I'm glad I decided to work on adding compression to the iOS sdk today, I'm learning so much!

nakajima:
Also the test failures don't seem super related to my changes here (they actually look similar to what's happening in https://github.com/xmtp/xmtp-js/pull/260, which is just a docs change). Any insight into what might be happening there would be super appreciated.

mkobetic:
> > so the thing I'm puzzled by is how did the v1 compression test pass?
> 
> I was wondering about that too. I don't think we have an assertion that says that anything was actually compressed to begin with in that code path. I can add some additional checks in there to make sure things are happening the way we expect. I'm glad I decided to work on adding compression to the iOS sdk today, I'm learning so much!

Oh, I see, I think you're right. If DEFLATE turned into a noop due to the original bug you were fixing, then the compression didn't happen, and so the received content was the same as the sent content event though we didn't decompress at all. That would explain it.

mkobetic:
> Also the test failures don't seem super related to my changes here (they actually look similar to what's happening in #260, which is just a docs change). Any insight into what might be happening there would be super appreciated.

I haven't touched the SDK in a while, I'm hoping someone else is more up to date here. But if needed I can dig into it tomorrow.

michaelx11:
This merged PR might fix tests: https://github.com/xmtp/xmtp-js/pull/262 - try updating the branch

nakajima:
@michaelx11 That did the trick, thanks!

PR #265, opened 2023-01-26
feat: add canMessage multiple powered by batch query

michaelx11:
**Goal**

Now that we BatchQueries in [node API](https://github.com/xmtp/xmtp-node-go/pull/202) we can begin to use the capability in the xmtp-js SDK. This first change will create a method overload for `canMessage` which takes multiple addresses and executes them via batch queries, returning a list of booleans as a Promise.

**Changes**
- Implement BatchQuery in apiClient.ts
- Implement BatchQuery in Client.ts
- Add `getUserContactsFromNetwork` (plural)
- Add overloaded type signatures for `canMessage`

**Testing**
- Unit tests

**Outdated Notes**

My question: `canMessage` isn't really used in a multiple address case by apps today right? I'm looking at ListConversations as the biggest bottleneck that could benefit from batch querying. If I use "Fast 3G" with 1000 convos, the network stalling can be 6+ seconds. On 500Mb/s network, the decryption of envelopes takes enough time that network stall is < 1 second.

Turns out folks want the use case of being able to see what addresses they can message from some set (e.g. these addresses own this NFT)

### DIFF
--- a/src/ApiClient.ts
+++ b/src/ApiClient.ts
@@ -320,9 +320,9 @@ export default class ApiClient {
     }
   }
 
-  // Take a list of queries and execute them in batches. queryOptions can be empty
+  // Take a list of queries and execute them in batches
   async batchQuery(queries: Query[]): Promise<messageApi.Envelope[][]> {
-    // Group content topics into batches of 50 (implicit server-side limit) and then perform BatchQueries
+    // Group queries into batches of 50 (implicit server-side limit) and then perform BatchQueries
     const BATCH_SIZE = 50
     // Keep a list of BatchQueryRequests to execute all at once later
     const batchRequests: messageApi.BatchQueryRequest[] = []
@@ -331,7 +331,7 @@ export default class ApiClient {
     for (let i = 0; i < queries.length; i += BATCH_SIZE) {
       const queriesInBatch = queries.slice(i, i + BATCH_SIZE)
       // Perform batch query by first compiling a list of repeated individual QueryRequests
-      // and then calling the BatchQuery API with that list of requests in a BatchQueryRequest
+      // then populating a BatchQueryRequest with that list
       const constructedQueries: messageApi.QueryRequest[] = []
 
       for (const queryParams of queriesInBatch) {
@@ -357,17 +357,20 @@ export default class ApiClient {
       batchRequests.map(async (batch) => this._batchQuery(batch))
     )
 
-    // For every batch, read all responses within the batch, and add them to a list of list of envelopes
+    // For every batch, read all responses within the batch, and add to a list of lists of envelopes
     // one top-level list for every original query
     const allEnvelopes: messageApi.Envelope[][] = []
     for (const batchResponse of batchQueryResponses) {
       if (!batchResponse.responses) {
+        // An error on any of the batch query is propagated to the caller
+        // for simplicity, rather than trying to return partial results
         throw new Error('BatchQueryResponse missing responses')
       }
       for (const queryResponse of batchResponse.responses) {
         if (queryResponse.envelopes) {
           allEnvelopes.push(queryResponse.envelopes)
         } else {
+          // If no envelopes provided, then add an empty list
           allEnvelopes.push([])
         }
       }
--- a/src/ApiClient.ts
+++ b/src/ApiClient.ts
@@ -56,6 +56,9 @@ export type QueryStreamOptions = Omit<QueryAllOptions, 'limit'> & {
   pageSize?: number
 }
 
+// All of the fields in both QueryParams and QueryStreamOptions
+export type Query = QueryParams & QueryStreamOptions
+
 export type PublishParams = {
   contentTopic: string
   message: Uint8Array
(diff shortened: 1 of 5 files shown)


src/ApiClient.ts
```diff
@@ -298,6 +317,81 @@ export default class ApiClient {
...
+    const batchTopics: string[][] = []
+
+    // Assemble batches
+    for (let i = 0; i < contentTopics.length; i += BATCH_SIZE) {
+      const topicsInBatch = contentTopics.slice(i, i + BATCH_SIZE)
+      batchTopics.push(topicsInBatch)
+      // Perform batch query by first compiling a list of repeated individual QueryRequests
+      // and then calling the BatchQuery API with that list of requests in a BatchQueryRequest
+      const queryRequests: messageApi.QueryRequest[] = []
+      for (let j = 0; j < topicsInBatch.length; j++) {
```
neekolas, line 344: This would be more readable as
```
for (const topic of topicsInBatch) {
...
}
```

michaelx11:
Adding some video evidence of a performance test with 1002 addresses that are present on dev network.

Without batch query (just invoking canMessage over and over). Trying all 1002 will not complete for over a minute at this rate.

https://user-images.githubusercontent.com/1329295/214957536-87c3e597-422f-4cb8-8274-bd540ff617e3.mov

With batchQuery (call `canMessage` once with all 1002 addresses). Completes in a handful of seconds.

https://user-images.githubusercontent.com/1329295/214957554-9170794e-264a-46c1-a23f-1461c21ac4a2.mov

src/Client.ts
```diff
@@ -272,6 +271,50 @@ export default class Client {
...
+    // The logic here is tricky because we need to do a batch query for any uncached bundles,
+    // then interleave back into an ordered array. So we create a map<string, keybundle|undefined>
+    // and fill it with cached values, then take any undefined entries and form a BatchQuery from there.
+    const addressToBundle = new Map<
+      string,
+      PublicKeyBundle | SignedPublicKeyBundle | undefined
+    >()
+    const uncachedAddresses = []
+    for (const address of normalizedAddresses) {
+      const existingBundle = this.knownPublicKeyBundles.get(address)
```
neekolas, line 293: I see us reading from `knownPublicKeyBundles`, but don't see us writing the results back in. WDYT about storing the results to save on future requests.

michaelx11: Good call, I'll make the change

src/Client.ts
```diff
@@ -280,12 +323,22 @@ export default class Client {
     this.knownPublicKeyBundles.delete(peerAddress)
   }
 
+  public async canMessage(peerAddress: string): Promise<boolean>
```
neekolas, line 326: Good call on the overloads here. I noticed there aren't overloads on the static method. We'll want those so that `Client.canMessage('foo')` has a return type of `boolean` instead of `boolean | boolean[]`

test/Client.test.ts
```diff
@@ -124,6 +124,68 @@ describe('canMessage', () => {
...
+    const registeredClients = await Promise.all(
+      Array.from({ length: 10 }, () => newLocalHostClient())
+    )
+    // Wait for all clients to be registered
+    await Promise.all(
+      registeredClients.map((client) => waitForUserContact(client, client))
+    )
+    // Now call canMessage with all of the peerAddresses
+    const canMessageRegisteredClients = await Client.canMessage(
+      [...registeredClients.map((client) => client.address)],
```
neekolas, line 139: Think this line could be simplified to
`registeredClients.map((client) => client.address)`

src/ApiClient.ts
```diff
@@ -298,6 +317,81 @@ export default class ApiClient {
     }
   }
 
+  // Get first pageSize envelopes for a number of different topics, intended to efficiently
+  // use the batch-query API in a targeted manner. Returns a map of topic to message envelopes
+  async batchQuery(
+    { contentTopics, startTime, endTime }: QueryParams,
```
neekolas, line 323: It feels a little funny to me that this method expects all queries in the batch to have the same `startTime`, `endTime`, `pageSize`, and `sortDirection`. That works fine for the current use-case, and likely for the next one, but give that this is the only exposed method to access the `batchQuery` endpoint I wonder if we want to be so prescriptive.

michaelx11: I can see that, especially with the defaults the common case of identical params should still be pretty simple. I'll make the change to take a list of QueryParams

michaelx11: I think I'll have to change the return type from `Promise<Map<string, envelopes[]>>` to just `Promise<envelopes[][]>`. The idea is that `batchQuery` just takes a list of normal queries and acts as `query` would.

neekolas: Ya. I think we can take this higher level abstraction of assuming the queries have the same parameters and just put it elsewhere.

src/ApiClient.ts
```diff
@@ -298,6 +317,79 @@ export default class ApiClient {
...
+  // Take a list of queries and execute them in batches. queryOptions can be empty
+  async batchQuery(
+    queries: QueryParams[],
+    queryOptions?: QueryStreamOptions[]
+  ): Promise<messageApi.Envelope[][]> {
+    // Require length of queries and queryOptions to be the same, or queryOptions to be empty
+    if (
+      queryOptions &&
+      queryOptions.length &&
+      queryOptions.length !== queries.length
```
neekolas, line 329: Having the query be in two arrays feels a little awkward for developers to actually use.

It feels like something closer to the proto `BatchQueryRequest` would be easier to work with than having to construct the two arrays of the same length

michaelx11: makes sense, I'll probably need to create a new type then? It did seem weird that our queries are split into `QueryParams` and `QueryAllOptions | QueryStreamOptions`

michaelx11: The `batchQuery` signature could be:

```
type Query {
  // union of fields from QueryParams and QueryAllOptions
}

async batchQuery(queries: Query[]): Promise<messageApi.Envelope[][]>
```

@neekolas thoughts?

src/Client.ts
```diff
@@ -532,3 +620,38 @@ async function getUserContactFromNetwork(
...
+  // by going through each message and attempting to decode
+  return Promise.all(
+    peerAddresses.map(async (address: string, index: number) => {
+      const envelopes = topicToEnvelopes[index]
+      if (!envelopes) {
+        return undefined
+      }
+      for (const env of envelopes) {
+        if (!env.message) continue
+        const keyBundle = decodeContactBundle(b64Decode(env.message.toString()))
```
neekolas, line 648: Think we will want to wrap this in a try/catch. A single undecodable bundle would bubble up and cause the entire `Promise.all` to fail

src/ApiClient.ts
```diff
@@ -298,6 +320,62 @@ export default class ApiClient {
...
+    const batchRequests: messageApi.BatchQueryRequest[] = []
+
+    // Assemble batches
+    for (let i = 0; i < queries.length; i += BATCH_SIZE) {
+      const queriesInBatch = queries.slice(i, i + BATCH_SIZE)
+      // Perform batch query by first compiling a list of repeated individual QueryRequests
+      // and then calling the BatchQuery API with that list of requests in a BatchQueryRequest
+      const constructedQueries: messageApi.QueryRequest[] = []
+
+      for (let j = 0; j < queriesInBatch.length; j++) {
```
neekolas, line 337: This could be simplified to
```
for (const queryParams of queriesInBatch) {
...
}
```

neekolas (approved):
LGTM. Only real concern is the error propagation in `getUserContactsFromNetwork`

src/ApiClient.ts
neekolas: This looks great.

src/Client.ts
michaelx11: Good catch, I'll make the change and see if it's easy to add a test for it too

michaelx11:
Just did an extra out-of-bad test with invalid contact bundles, confirmed that only the invalid ones are marked false rather than the entire `canMessage(many)` request failing.

neekolas (approved):
Great stuff. Expands the capabilities of the SDK, and in a performant way. :shipit:

PR #270, opened 2023-02-08
Message Backups: Configure XMTP backups for localhost

richardhuaaa:
This is the beginning of a rudimentary post-delivery storage feature. In the long-term, this feature will be user-configurable based on their security preference - message backups can potentially be stored on XMTP, on local disk, disabled, on iCloud, etc. The user will configure this preference in any app, we will store this preference on the server (alongside a user signature), and this preference will follow the user around to all apps that they use.

How this PR works:
1. We define a `BackupClient` interface that is implemented by `XMTPBackupClient` and `NoBackupClient`.
2. When creating a `BackupClient`, we'll first check if a preference already exists on the server (TODO), and if not, we'll ask for a preference using the `selectBackupProvider` callback, before storing it on the server (TODO).
3. We'll use this preference to create the correct backup client.

While this feature is being developed, I'm just internally hard-coding `selectBackupProvider` to choose XMTP backups if we're on localhost, and disable them in all other cases. In the long-term, `selectBackupProvider` can be passed down to us from the app itself, defaulting to XMTP if it is omitted.

The encryption scheme we'll use for backups is super basic right now (hardcoded encryption key used only on localhost), we can iterate on this later.

### DIFF
--- a/src/message-backup/BackupClient.ts
+++ b/src/message-backup/BackupClient.ts
@@ -12,9 +12,11 @@ export type SelectBackupProvider = () => Promise<BackupProvider>
 
 export interface NoBackupConfiguration {
   type: BackupType.none
+  version: number
 }
 export interface TopicStoreBackupConfiguration {
   type: BackupType.xmtpTopicStore
+  version: number
   // The location where the backup will be stored
   topic: string
   // The symmetric encryption key used to encrypt/decrypt backups (optional for now)
--- a/src/message-backup/NoBackupClient.ts
+++ b/src/message-backup/NoBackupClient.ts
@@ -7,6 +7,7 @@ export default class NoBackupClient implements BackupClient {
   public static createConfiguration(): NoBackupConfiguration {
     return {
       type: BACKUP_TYPE,
+      version: 0,
     }
   }
 
--- a/src/message-backup/TopicStoreBackupClient.ts
+++ b/src/message-backup/TopicStoreBackupClient.ts
@@ -13,7 +13,8 @@ export default class TopicStoreBackupClient implements BackupClient {
     // TODO: randomly generate topic and encryption key
     return {
       type: BACKUP_TYPE,
-      topic: 'history-v0-0:' + walletAddress,
+      version: 0,
+      topic: 'history-v0:' + walletAddress,
     }
   }
 


src/Client.ts
```diff
@@ -187,7 +200,11 @@ export default class Client {
     const apiClient = createApiClientFromOptions(options)
     const keys = await loadOrCreateKeysFromOptions(options, wallet, apiClient)
     apiClient.setAuthenticator(new Authenticator(keys.identityKey))
-    const client = new Client(keys, apiClient)
+    const backupClient = await Client.setupBackupClient(
+      keys.identityKey,
```
neekolas, line 204: How is this going to work in a world where the `Client` doesn't have access to the IdentityKey itself? Once the KeyStore work goes in, the `IdentityKey` will only live inside the KeyStore, and the KeyStore may be remote (for example, MM Snap)

src/Client.ts
```diff
@@ -200,6 +217,19 @@ export default class Client {
...
   }
 
+  private static async setupBackupClient(
+    identityKey: PrivateKey,
+    env: keyof typeof ApiUrls
+  ): Promise<BackupClient> {
+    // Hard-code the provider to use for now
+    const backupProviderSelector = async () => {
+      return Promise.resolve(
+        env === 'local' ? BackupProvider.xmtp : BackupProvider.none
```
neekolas, line 227: I'm not sure why we would only do this when connecting to `localhost`. It seems equally easy to make the backup provider a `ClientCreateOption` and just default it to none for now. 

Feels like that is going to be easier to write test cases against, and is a more explicit/simpler mental model.

src/message-backup/BackupClient.ts
```diff
@@ -0,0 +1,20 @@
...
+  xmtp,
+}
+export type SelectBackupProvider = () => Promise<BackupProvider>
+
+export interface BackupConfiguration {
+  provider: BackupProvider
+  // For the XMTP BackupProvider, this is the name of a topic
+  location?: string
+  // For the XMTP BackupProvider, this is the symmetric encryption key used to encrypt/decrypt backups
+  encryptionKey?: string
```
mkobetic, line 15: Why string? Using strings for keys usually makes me suspicious as they bring in ambiguity about how exactly will they be turned into key bytes.

src/message-backup/BackupClient.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Where message backups should be stored
+ */
+export enum BackupProvider {
```
mkobetic, line 4: this is nitpick, but I'd expect `BackupProvider` to be an interface or something more substantial. This seems to be a provider id or type at this point. Do you expect this to evolve into something bigger down the line?

src/message-backup/BackupClientFactory.ts
```diff
@@ -0,0 +1,53 @@
...
+): Promise<BackupClient> {
+  const configuration = await fetchOrCreateConfiguration(
+    walletAddress,
+    selectBackupProvider
+  )
+  switch (configuration.provider.type) {
+    case BackupType.none:
+      return new NoBackupClient(configuration)
+    case BackupType.xmtp:
+      return new XMTPBackupClient(configuration)
```
jazzz, line 32: What is the significance of XTMP in the name XMTPBackupClient. What other BackupClients do you imagine in the future?

src/message-backup/BackupClient.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+export type SelectBackupProvider = () => Promise<BackupProvider>
+
+export interface BackupConfiguration {
+  provider: BackupProvider
+  // For the XMTP BackupProvider, this is the name of a topic
+  location?: string
+  // For the XMTP BackupProvider, this is the symmetric encryption key used to encrypt/decrypt backups
+  secret?: Uint8Array
+}
```
jazzz, line 19: Is this intention here that all backupConfigurations will follow this format? or will the configuration format be dependent upon the provider?

src/message-backup/BackupClient.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Where message backups should be stored
+ */
+export enum BackupType {
+  none,
+  xmtp,
```
michaelx11, line 6: nit: I thought enum members should always be capitalized but looks like we're already inconsistent with this repo.

Should we rename `xmtp` to `topicStore`? Kind of mirroring the:
```
export enum KeyStoreType {
  networkTopicStoreV1,
  static,
}
```

michaelx11 (approved):
overall LGTM. Is the intended future use going to be an invocation of `backupClient.backupMessage(...)` within the Client class? Or not fully locked yet?

src/message-backup/BackupClient.ts
michaelx11: My main thinking being an "xmtp" type backup doesn't tell me much as an external person. I'd probably think "yeah of course this is an XMTP backup", when really it should communicate the type/location of their "XMTP" backup. So like "xmtpTopicStore" or something

src/Client.ts
jazzz: I also don't quite follow this. 

I'm not sure I understand the benefit, and it seems like this would complicate the tests by hardcoding the environment.

src/message-backup/BackupClient.ts
jazzz: I had the same confusion. In a world where there may be many different backupClients, specificity now will help.

src/message-backup/BackupClient.ts
jazzz: If the Backup configuration is going to be persisted to the network, we'll likely want an associated version so the configuration could be self describing.

src/message-backup/BackupClient.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+export interface BackupProvider {
+  type: BackupType
+}
+export type SelectBackupProvider = () => Promise<BackupProvider>
+
+export interface BackupConfiguration {
```
jazzz, line 13: What is the relation ship between BackupConfiguration and Account. Is it a strict 1:1 relationship?

src/Client.ts
jazzz: I won't block it, I trust your judgement here. 

I don't particularly see a need to, keep these changes off the development network is all. (production makes total sense to me). As long as the branch is marked beta I think that is enough of a demarcation for developers. 

> For this feature specifically, the initial message backups that I make will be unencrypted, so it's really only safe to enable for localhost.

I'm less concerned about DevNetwork Beta branch data, but glad its on your mind. The XMTPLabs ran dev network can always ben wiped if needed.

🚢

PR #271, opened 2023-02-10
fix: allow backwards compatibility with ledger enableidentity signatures

michaelx11:
References: https://github.com/xmtp-labs/hq/issues/834

**Background**

XMTP uses a specific signature as secret material to protect the encrypted store. The signature is a `secp256k1` ECDSA signature, with `r,s` encoded into 64 bytes (32 each). There is a 65th byte though, which is the "recovery id". This allows verifiers to recover the public key unambiguously from the signature and the signed message.

Unfortunately, it's [documented](https://ethereum.stackexchange.com/questions/111610/how-do-ledger-hardware-wallet-signatures-differ-from-web3-eth-personal-sign) that Ledger in personal signatures uses a canonical version of the recovery id which starts at `v=0`. Whereas Metamask normally uses `v+27` as the recovery id.

**Problem**

If a user creates and encrypts their identity bundle with one format (Ledger canonical) and then attempts to use XMTP with another wallet provider, the encryption key will not be a match.

Example signatures:
```
Ledger: 192deeb77a01e8a9daaa41f4ac603b1e434121fd69a4f51317a0e8513b23799555aa94c3375a1cd8b4579d14a1b20f91c20f5385217de7c502b7803528c3c38c01

Metamask: 192deeb77a01e8a9daaa41f4ac603b1e434121fd69a4f51317a0e8513b23799555aa94c3375a1cd8b4579d14a1b20f91c20f5385217de7c502b7803528c3c38c1c
```

**Solution**

Luckily there are only two cases we need to consider, canonical and not. We can try both during decryption.

**Test Cases**
- Added a new unit test case
- If I remove the new code, the test case fails

### DIFF
--- a/src/store/EncryptedStore.ts
+++ b/src/store/EncryptedStore.ts
@@ -122,10 +122,40 @@ export default class EncryptedKeyStore implements KeyStore {
     const secret = hexToBytes(
       await wallet.signMessage(storageSigRequestText(eBundle.walletPreKey))
     )
-    const ciphertext = new Ciphertext(eBundle.ciphertext)
-    const decrypted = await decrypt(ciphertext, secret)
-    const [bundle, needsUpdate2] = getPrivateBundle(decrypted)
-    return [bundle, needsUpdate || needsUpdate2]
+
+    // Ledger uses the last byte = v=[0,1,...] but Metamask and other wallets generate with
+    // v+27 as the last byte. We need to support both for interoperability. Doing this
+    // on the decryption side provides an immediate retroactive fix.
+    // Ledger is using the canonical way, whereas Ethereum adds 27 due to some legacy stuff
+    // https://github.com/ethereum/go-ethereum/issues/19751#issuecomment-504900739
+    try {
+      // Try the original version of the signature first
+      const ciphertext = new Ciphertext(eBundle.ciphertext)
+      const decrypted = await decrypt(ciphertext, secret)
+      const [bundle, needsUpdate2] = getPrivateBundle(decrypted)
+      return [bundle, needsUpdate || needsUpdate2]
+    } catch (e) {
+      // Assert that the secret is length 65 (encoded signature + recovery byte)
+      if (secret.length !== 65) {
+        throw new Error(
+          'Expected 65 bytes before trying a different recovery byte'
+        )
+      }
+      // Try the other version of recovery byte, either +27 or -27
+      const lastByte = secret[secret.length - 1]
+      let newSecret = secret.slice(0, secret.length - 1)
+      if (lastByte < 27) {
+        // This is a canonical signature, so we need to add 27 to the recovery byte and try again
+        newSecret = new Uint8Array([...newSecret, lastByte + 27])
+      } else {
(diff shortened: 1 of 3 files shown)


src/store/EncryptedStore.ts
```diff
@@ -122,10 +122,40 @@ export default class EncryptedKeyStore implements KeyStore {
...
+      // Try the other version of recovery byte, either +27 or -27
+      const lastByte = secret[secret.length - 1]
+      let newSecret = secret.slice(0, secret.length - 1)
+      if (lastByte < 27) {
+        // This is a canonical signature, so we need to add 27 to the recovery byte and try again
+        newSecret = new Uint8Array([...newSecret, lastByte + 27])
+      } else {
+        // This canocalizes v to 0 or 1 (or maybe 2 or 3 but very unlikely)
+        newSecret = new Uint8Array([...newSecret, lastByte - 27])
+      }
```
jazzz, line 153: Is there a reason to do check for the signature format only on failure?
Can't we detect this case prior to decryption, and make the adjustment?

jazzz (commented):
This is a great find. I never knew the ledge issues were related to the V format.

src/store/EncryptedStore.ts
michaelx11: I'm not sure if we can determine which case (0,1 or 27,28) was used for the original encryption operation without trying one first

michaelx11:
Tested this with some debug lines and my throwaway ledger account. This uses Metamask direct Ledger connection (connect harware wallet option). Funnily enough, wallet connect -> Ledger Live uses the Ethereum "v" value (27 or 28) so things are totally compatible.

- Fails without the retry
Screen Shot 2023-02-13 at 11 24 17 AM

- Succeeds with retry
Screen Shot 2023-02-13 at 11 30 05 AM

elisealix22 (approved):
Tested this locally and was able to connect my wallet via Ledger Live! 🥳 So excited for this fix!

src/store/EncryptedStore.ts
jazzz: Ah I see. 

1) Can this be addressed in `toEncryptedBytes` as well standardizing our signature format would reduce complexity in the future. From a specification perspective its easier to say "EnableSignatures MUST be In X format" 
2) (OOS) but What would a migration look like for this? Is it possible to update the affected accounts in the future? so this code could be simplified?

jazzz (approved):

src/store/EncryptedStore.ts
mkobetic: If we want to make changes to how we generate the secret, then I wouldn't pick v=0 or v=27, I'd just drop that byte altogether, especially since it's not entirely random.

mkobetic (approved):
Excellent work @michaelx11! 👍

src/store/EncryptedStore.ts
michaelx11: Yup, I'm tempted to make the `toEncryptedBytes` change separately, just to avoid too many moving pieces in a single PR. Your second point is a great one, we should be able to do a semi-transparent re-encryption the next time the user signs in

PR #278, opened 2023-02-15
WIP: feat: decouple encode/decode from conversation

tg44:
My main goal is to separate the message encode/decode to workers. The problem with workers, that you can't really move functions between the main thread and the worker thread, just "serialisable" data. The problem with the current implementation that the encode/decode functions are hardly coupled with the `Conversation` for mostly no real reason.

The PR is mostly just copy-paste refactoring.

The PR is WIP because the encodeV1 test is broken. I have no idea why, I can't found any mistake, maybe a sleep will help, but until then more eyes see more.

The list of changes;
 - `Client` was used most of the times, where we could use `CodecRegistry`
 - `encodeContent` was stripped from `Client` so now it can be used with any `CodecRegistry` and without the `Client`
 - `DecodedMessage` has a `Conversation` in it, so I introduced `DecodedMessageExport` which has a `ConversationExport` in it
 - `PrivateKeyBundleV2` and `PrivateKeyBundleV1` now can be "created" with a static from, which can distinguish `PrivateKeyBundle` and `proto.PrivateKeyBundle` from an input to save a class instantiation if we can
 - `PrivateKey`, `PrivateKeyBundle`, `PublicKey`, `PublicKeyBundle`, and `Signature` got a buff, so now they can handle a `JSON.parse(JSON.serialize(uint8Array))` too
 - `Signature` and `encription` now can handle base64 stringed uint8Arrays too natively
 - both v1 and v2 conversation got a `getClient` function
 - `decodeMessageV1`, `decodeMessageV2`, `encodeMessageV1`, `encodeMessageV2` got stripped out from `ConversationV1` and `ConversationV2`, they use proto.Key-s, and Exports as input/output
 - I preserved all of the previous interfaces

### DIFF
--- a/test/conversations/Conversation.test.ts
+++ b/test/conversations/Conversation.test.ts
@@ -1,14 +1,19 @@
-import { ConversationV1 } from './../../src/conversations/Conversation'
-import { DecodedMessage, MessageV1 } from './../../src/Message'
-import { buildDirectMessageTopic } from './../../src/utils'
+import {
+  ConversationV1,
+  decodeMessageV1,
+  decodeMessageV2,
+  encodeMessageV1, encodeMessageV2
+} from "./../../src/conversations/Conversation";
+import { DecodedMessage, DecodedMessageExport, MessageV1 } from "./../../src/Message";
+import { buildDirectMessageTopic, buildUserIntroTopic } from "../../src/utils";
 import {
   Client,
   Compression,
   ContentTypeFallback,
   ContentTypeId,
-  ContentTypeText,
-} from '../../src'
-import { SortDirection } from '../../src/ApiClient'
+  ContentTypeText, PublicKeyBundle, TextCodec
+} from "../../src";
+import { SortDirection } from "../../src"
 import { sleep } from '../../src/utils'
 import { newLocalHostClient, newWallet, waitForUserContact } from '../helpers'
 import {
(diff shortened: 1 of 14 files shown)


test/conversations/Conversation.test.ts
```diff
@@ -660,4 +664,219 @@ describe('conversation', () => {
...
+      console.log(PrivateKeyBundleV1.from(keys).getPublicKeyBundle().preKey.getEthereumAddress())
+
+      const msg = await encodeMessageV1(
+        "Hi Alice!",
+        keys,
+        recipient,
+        {codecFor: (c: ContentTypeId) => baseCodecs[`${c.authorityId}/${c.typeId}`]},
+        undefined
+      )
+      console.log(bob.address)
```
tg44, line 765: This test is the broken one, here bob.address and the message sender are somehow not the same...

neekolas:
Woah. This is very cool @tg44. I remember @killthebudd4 talking about this idea last summer.

We are also working on tackling the same issue (ability to move cryptographic work out of the main thread) in a different way. You may have seen [this PR](https://github.com/xmtp/xmtp-js/pull/267) land on Monday. It's the first step in a major refactor we are doing to the codebase to separate cryptographic operations from the rest of the SDK and to have an interface between the main SDK and the Keystore that can be easily JSON serialized.

The idea here is that we are ultimately going to have many different Keystore implementations that can be used on devices with different capabilities. One of those is going to be a Service Worker Keystore, that can proxy messages to a Keystore implementation running in a service worker. 

My goal is to have v8 of `xmtp-js` ready for testing by the middle of March. The big change in v8 is that all cryptographic operations are run through the Keystore API, which will then unlock the ability to build many different Keystore implementations. 

As for this PR. If you need the performance bump immediately, feel free to use this as a fork. It looks great. But I'm not sure we will be able to merge this as-is, given how close we are to having the Keystore API ready for use. Once the Keystore API is ready to go, love to chat with you more about how we can collaborate on a Service Worker Keystore implementation.

neekolas:
I've shared some more details about the Keystore API here https://github.com/orgs/xmtp/discussions/25

tg44:
To be fair, this is not for service workers, but for worker threads... With a service worker we can instantiate the whole Client in the worker. What I want to do is to scale the Client implementation to multiple threads.

Currently I have some bugs with V1 encode, can you link me something, so I can understand why V1 and V2 is a separate thing and how much effort should I do to still support V1?

As I think in the deepcopy phase I ruin one of my legacy bundle, so my sender is not the valid sender, is there any good method to find out why the sender goes wrong?

rygine:
closing this PR due to inactivity. please reach out if you have any questions.

PR #282, opened 2023-02-21
Use Keystore For V1 Convos

neekolas:
## Summary

- Uses the Keystore for V1 conversation encryption/decryption
- This should be a non-breaking change, as we are using the Keystore and `client.keys` in parallel. Only V1 message encoding/decoding uses the new Keystore, while other flows are unchanged.

## Notes
- I have deferred the work of bootstrapping the Keystore until later in the project. This allows the Keystore and `client.keys` to exist in parallel, which allows us to merge this as a non-breaking change. Later, we will have to refactor `Client.create(...)` to support the creation of a Keystore based on the `ClientCreateOptions`. 
- A similar non-breaking path for V2 messages is going to be tricky, since the `TopicKeys` need to move from the SDK to the Keystore. Would be more complicated to have them both in sync, and would have a meaningful performance penalty since we would have to decrypt invites twice. Think that one will have to go to the beta branch only


jazzz (approved):
Splitting V1 and V2 makes a lot of sense to me.

src/conversations/Conversation.ts
```diff
@@ -69,38 +78,154 @@ export class ConversationV1 {
...
+    )
   }
 
   messagesPaginated(
     opts?: ListMessagesPaginatedOptions
   ): AsyncGenerator<DecodedMessage[]> {
     return this.client.listEnvelopesPaginated(
       [this.topic],
+      // This won't be performant once we start supporting a remote keystore
+      // TODO: Either better batch support or we ditch this under-utilized feature
```
michaelx11, line 194: great comment

michaelx11 (approved):
lgtm

PR #283, opened 2023-02-23
Keystore persistence

neekolas:
## Summary

- Start implementing the basic persistence layer for the Keystore. Wanted to get the building block pieces in before I start work on the encryption layer.
- Includes a LocalStorage polyfill that doesn't rely on the filesystem or pollute globals, to avoid past pains with off-the-shelf polyfills.

## Next steps
- Implement encrypted persistence
- Implement conversation storage using encrypted persistence class

## Notes
- A lot of this code is very similar to what lives in `src/store`. The store folder is ultimately going to be deleted once the keystore bootstrapping part of V8 comes in. Some of the code will make its way into the keystore, which will resolve some of the naming confusion (`store` has a class named `KeyStore` which will be integrated into the new Keystore and renamed to something less confusing)

### DIFF
--- a/src/keystore/persistence/LocalStoragePonyfill.ts
+++ b/src/keystore/persistence/LocalStoragePonyfill.ts
@@ -15,12 +15,6 @@ export default class LocalStoragePonyfill implements Storage {
     this.store = new Map()
   }
 
-  private validateString(key: string): void {
-    if (!(typeof key === 'string')) {
-      throw new TypeError('Key must be a string')
-    }
-  }
-
   getItem(key: string): string | null {
     this.validateString(key)
 
@@ -58,5 +52,11 @@ export default class LocalStoragePonyfill implements Storage {
     this.validateString(key)
     this.validateString(value)
     this.store.set(String(key), String(value))
+  }
+
+  private validateString(key: string): void {
+    if (!(typeof key === 'string')) {
+      throw new TypeError('Key must be a string')
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


src/keystore/persistence/LocalStoragePonyfill.ts
```diff
@@ -0,0 +1,62 @@
+// Fully in-memory polyfill for the browser storage API.
+// Borrowed from https://github.com/MitchellCash/node-storage-polyfill but implemented as a ponyfill instead of a polyfill
+
+export default class LocalStoragePonyfill implements Storage {
```
michaelx11, line 4: TIL: ponyfill is a thing

src/keystore/persistence/LocalStoragePonyfill.ts
```diff
@@ -0,0 +1,62 @@
...
+  store: Map<string, string>
+  constructor() {
+    this.store = new Map()
+  }
+
+  get length(): number {
+    return this.store.size
```
michaelx11, line 11: I have the worst suggestion ever: Would it be worth tracking a `total_bytes` parameter and having a `byte_size` method? Might be useful for metrics and stuff but probably not worth the bookkeeping right now.

michaelx11 (approved):
lgtm

src/keystore/persistence/LocalStoragePonyfill.ts
neekolas: So, this ponyfill is meant to be a 1:1 match of the browser `Storage` API.

If we wanted to track `total_bytes` I'd be tempted to do that bookkeeping on top of the real `LocalStorage`, which does have some hard limits (5mb total).

jazzz (approved):
🚀

src/keystore/persistence/LocalStoragePonyfill.ts
michaelx11: Oh that makes a lot more sense

PR #293, opened 2023-03-01
Encrypted persistence

neekolas:
## Summary

- Adds an ecrypted persistence layer using ECIES for single party encryption
- Uses [`bitchan/eccrypto`](https://github.com/bitchan/eccrypto) as recommended by [Cryptobook](https://cryptobook.nakov.com/asymmetric-key-ciphers/exercises-ecies-encrypt-decrypt) and @michaelx11

## Notes
- I don't love `bitchan/eccrypto` . It's also been basically untouched in 3+ years. It uses a different implementation of secp256k1 than we do, which just adds some bloat to our bundle. All the code is one file. The good news is that the library is quite small, and mostly just calls out to the underlying crypto libraries.
- This PR is not opinionated on where the key comes from, but I assume I want to use the `IdentityKey` here, given that we don't have metadata to describe which `preKey` to use for decryption and `preKeys` could theoretically change.
- Targeting this PR at `nmolnar/v2-conversation-persistence` until https://github.com/xmtp/xmtp-js/pull/292 merges. Then will target at `main`
- Still need to do some browser testing before I merge to make sure that `eccrypto` works correctly with webpack and uses the native browser crypto libs.

### DIFF
--- a/src/keystore/persistence/EncryptedPersistence.ts
+++ b/src/keystore/persistence/EncryptedPersistence.ts
@@ -4,6 +4,7 @@ import eccrypto, { Ecies } from 'eccrypto'
 const IV_LENGTH = 16
 const EPHEMERAL_PUBLIC_KEY_LENGTH = 65
 const MAC_LENGTH = 32
+const CIPHERTEXT_BLOCK_SIZE_BYTES = 16
 
 const assertEciesLengths = (ecies: Ecies): void => {
   if (ecies.iv.length !== IV_LENGTH) {
@@ -12,7 +13,10 @@ const assertEciesLengths = (ecies: Ecies): void => {
   if (ecies.ephemPublicKey.length !== EPHEMERAL_PUBLIC_KEY_LENGTH) {
     throw new Error('Invalid ephemPublicKey length')
   }
-  if (ecies.ciphertext.length < 1 || ecies.ciphertext.length % 16 !== 0) {
+  if (
+    ecies.ciphertext.length < 1 ||
+    ecies.ciphertext.length % CIPHERTEXT_BLOCK_SIZE_BYTES !== 0
+  ) {
     throw new Error('Invalid ciphertext length')
   }
   if (ecies.mac.length !== MAC_LENGTH) {
@@ -34,7 +38,13 @@ const serializeEcies = (ecies: Ecies): Uint8Array => {
 }
 
 const deserializeEcies = (data: Uint8Array): Ecies => {
-  if (data.length < IV_LENGTH + EPHEMERAL_PUBLIC_KEY_LENGTH + MAC_LENGTH) {
+  if (
+    data.length <
+    IV_LENGTH +
+      EPHEMERAL_PUBLIC_KEY_LENGTH +
+      MAC_LENGTH +
+      CIPHERTEXT_BLOCK_SIZE_BYTES
+  ) {
     throw new Error('Invalid data length')
   }
   const iv = data.slice(0, IV_LENGTH)
--- a/src/keystore/persistence/EncryptedPersistence.ts
+++ b/src/keystore/persistence/EncryptedPersistence.ts
@@ -66,6 +66,8 @@ const deserializeEcies = (data: Uint8Array): Ecies => {
 
 /**
  * EncryptedPersistence is a Persistence implementation that uses ECIES to encrypt all values
+ * ECIES encryption protects against unauthorized reads, but not unauthorized writes.
+ * A third party with access to the underlying store could write malicious data using the public key of the owner
  */
 export default class EncryptedPersistence implements Persistence {
   private persistence: Persistence
--- a/src/keystore/persistence/EncryptedPersistence.ts
+++ b/src/keystore/persistence/EncryptedPersistence.ts
@@ -1,0 +1,82 @@
+import { Persistence } from './interface'
+import eccrypto, { Ecies } from 'eccrypto'
+
+const assertEciesLengths = (ecies: Ecies): void => {
+  if (ecies.iv.length !== 16) {
+    throw new Error('Invalid iv length')
+  }
+  if (ecies.ephemPublicKey.length !== 65) {
+    throw new Error('Invalid ephemPublicKey length')
+  }
+  if (ecies.ciphertext.length < 1) {
+    throw new Error('Invalid ciphertext length')
+  }
+  if (ecies.mac.length !== 32) {
+    throw new Error('Invalid mac length')
+  }
+}
+
+const serializeEcies = (ecies: Ecies): Uint8Array => {
+  assertEciesLengths(ecies)
+  const { iv, ephemPublicKey, ciphertext, mac } = ecies
+  const result = new Uint8Array(
+    iv.length + ephemPublicKey.length + ciphertext.length + mac.length
+  )
+  result.set(iv, 0)
+  result.set(ephemPublicKey, iv.length)
+  result.set(ciphertext, iv.length + ephemPublicKey.length)
+  result.set(mac, iv.length + ephemPublicKey.length + ciphertext.length)
+  return result
+}
+
+const deserializeEcies = (data: Uint8Array): Ecies => {
+  const iv = data.slice(0, 16)
+  const ephemPublicKey = data.slice(16, 81)
+  const ciphertext = data.slice(81, data.length - 32)
+  const mac = data.slice(data.length - 32, data.length)
+  return {
+    iv: Buffer.from(iv),
+    ephemPublicKey: Buffer.from(ephemPublicKey),
+    ciphertext: Buffer.from(ciphertext),
+    mac: Buffer.from(mac),
+  }
+}
+
+/**
+ * EncryptedPersistence is a Persistence implementation that uses ECIES to encrypt all values
+ */
+export default class EncryptedPersistence implements Persistence {
+  private persistence: Persistence
+  private privateKey: Buffer
+  private publicKey: Buffer
+
+  constructor(persistence: Persistence, privateKey: Uint8Array) {
+    this.persistence = persistence
+    this.privateKey = Buffer.from(privateKey)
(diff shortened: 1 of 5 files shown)


src/keystore/persistence/EncryptedPersistence.ts
```diff
@@ -0,0 +1,82 @@
...
+    iv.length + ephemPublicKey.length + ciphertext.length + mac.length
+  )
+  result.set(iv, 0)
+  result.set(ephemPublicKey, iv.length)
+  result.set(ciphertext, iv.length + ephemPublicKey.length)
+  result.set(mac, iv.length + ephemPublicKey.length + ciphertext.length)
+  return result
+}
+
+const deserializeEcies = (data: Uint8Array): Ecies => {
```
jazzz, line 32: This looks dangerous. 
the length of data is not verified prior to usage. Input bytes should be checked to ensure they are a sane length.

src/keystore/persistence/EncryptedPersistence.ts
```diff
@@ -0,0 +1,82 @@
...
+import eccrypto, { Ecies } from 'eccrypto'
+
+const assertEciesLengths = (ecies: Ecies): void => {
+  if (ecies.iv.length !== 16) {
+    throw new Error('Invalid iv length')
+  }
+  if (ecies.ephemPublicKey.length !== 65) {
+    throw new Error('Invalid ephemPublicKey length')
+  }
+  if (ecies.ciphertext.length < 1) {
```
jazzz, line 11: AES produces blocks of 128bits -- I would expect `ciphertext.length % 16 ==0`

src/keystore/persistence/EncryptedPersistence.ts
```diff
@@ -0,0 +1,82 @@
...
+  )
+  result.set(iv, 0)
+  result.set(ephemPublicKey, iv.length)
+  result.set(ciphertext, iv.length + ephemPublicKey.length)
+  result.set(mac, iv.length + ephemPublicKey.length + ciphertext.length)
+  return result
+}
+
+const deserializeEcies = (data: Uint8Array): Ecies => {
+  const iv = data.slice(0, 16)
```
jazzz, line 33: From a crypto auditability/ readability POV the slice indexing here could be improved. Extracting the slice indexes into variables which are used for both Serialize and Deserialize would add symmetry and make mistakes easier to detect in future PRs

src/keystore/persistence/EncryptedPersistence.ts
neekolas: I can add a check for that

src/keystore/persistence/EncryptedPersistence.ts
neekolas: What would a sane length be? > 132?

src/keystore/persistence/EncryptedPersistence.ts
neekolas: Sounds reasonable

neekolas: Done

src/keystore/persistence/EncryptedPersistence.ts
jazzz: For the serialization above I get:

```
fixed_length = IV_LENGTH + EPHEMERAL_PUBLIC_KEY_LENGTH + MAC_LENGTH + AES_BLOCKSIZE //(minimum 1 block of ciphertext) = 129

// Ensure data is atleast 'fixed_length' bytes and the remaining bytes are valid AES_blocks
(data.length > fixed_length) && ((data.length - fixed_length) % AES_BLOCKSIZE) == 0
```

neekolas: Updated

jazzz (approved):
🟢

src/keystore/persistence/EncryptedPersistence.ts
```diff
@@ -0,0 +1,104 @@
...
+    return null
+  }
+
+  async setItem(key: string, value: Uint8Array): Promise<void> {
+    const encrypted = await this.encrypt(value)
+    await this.persistence.setItem(key, encrypted)
+  }
+
+  private async encrypt(value: Uint8Array): Promise<Uint8Array> {
+    const ecies = await eccrypto.encrypt(this.publicKey, Buffer.from(value))
```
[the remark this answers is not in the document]
neekolas: `LocalStorage` persistence (and the potential future `FilesystemPersistence`) both have the possibility of malicious entities deleting the store or inserting bad data. For example, another script on the page can write anything it wants to `LocalStorage`.

I'm not sure I understand what problem authentication would solve. If the malicious entity encrypted their data with another key or wrote garbage to the storage, decryption would fail and the process would bail out. That feels like the correct behaviour.

src/keystore/persistence/EncryptedPersistence.ts
```diff
@@ -0,0 +1,104 @@
...
+const serializeEcies = (ecies: Ecies): Uint8Array => {
+  assertEciesLengths(ecies)
+  const { iv, ephemPublicKey, ciphertext, mac } = ecies
+  const result = new Uint8Array(
+    iv.length + ephemPublicKey.length + ciphertext.length + mac.length
+  )
+  result.set(iv, 0)
+  result.set(ephemPublicKey, IV_LENGTH)
+  result.set(ciphertext, IV_LENGTH + EPHEMERAL_PUBLIC_KEY_LENGTH)
+  result.set(mac, IV_LENGTH + EPHEMERAL_PUBLIC_KEY_LENGTH + ciphertext.length)
```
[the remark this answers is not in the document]
neekolas: I wasn't able to find a nice existing library. We _could_ do the serialization by creating a Protobuf message for ECIES data, but given all the fields but one are of a fixed length it feels unnecessary. We'd still need to validate the lengths of all the fields and otherwise validate the data.

test/keystore/persistence/EncryptedPersistence.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    await encryptedPersistence.setItem(TEST_KEY, data)
+    await encryptedPersistence.setItem(TEST_KEY_2, data)
+
+    const [rawResult1, rawResult2] = await Promise.all([
+      persistence.getItem(TEST_KEY),
+      persistence.getItem(TEST_KEY_2),
+    ])
+    expect(rawResult1).not.toEqual(rawResult2)
+  })
```
michaelx11, line 69: Can we add some negative test cases here where it catches malformed inputs? I'm happy to contribute those as a commit.

michaelx11 (commented):
I apologize for the eccrypto library choice. Also not my ideal pick but preferred it to https://github.com/ecies/js which is single-maintainer, and also has not had a code update in ~3 years.

test/keystore/persistence/EncryptedPersistence.test.ts
michaelx11: Ah nvm, it is hard to test here since encryption happens at a layer below

neekolas:
> I apologize for the eccrypto library choice

Nothing to apologize for. It's the best amongst a bunch of mediocre options, far as I can tell.

michaelx11 (approved):
ship it, also feel free to touch up the test cases I added if style is wonky

src/keystore/persistence/EncryptedPersistence.ts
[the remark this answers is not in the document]
neekolas: You're totally right. Someone could use a valid private key and encrypt a message for the user's public key that is decryptable. 

I still don't see where the attack is. Making up topics and putting them into the conversation store feels more like mischief than an attack, with the way the Keystore works.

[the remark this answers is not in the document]
neekolas: At least for the current use-case I care about unauthenticated reads and care less about unauthenticated writes. But I can't speak to other use-cases for the persistence. 

Do we have a clear idea of how we would protect against both if we wanted to future proof the solution?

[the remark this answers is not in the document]
neekolas: That wouldn't be too hard. Given that this code isn't used yet, it's safe and easy for me to merge and then continue to improve the implementation.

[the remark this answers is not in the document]
jazzz: From a threat perspective there are a few cases to consider

Fake Entry generation: An attacker generates malicious entries. 
- Would be immediately detectable since the attacker does not have access to the the other valid entries
- Harm is injecting a topic/topic key into the inbox. Since topic keys are not updated, and attacker could DOS a conversation by blocking the correct key form being entered into the store. 

Persistence Entry Replay: An attacker deletes the current entry with an previous blob. 
- End result is a few missing entries which I imagine would be re-added once a KeyStore recieves a message it doesn't have a key for 

Persistence Entry Deletion: Attacker wipes localstorage
- Keys are rebuild
- No harm

Adding a signature would help with 1 but the placement and replay protection is something we'd want to consider before implementing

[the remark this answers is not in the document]
neekolas: OK. I'm going to merge as-is and make a ticket to add a signature. Will get that in before v8 is released.

Looking at this some more, I think there is an attack vector where an attacker could forge an invite to a conversation purporting to be with user X, but with weakened encryption that the attacker can read. They would then be able to read one side of the conversation and write to it.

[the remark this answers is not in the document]
jazzz: Sounds good. 

> and write to it
I'd be be interested to see the analysis on this. 
Messages are signed by the identity key of the sender, so an attacker would need access to the Identity key if they are attempting to spoof a message. 

If a spoofed message can be injected into an inbox then that should be addressed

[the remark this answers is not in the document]
neekolas: Ya. You're right. They could write to the topic, but the signatures would fail to verify and the message would be discarded. Could read though.

PR #298, opened 2023-03-07
fix: verify prekey signed by identity key in decodeMessage

michaelx11:
**Issue**

Messages are signed by the users PreKey, however in all SDK implementations the Prekey is not verified to be signed by the IdentityKey. This means a crafted bundle using Amal's IdentityKey, and Mallory's PreKey resulting in a message signature being considered valid. As the sender's address is derived from the wallet signature on the identity key, Mallory's message appears to be sent from Amal.

**Changes**

- Add test case to catch this issue (uses an exported conversation and a hand-crafted malicious message to reproduce)
- Add check in `decodeMessage` for conversation V2

**Tests**

- Unit test
- Verified in SDK integration

### DIFF
--- a/src/conversations/Conversation.ts
+++ b/src/conversations/Conversation.ts
@@ -544,17 +544,14 @@ export class ConversationV2 {
     if (!senderPreKey || !senderPreKey.signature || !senderPreKey.keyBytes) {
       throw new Error('missing pre-key or pre-key signature')
     }
-    const keyDigest = await sha256(senderPreKey.keyBytes)
     const senderIdentityKey = signed.sender?.identityKey
     if (!senderIdentityKey) {
       throw new Error('missing identity key in bundle')
     }
-    if (
-      !new SignedPublicKey(senderIdentityKey).verify(
-        new Signature(senderPreKey.signature),
-        keyDigest
-      )
-    ) {
+    const isValidPrekey = await new SignedPublicKey(
+      senderIdentityKey
+    ).verifyKey(new SignedPublicKey(senderPreKey))
+    if (!isValidPrekey) {
       throw new Error('pre key not signed by identity key')
     }
 
--- a/src/conversations/Conversation.ts
+++ b/src/conversations/Conversation.ts
@@ -537,6 +537,27 @@ export class ConversationV2 {
       throw new Error('incomplete signed content')
     }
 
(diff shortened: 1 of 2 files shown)


jazzz (approved):

michaelx11:
Tried to dynamically generate the test vector, but would need to rework some of our SDK visibility or get a lot more clever. I'll leave as is for now and we can clean up the test later. @neekolas points out that right off the bat, conversation exports may be going away.

src/conversations/Conversation.ts
```diff
@@ -537,6 +537,27 @@ export class ConversationV2 {
...
+    if (!senderPreKey || !senderPreKey.signature || !senderPreKey.keyBytes) {
+      throw new Error('missing pre-key or pre-key signature')
+    }
+    const keyDigest = await sha256(senderPreKey.keyBytes)
+    const senderIdentityKey = signed.sender?.identityKey
+    if (!senderIdentityKey) {
+      throw new Error('missing identity key in bundle')
+    }
+    if (
+      !new SignedPublicKey(senderIdentityKey).verify(
```
mkobetic, line 553: We used to have `verifyKey` function.

michaelx11: Great call

mkobetic:
> I wonder if it makes sense to pull this logic and the logic in `MessageV2.create()` into one unified method called `verify/deriveMessageSender()` or something to that effect, and call it either here or in `create()`. But don't have a strong opinion here

I was thinking something along those lines, how about PublicKeyBundle.validate() or something and call it from both the `sharedSecret` function and here.

michaelx11:
> > I wonder if it makes sense to pull this logic and the logic in `MessageV2.create()` into one unified method called `verify/deriveMessageSender()` or something to that effect, and call it either here or in `create()`. But don't have a strong opinion here
> 
> I was thinking something along those lines, how about PublicKeyBundle.validate() or something and call it from both the `sharedSecret` function and here.

I'm in agreement for all of these! I think for now we can keep the patch minimal and correct, then do a larger refactor immediately after

neekolas:
You might want to do the bigger refactor against the `beta` branch, given how much of this code is restructured in v8 to accomodate the Keystore batch calls.

PR #299, opened 2023-03-09
Keystore provider

neekolas:
## Summary
- Creates the infrastructure for loading keys from the network and creating a Keystore
- Tests for those flows, including legacy keys and ledger wallets
- Adds new `TopicPersistence` class for persisting data on the network
- Adds `StaticKeystoreProvider` for cases where the user provides a `privateKeyOverride`
- Adds a `KeyGeneratorKeystoreProvider`, which is a truly awful name that I am open to suggestions on, for creating a Keystore from a newly generated key.

## Next up
- Actually bootstrap the Keystore as part of `Client.create`
- Delete `store` folder and remove all duplicate code
- Implement `Client.getKeys`, but with the possibility that some keystore implementations will throw an error (for example, remote Keystores)
- Implement a mechanism for getting API auth tokens from the Keystore

### DIFF
--- a/src/keystore/persistence/TopicPersistence.ts
+++ b/src/keystore/persistence/TopicPersistence.ts
@@ -1,0 +1,52 @@
+import { messageApi } from '@xmtp/proto'
+import ApiClient from '../../ApiClient'
+import Authenticator from '../../authn/Authenticator'
+import { b64Decode } from '../../utils/bytes'
+import { buildUserPrivateStoreTopic } from '../../utils/topic'
+import { Persistence } from './interface'
+
+export default class TopicPersistence implements Persistence {
+  apiClient: ApiClient
+  constructor(apiClient: ApiClient) {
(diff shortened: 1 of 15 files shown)


src/keystore/providers/NetworkKeyManager.ts
```diff
@@ -0,0 +1,207 @@
+import { utils } from 'ethers'
```
neekolas, line 1: This is borrowed from `store/EncryptedStore`. That will be deleted in the next PR

src/keystore/persistence/TopicPersistence.ts
```diff
@@ -0,0 +1,52 @@
+import { messageApi } from '@xmtp/proto'
```
neekolas, line 1: This is borrowed from `store/PrivateTopicStore.ts`

michaelx11 (approved):
lgtm!

PR #301, opened 2023-03-16
Keystore Bootstrapping

neekolas:
## Summary

- Bootstraps the Keystore as part of client setup
- Removes `client.keys` and `client.legacyKeys`.


src/Client.ts
```diff
@@ -263,27 +258,27 @@ export default class Client {
...
-      this.keys.getPublicKeyBundle().equals(bundle)
+      this.signedPublicKeyBundle.equals(bundle)
     ) {
       return
     }
     // TEMPORARY: publish V1 contact to make sure there is one in the topic
     // in order to preserve compatibility with pre-v7 clients.
     // Remove when pre-v7 clients are deprecated
-    this.publishUserContact(true)
+    await this.publishUserContact(true)
```
neekolas, line 268: I caught this bug along the way. We should await the publishing of the contacts to ensure that no races can change the ordering of the bundles. If the legacy bundle is the newest one, other clients may attempt to create V1 conversations instead of V2.

test/crypto/PrivateKeyBundle.test.ts
```diff
@@ -32,42 +28,6 @@ describe('Crypto', function () {
           .equals((bundle2 as PrivateKeyBundleV2).getPublicKeyBundle())
       )
     })
-    it('encrypts private key bundle for storage using a wallet', async function () {
```
neekolas, line 35: This gets covered as part of the `keystore/providers` test suite

src/Client.ts
michaelx11: nice find! Do we normally get warnings emitted for unchecked Promise or something similar? Wondering if there are any more bugs like this lurking around.

src/Client.ts
```diff
@@ -562,76 +557,6 @@ export default class Client {
       mapper
     )
   }
-
-  async signBytes(bytes: Uint8Array): Promise<Signature> {
```
michaelx11, line 566: I see, this is why the Keystore `sign` function was so generic. The goal was to push this already super generic function into the Keystore API?

src/Client.ts
```diff
@@ -703,3 +628,33 @@ async function getUserContactsFromNetwork(
...
 }
+
+export function defaultKeystoreProviders(): KeystoreProvider[] {
+  return [
+    // First check to see if a `privateKeyOverride` is provided and use that
+    new StaticKeystoreProvider(),
+    // Next check to see if a EncryptedPrivateKeyBundle exists on the network for the wallet
+    new NetworkKeystoreProvider(),
+    // If the first two failed with `KeystoreProviderUnavailableError`, then generate a new key and write it to the network
+    new KeyGeneratorKeystoreProvider(),
```
michaelx11, line 639: I really this pattern but does it mean that transient errors with NetworkKeystoreProvider could cause a new identity key + pre key bundle to be created?

michaelx11: Oh I guess the code below contains a partial answer: the bootstrap process will only continue if a KeystoreProviderUnavailableError is thrown. That shouldn't happen if it's a transient 404 on a topic right?

src/conversations/Conversations.ts
```diff
@@ -384,16 +381,11 @@ export default class Conversations {
...
+    const topic = buildUserIntroTopic(this.client.address)
     const messages = await this.client.listEnvelopes(
-      [buildUserIntroTopic(this.client.address)],
-      async (env) => {
-        const msg = await MessageV1.fromBytes(
-          b64Decode(env.message as unknown as string)
-        )
-
-        // Decrypt the message to ensure it is valid. Ignore the contents
-        await msg.decrypt(this.client.legacyKeys)
```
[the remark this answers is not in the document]
michaelx11: I see, if we do a decrypt here it'll happen for all envelopes regardless of whether there was a previous success. Nice catch!

src/utils/keystore.ts
```diff
@@ -0,0 +1,48 @@
...
+  messages: MessageV1[],
+  myPublicKeyBundle: PublicKeyBundle
+): keystore.DecryptV1Request => {
+  return {
+    requests: messages.map((m: MessageV1) => {
+      const sender = new PublicKeyBundle({
+        identityKey: m.header.sender?.identityKey,
+        preKey: m.header.sender?.preKey,
+      })
+
```
michaelx11, line 32: Gut checking: we don't need to check bundle validity because it happens when we[ derive the shared secret](https://github.com/xmtp/xmtp-js/blob/0957848/src/crypto/PrivateKeyBundle.ts#L85) right? I'm paranoid about checking bundles now

src/Client.ts
```diff
@@ -211,15 +201,19 @@ export default class Client {
   ): Promise<Client> {
     const options = defaultOptions(opts)
     const apiClient = createApiClientFromOptions(options)
-    const keys = await loadOrCreateKeysFromOptions(options, wallet, apiClient)
-    // TODO: Properly bootstrap the keystore and replace `loadOrCreateKeysFromOptions`
-    const keystore = await InMemoryKeystore.create(keys)
+    const keystore = await bootstrapKeystore(options, apiClient, wallet)
+    const publicKeyBundle = new PublicKeyBundle(
+      await keystore.getPublicKeyBundle()
```
michaelx11, line 206: Is the idea here that the public keybundle never changes and calling into Keystore each time is 1) a waste and 2) an async operation

test/Message.test.ts
```diff
@@ -19,23 +20,33 @@ describe('Message', function () {
...
     assert.equal(msg1.recipientAddress, bobWalletAddress)
-    const decrypted = await msg1.decrypt(alice)
+    const decrypted = await msg1.decrypt(
+      aliceKeystore,
+      alice.getPublicKeyBundle()
+    )
     assert.deepEqual(decrypted, content)
 
-    // Bob decodes message from Alice
+    // // Bob decodes message from Alice
```
michaelx11, line 44: super nit double comment

test/Message.test.ts
```diff
@@ -45,24 +56,37 @@ describe('Message', function () {
...
+      alice.getPublicKeyBundle(),
       bob.getPublicKeyBundle(),
-      new TextEncoder().encode('hi'),
       new Date()
     )
     assert.ok(!msg.error)
-    expect(msg.decrypt(eve)).rejects.toThrow(NoMatchingPreKeyError)
+    const eveResult = msg.decrypt(eveKeystore, eve.getPublicKeyBundle())
+    expect(eveResult).rejects.toThrow(KeystoreError)
+    // expect(eveResult).rejects.toThrow(
```
michaelx11, line 71: nit: commented dead code

test/conversations/Conversation.test.ts
```diff
@@ -309,20 +311,19 @@ describe('conversation', () => {
...
-      alice.legacyKeys = aliceKeys
+      alice.keystore = aliceKeystore
       await aliceConvo.send('Hello from Alice')
       const result = await stream.next()
       const msg = result.value as DecodedMessage
       expect(msg.senderAddress).toBe(alice.address)
       expect(msg.content).toBe('Hello from Alice')
       await stream.return()
-      expect(consoleWarn).toBeCalledTimes(1)
+      expect(consoleWarn).toBeCalledTimes(2)
```
michaelx11, line 326: dumb question: where does the second warn come from?

michaelx11 (approved):
Overall LGTM - great changes!

src/Client.ts
neekolas: Great question. My goal was to replicate the behaviour today.

If the network has an error and does not return a valid response, that error will bubble and the bootstrapping process will abort.

But if the network returns a valid empty response we will create a new bundle and publish it.

michaelx11: Awesome, thanks!

src/Client.ts
neekolas: That's right. If we introduce changes where the Client might mutate the publicKeyBundle, we will need to be mindful of that and update this code to either not store it in the client or refresh on mutation.

src/Client.ts
neekolas: Actually, this code is totally unused.

The reason why we need the generic sign function in the Keystore is [here](https://github.com/xmtp/xmtp-js/blob/main/src/conversations/Conversation.ts#L493). For V2 messages, the protocol expects the signature to be against the digest of a concatenation of the `payload` and `headerBytes`. Shipping both input values over the wire to the Keystore feels unnecessary, and adding any additional context or metadata would not be backwards compatible.

test/conversations/Conversation.test.ts
neekolas: Was a change to the sequence of where we log decryption errors. But looking back at it, we probably don't need to log twice in this case. Moved the console.log a few lines down and it's back to 1.

src/utils/keystore.ts
neekolas: That's right

src/Client.ts
neekolas: This does seem like something we should be able to catch with a linter. I'll look into that.

src/conversations/Conversations.ts
```diff
@@ -229,11 +229,11 @@ export default class Conversations {
       if (env.contentTopic === introTopic) {
         const messageBytes = b64Decode(env.message as unknown as string)
         const msg = await MessageV1.fromBytes(messageBytes)
-        await msg.decrypt(this.client.legacyKeys)
         const peerAddress = this.getPeerAddress(msg)
         if (!newPeer(peerAddress)) {
           return undefined
         }
+        await msg.decrypt(this.client.keystore, this.client.publicKeyBundle)
```
jazzz, line 236: Having to pass in the publicBundle here seems like its exposing complexity with minimal gain.

Today there is only a single PreKey, so the client gets the PublicBundle from the keyStore then it passes it back in.

In the future, there are multiple preKeys and the client will need to choose the correct matching bundle for the preKey which was used by the sender. Since this may not be the most current, the keyStore will need to provide access to multiple preKeys and then the client will need to choose which one to pass in.

In both these cases it seems that the KeyStore could perform all this work behind the scenes, and decrypt messages which it has keys for.

jazzz (approved):
No issues with this PR, few questions around the larger keyStore API which are arising now that I'm seeing it in action.

src/conversations/Conversations.ts
neekolas: The Keystore doesn't actually get the `PublicBundle` of the user. It just gets a boolean `isSender` flag, and the keys of the peer. This `decrypt` method just needs the client's bundle to know whether the user is the sender or the recipient to craft the right request to the Keystore.

neekolas:
```proto
    // A single decryption request
    message Request {
        xmtp.message_contents.Ciphertext payload = 1;
        xmtp.message_contents.PublicKeyBundle peer_keys = 2;
        bytes header_bytes = 3;
        bool is_sender = 4;
    }
```

jazzz: Ah yes, traced it incorrectly 👍

src/Client.ts
michaelx11: Ah okay that makes sense

PR #305, opened 2023-03-20
Support for receiving send key invites

richardhuaaa:
This change is required for forward compatibility for send keys. The only change to production behavior is in `InMemoryKeystore`, where we now accept V2 invites in addition to existing V1 invites. Add `?w=1` to the URL to ignore whitespace changes.

Apologies for the size of this PR - most of it is boilerplate. As discussed, where possible I've copy-pasted/followed existing patterns rather than refactoring, seeing as we will be re-implementing most of this in rust next quarter.

I have another PR coming up shortly with unit tests - I had to write part of the invite creation flow in order to write the tests, and I didn't want to make this PR even larger.


src/Invitation.ts
```diff
@@ -67,6 +74,148 @@ export class InvitationV1 implements invitation.InvitationV1 {
...
+    if (peerHeader && selfHeader) {
+      throw new Error(
+        'SealedInvitationHeaderV2 cannot have both peer and self header'
+      )
+    } else if (!peerHeader && !selfHeader) {
+      throw new Error('SealedInvitationHeaderV2 missing peer or self header')
+    }
+
+    this.peerHeader = peerHeader
+    this.selfHeader = selfHeader
```
michaelx11, line 103: nit: would it be easier to do the conditional once e.g. `const usedHeader = peerHeader || selfHeader` and then do away with all the `||` below?

src/crypto/PublicKey.ts
```diff
@@ -211,6 +217,123 @@ export class SignedPublicKey
...
+{
+  keyBytes: Uint8Array
+  staticSignature: AccountLinkedStaticSignature | undefined
+  siweSignature: signature.AccountLinkedSIWESignature | undefined
+
+  constructor(obj: publicKey.AccountLinkedPublicKey_V1) {
+    if (!obj.keyBytes) {
+      throw new Error('Invalid AccountLinkedPublicKeyV1 has no keyBytes')
+    }
+    super(publicKey.UnsignedPublicKey.decode(obj.keyBytes))
```
michaelx11, line 232: TIL super doesn't have to be the first statement in a constructor (apparently fixed by this guy [PR](https://blog.joshuakgoldberg.com/code-before-super/))

src/crypto/PublicKey.ts
```diff
@@ -211,6 +217,123 @@ export class SignedPublicKey
...
+        signatureText !==
+        WalletSigner.accountLinkRequestText(this.bytesToSign(), role)
+      ) {
+        throw new Error('Signature text does not match expected text')
+      }
+      const digest = hexToBytes(utils.hashMessage(signatureText))
+      const signature = this.staticSignature.walletEcdsaCompact
+      const publicKey = ecdsaSignerKey(digest, signature)
+      if (!publicKey) {
+        throw new Error('Static signature is not valid for given role')
```
michaelx11, line 264: I'm not sure if this error would get thrown for an invalid signature, I think it must just return an unexpected ethereum address

src/crypto/PublicKey.ts
```diff
@@ -211,6 +217,123 @@ export class SignedPublicKey
...
+    )
+    const key = new AccountLinkedPublicKey({
+      v1: new AccountLinkedPublicKeyV1({
+        keyBytes: legacyKey.keyBytes,
+        staticSignature,
+        siweSignature: undefined,
+      }),
+    })
+    if (!key.getLinkedAddress(role)) {
+      throw new Error('Legacy key has invalid signature for given role')
```
michaelx11, line 331: Same comment as above, I could be wrong but I think it'll return some random address rather than throw or return null

src/crypto/PrivateKey.ts
```diff
@@ -172,6 +178,71 @@ export class SignedPrivateKey
...
+    if (!obj.secp256k1) {
+      throw new Error('invalid private key')
+    }
+    secp256k1Check(obj.secp256k1)
+    this.secp256k1 = obj.secp256k1
+    this.createdNs = obj.createdNs
+    if (!obj.publicKey) {
+      throw new Error('missing public key')
+    }
+    this.publicKey = new AccountLinkedPublicKey(obj.publicKey)
```
michaelx11, line 198: Does it make sense to regenerate the `publicKey` from the private key bytes and check?

src/crypto/PrivateKeyBundle.ts
```diff
@@ -1,11 +1,124 @@
...
+    if (!(await peer.accountLinkedKey.verifyKey(peer.preKey))) {
+      throw new Error('peer preKey signature invalid')
+    }
+    if (!this.accountLinkedKey) {
+      throw new Error('missing account linked key')
+    }
+    let dh1: Uint8Array, dh2: Uint8Array, preKey: SignedPrivateKey
+    if (isRecipient) {
+      if (
+        !this.getLinkedAddress(AccountLinkedRole.INBOX_KEY) ||
```
michaelx11, line 94: Sorry for broken record, but let's double check this will fail rather than returning a garbage address. I think xmtp-js currently allows "garbage" addresses in this method, but the address will get surfaced to the user as an unexpected one.

src/keystore/InMemoryKeystore.ts
```diff
@@ -320,6 +346,11 @@ export default class InMemoryKeystore implements Keystore {
   }
 
   async getAccountAddress(): Promise<string> {
-    return this.v2Keys.getPublicKeyBundle().walletSignatureAddress()
+    if (!this.accountAddress) {
+      this.accountAddress = await this.v2Keys
+        .getPublicKeyBundle()
+        .walletSignatureAddress()
+    }
+    return this.accountAddress
```
michaelx11, line 354: Dumb question: does this still return a promise?

src/crypto/Signature.ts
```diff
@@ -132,6 +190,31 @@ export class WalletSigner implements KeySigner {
     this.wallet = wallet
   }
 
+  private static accountLinkedRoleRequestText(role: AccountLinkedRole): string {
+    switch (role) {
+      case AccountLinkedRole.INBOX_KEY:
+        return 'Create Identity'
+      case AccountLinkedRole.SEND_KEY:
+        return 'Grant Send Permissions'
```
michaelx11, line 198: Glorious

michaelx11 (commented):
Really liking the code structure so far! Added a first pass of comments with only one concern which could be a nothing-burger but worth verifying

test/Invitation.test.ts
```diff
@@ -43,7 +43,7 @@ describe('Invitations', () => {
         SealedInvitation.fromBytes(newInvitation.toBytes()).toBytes()
       )
       // Ensure the headers haven't been mangled
-      const v1 = newInvitation.v1
+      const v1 = newInvitation.v1!
```
neekolas, line 46: I thought we had a lint rule warning on non-null assertions

src/keystore/InMemoryKeystore.ts
michaelx11: fancy, I didn't realize it autoconverted to a promise

src/crypto/PublicKey.ts
michaelx11: I thought I remembered from mucking around with tests a while back, for example with this code change:
```
diff --git a/src/crypto/Signature.ts b/src/crypto/Signature.ts
index 808838b..e091231 100644
--- a/src/crypto/Signature.ts
+++ b/src/crypto/Signature.ts
@@ -36,6 +36,7 @@ export function ecdsaSignerKey(
   digest: Uint8Array,
   signature: ECDSACompactWithRecovery
 ): UnsignedPublicKey | undefined {
+  digest[0] &= 0x1f // clear recovery bit
   const bytes = secp.recoverPublicKey(
     digest,
     signature.bytes,
```

you'll see failures like:

```
  ● InMemoryKeystore › getAccountAddress › returns the wallet address

    expect(received).toEqual(expected) // deep equality

    Expected: "0xacc9B903129E2693b0c09720cBa3972e35666066"
    Received: "0x3Db7e37217f83a89A222Ff7817a957ec393D1bA1"

      516 |       const returnedAddress = await aliceKeystore.getAccountAddress()
      517 |
    > 518 |       expect(aliceAddress).toEqual(returnedAddress)
          |                            ^
      519 |     })
      520 |   })
      521 | })

      at test/keystore/InMemoryKeystore.test.ts:518:28
      at fulfilled (test/keystore/InMemoryKeystore.test.ts:5:58)
```

src/keystore/InMemoryKeystore.ts
```diff
@@ -320,6 +346,11 @@ export default class InMemoryKeystore implements Keystore {
   }
 
   async getAccountAddress(): Promise<string> {
-    return this.v2Keys.getPublicKeyBundle().walletSignatureAddress()
+    if (!this.accountAddress) {
+      this.accountAddress = await this.v2Keys
```
neekolas, line 350: Thanks for optimizing this one

test/Invitation.test.ts
```diff
@@ -43,7 +43,7 @@ describe('Invitations', () => {
         SealedInvitation.fromBytes(newInvitation.toBytes()).toBytes()
```
neekolas, line 43: Would be great to get some tests using `SealedInvitationV2`

src/Invitation.ts
```diff
@@ -67,6 +74,148 @@ export class InvitationV1 implements invitation.InvitationV1 {
...
+      !viewerInboxKeyBundle.accountLinkedKey.matches(
+        header.inboxKeyBundle.accountLinkedKey
+      )
+    ) {
+      throw new Error('Inbox key bundle does not match invite')
+    }
+    const secret = await viewerInboxKeyBundle.sharedSecret(
+      header.sendKeyBundle,
+      header.inboxKeyBundle.preKey,
+      true /* isRecipient */
```
neekolas, line 207: Might be worth adding some comments explaining how `SealedInvitationV2` is never used for self-invites and that's why you can hardcode to this value to `true`

src/crypto/PublicKey.ts
```diff
@@ -211,6 +217,123 @@ export class SignedPublicKey
...
+      )
+    }
+  }
+
+  // Return bytes of the encoded unsigned key.
+  private bytesToSign(): Uint8Array {
+    return this.keyBytes
+  }
+
+  public getLinkedAddress(role: AccountLinkedRole): string {
```
jazzz, line 250: Like seeing the accountLinkedRole System!

I initially assumed validation would occur on initialization, but I can see why coupling it with getAddress makes sense and keeps the number of types under control.

src/keystore/InMemoryKeystore.ts
```diff
@@ -197,27 +200,50 @@ export default class InMemoryKeystore implements Keystore {
...
+          const headerTime = sealed.v2.header.createdNs
+          if (!headerTime.equals(timestampNs)) {
+            throw new Error('envelope and header timestamp mismatch')
+          }
+
+          const invitation = await sealed.v2.getInvitation(
+            PrivateKeyBundleV3.fromLegacyBundle(
+              this.v2Keys,
+              AccountLinkedRole.INBOX_KEY
+            )
```
jazzz, line 235: This is going to be problematic with InboxKey Rotation, but I'll take that one on. It's going to be a messy refactor given PrivateKeyBundleV1s can no longer be converted to PrivateKeyBundleV3s.

jazzz (commented):
Looking clean especially given all that type wrangling that was needed!

src/crypto/PublicKey.ts
michaelx11: Yeah the way the error was worded made me a bit worried we were counting on the check, even the check now still seems a bit meaningless because it will only ever trigger in exceptionally rare circumstances (I think the r,s values have to hit an edge case in the math).

As a reader/consumer of the API, the `if` statement and error appears to catch "bad" signatures but it'll let basically any signature through. I know this seems like a nitpick but the first time I read it I genuinely though it was meant to be a signature validation check.

One solution: for an `AccountLinkedPublicKey` why not just carry the expected address as part of the object? Semantically it makes sense that an AccountLinkedPublicKey should have an expected account linked with it.  Then we could actually do a full validity check here (recovery key from signature, compare it).

michaelx11 (approved):
LGTM - final comment: For AccountLinked-* objects, should we just make the expected address a first class citizen? Then we can fully validate an AccountLinked object before use. Right now all we can do is recover "some" key and address from a potentially arbitrary signature, and hope our other code cross-references the address somewhere.

src/crypto/PublicKey.ts
```diff
@@ -211,6 +217,118 @@ export class SignedPublicKey
...
+        WalletSigner.accountLinkRequestText(this.bytesToSign(), role)
+      )
+      if (!equalBytes(this.staticSignature.text, expectedTextBytes)) {
+        throw new Error('Signature text does not match expected text')
+      }
+      const digest = hexToBytes(utils.hashMessage(this.staticSignature.text))
+      const signature = this.staticSignature.walletEcdsaCompact
+      const publicKey = ecdsaSignerKey(digest, signature)
+      if (!publicKey) {
+        throw new Error('Unable to derive address from signature')
```
michaelx11, line 262: Nice, I do like this error message a lot better. Still wondering if we should just keep an expected peerAddress with the AccountLinkedKey so we can actually validate.

michaelx11:
@richardhuaaa sounds good, we can defer this. I see what you're saying w.r.t. no additional security, in that a user can trivially derive the wallet address from the signature (even if it's a completely random address) - at which point the user must ensure that the wallet address is who they expected to talk to.

However, I think there is potential value in storing that peerAddress up-front in terms of developer ergonomics and avoiding accidental confusion.

Currently validating an AccountLinkedKey signature doesn't ever really happen. Even fraudulent AccountLinkedKeys are "valid" AccountLinkedKeys but just for some random unpredictable address X. I'm wondering if we by being more explicit we can give the developer (and us) a chance to look at an AccountLinkedKey up front and say "this is a valid signature".

PR #306, opened 2023-03-20
DecodedMessage Serialization

neekolas:
## Summary

- Many developers have been asking for easy and reliable ways to serialize messages so that they can be stored and cached in their application. This makes use of a new protobuf type to serialize `DecodedMessage` instances, including the raw decrypted content bytes. Other client instances can then restore this content and decode the content (including content that decodes to types that are not easily serializable themselves).
- I did some cleanup on the `encrypt` and `decrypt` methods to make it possible to extract the result type/throw for errors in a type safe manner.

## Links
https://github.com/xmtp-labs/hq/issues/937


test/Message.test.ts
```diff
@@ -1,22 +1,34 @@
...
+import { Wallet } from 'ethers'
+import { ContentTypeTestKey, TestKeyCodec } from './ContentTypeTestKey'
 
 describe('Message', function () {
+  let aliceWallet: Wallet
+  let bobWallet: Wallet
+  let alice: PrivateKeyBundleV1
+  let bob: PrivateKeyBundleV1
+
+  beforeEach(async () => {
```
neekolas, line 20: Added this `beforeEach` step to reduce some of the boilerplate in these tests

bench/decode.ts
```diff
@@ -75,7 +75,8 @@ const decodeV2 = () => {
         new Date(),
         undefined
       )
-      const encodedMessage = await convo.encodeMessage(message)
+      const payload = await alice.encodeContent(message)
+      const encodedMessage = await convo.createMessage(payload)
```
jazzz, line 79: This logical division makes sense to me

jazzz (approved):

src/Message.ts
```diff
@@ -243,19 +239,21 @@ export class DecodedMessage {
...
   error?: Error
+  contentBytes: Uint8Array
 
   constructor({
     id,
     messageVersion,
     senderAddress,
     recipientAddress,
     conversation,
+    contentBytes,
```
michaelx11, line 250: probably dumb question, but why do we need both `contentBytes` and `content: any` members?

michaelx11: Oh I think I get it, usually you'll have both members but when deserializing off disk we start with `contentBytes` and then use `decodeContent` to get `content`.

michaelx11 (approved):
LGTM, this is effectively enabling the start of post-delivery storage right? Albeit in a non-prescriptive way

src/Message.ts
neekolas: Bingo.

I thought about making `contentBytes` a private property, but decided it was harmless to expose and might be useful for constructing test cases.

neekolas:
> this is effectively enabling the start of post-delivery storage right? Albeit in a non-prescriptive way

Definitely one of the requirements for PDS. With this feature, developers could build their own PDS today.

PR #307, opened 2023-03-21
Export buildDirectMessageTopic and buildDirectMessageTopicV2 functions

nplasterer:
This exports the `buildDirectMessageTopic` so consumers can construct topics easier

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,3 +43,4 @@ export {
   toNanoString,
   mapPaginatedStream,
 } from './utils'
+export { buildDirectMessageTopic } from './utils'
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -43,3 +43,4 @@ export {
   toNanoString,
   mapPaginatedStream,
 } from './utils'
+export { buildDirectMessageTopic } from './utils'
```
nplasterer, line 46: Should we also do V2 for good measure?

```suggestion
export { buildDirectMessageTopic, buildDirectMessageTopicV2 } from './utils'
```

yijuchung:
Thanks for adding this.

neekolas (approved):

PR #313, opened 2023-03-23
Limit queries across topics

neekolas:
## Summary

As part of our decentralization roadmap, we are limiting the ability to query across multiple topics in a single query. This will allow for easier sharding across the network as `topic` will be the sharding key. The API endpoint will remain the same, but will throw an exception if the `contentTopics` array has a length > 1.

This adjusts our SDK to only allow for querying on a single topic.

## Notes

The ability to query across multiple topics is a relatively unused feature in our SDK, as none of our higher level methods take advantage of it. There are only a few usages of this feature in clients that use the `listEnvelopes` API directly. Those will break and will need to be adjusted as part of the v8 migration. `batchQuery` is a reasonable replacement for most of those use-cases.

### DIFF
--- a/src/ApiClient.ts
+++ b/src/ApiClient.ts
@@ -45,7 +45,7 @@ export type GrpcError = Flatten<Error & { code?: GrpcStatus }>
 export type QueryParams = {
   startTime?: Date
   endTime?: Date
-  contentTopics: string[]
+  contentTopic: string
 }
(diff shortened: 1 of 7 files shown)


michaelx11:
LGTM - still draft phase though

michaelx11 (approved):

PR #314, opened 2023-03-24
feat: SIWE signature verification code

michaelx11:
**Overview**

Initial support for SIWE signatures in AccountLinkedRoles. We only use the Resources part of the SIWE, and it's open to any and all feedback. The goal here is to be able to read these SIWE signatures (lock down the Resource / SIWE text spec) but actually constructing them is a more intricate problem. In this PR, we provide default XMTP-related SIWE account linking message generation, but most apps will want to use their own Login SIWE text with our Resource string.

**Tests**
- Unit tests

I was using an old version of SIWE. All good now. ~~NOTE: tests are failing because I had to force upgrade for `siwe` package, looking now. Downgraded `ethers` from `^5.5.3` to `^5.5.1` (siwe seems to want 5.5.1).~~

### DIFF
--- a/src/crypto/PrivateKeyBundle.ts
+++ b/src/crypto/PrivateKeyBundle.ts
@@ -7,6 +7,7 @@ import {
 import {
   AccountLinkedRole,
   StaticWalletAccountLinkSigner,
+  SIWEWalletAccountLinkSigner,
   WalletSigner,
 } from './Signature'
 import { PublicKey, SignedPublicKey } from './PublicKey'
@@ -56,8 +57,8 @@ export class PrivateKeyBundleV3 implements proto.PrivateKeyBundleV3 {
     wallet: Signer,
     role: AccountLinkedRole
   ): Promise<PrivateKeyBundleV3> {
-    const accountLinkedKey = await AccountLinkedPrivateKey.generateSIWE(
-      new StaticWalletAccountLinkSigner(wallet),
+    const accountLinkedKey = await AccountLinkedPrivateKey.generate(
+      new SIWEWalletAccountLinkSigner(wallet),
       role
     )
     const bundle = new PrivateKeyBundleV3({
--- a/src/crypto/PrivateKeyBundle.ts
+++ b/src/crypto/PrivateKeyBundle.ts
@@ -49,6 +49,25 @@ export class PrivateKeyBundleV3 implements proto.PrivateKeyBundleV3 {
     return bundle
   }
 
+  // Generate a new key bundle with the preKey signed by the accountLinkedKey.
+  // Sign the accountLinkedKey with the provided wallet as well. Same as above
+  // but uses SIWE format signature text
+  static async generateSIWE(
+    wallet: Signer,
+    role: AccountLinkedRole
+  ): Promise<PrivateKeyBundleV3> {
+    const accountLinkedKey = await AccountLinkedPrivateKey.generateSIWE(
+      new StaticWalletAccountLinkSigner(wallet),
+      role
+    )
+    const bundle = new PrivateKeyBundleV3({
+      accountLinkedKey,
+      preKeys: [],
+    })
+    await bundle.addPreKey()
+    return bundle
+  }
+
   public static fromLegacyBundle(
     bundle: PrivateKeyBundleV2,
     role: AccountLinkedRole
--- a/src/crypto/PrivateKeyBundle.ts
+++ b/src/crypto/PrivateKeyBundle.ts
@@ -52,7 +52,9 @@ export class PrivateKeyBundleV3 implements proto.PrivateKeyBundleV3 {
 
   // Generate a new key bundle with the preKey signed by the accountLinkedKey.
   // Sign the accountLinkedKey with the provided wallet as well. Same as above
-  // but uses SIWE format signature text
+  // but uses SIWE format signature text. NOTE: in practice, consumers should
+  // only use this method if they're okay with using a default SIWE rather than
+  // reusing their app-specific login SIWE.
   static async generateSIWE(
     wallet: Signer,
     role: AccountLinkedRole
--- a/src/crypto/PublicKey.ts
+++ b/src/crypto/PublicKey.ts
@@ -252,8 +252,8 @@ export class AccountLinkedPublicKeyV1
     return this.keyBytes
   }
 
-  public getSIWELinkedAddressWithoutDomainCheck(
-    expectedAddress: string,
+  // Return the extracted address from the SIWE signature checking role and keys, but not domain
+  private getSIWELinkedAddressWithoutDomainCheck(
     role: AccountLinkedRole
   ): string {
     // The criteria for successful SIWE-signed keys with roles are:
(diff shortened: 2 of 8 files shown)


michaelx11:
Wow I hate spruce/siwe - I think they're hardcoding the `ethers` version their library expects and this is causing a dependency conflict.

src/crypto/PublicKey.ts
```diff
@@ -248,6 +252,62 @@ export class AccountLinkedPublicKeyV1
...
+    //   return false
+    // }
+    // Rule 2: Check the resources
+    const resources = siwe.resources || []
+    const roleString =
+      StaticWalletAccountLinkSigner.accountLinkedSIWERoleRequestText(role)
+    const resourceRegex = new RegExp(
+      `https://xmtp.org/${roleString}/${bytesToHex(this.bytesToSign())}`
+    )
+    if (!resources.some((resource) => resourceRegex.test(resource))) {
```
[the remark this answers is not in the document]
michaelx11: Good call I'll change it to exact match!

src/crypto/Signature.ts
```diff
@@ -242,6 +296,11 @@ export class WalletSigner implements KeySigner {
 }
 
 export interface AccountLinkSigner {
+  signKeyWithRoleSIWE(
+    key: UnsignedPublicKey,
+    role: AccountLinkedRole
+  ): Promise<AccountLinkedPublicKey>
+
```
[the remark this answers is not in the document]
michaelx11: Sounds good

michaelx11:
@jazzz CYTAL when you get a chance?

src/crypto/PublicKey.ts
```diff
@@ -248,6 +253,63 @@ export class AccountLinkedPublicKeyV1
...
+    //   return false
+    // }
+    // Rule 2: Check the resources
+    const resources = siwe.resources || []
+    const roleString =
+      SIWEWalletAccountLinkSigner.accountLinkedSIWERoleRequestText(role)
+
+    const expectedResource = `https://xmtp.org/siwe/${roleString}/secp256k1/${bytesToHex(
+      this.bytesToSign()
+    )}`
```
jazzz, line 290: 👍 , This is under discussion, but can be changed in the future

src/crypto/PublicKey.ts
```diff
@@ -248,6 +253,63 @@ export class AccountLinkedPublicKeyV1
...
+    }
+
+    // Rule 3: Finally, check the signature
+    const digest = hexToBytes(utils.hashMessage(siweSignature.text))
+    const signerKey = ecdsaSignerKey(digest, siweSignature.walletEcdsaCompact)
+    if (!signerKey) {
+      throw new Error('Failed to recover signer key')
+    }
+    const signerAddress = signerKey.getEthereumAddress()
+    if (signerAddress !== siwe.address) {
```
jazzz, line 304: 👍

src/crypto/Signature.ts
```diff
@@ -248,14 +302,107 @@ export interface AccountLinkSigner {
...
+    const keyBytes = key.toBytes()
+    const resource =
+      SIWEWalletAccountLinkSigner.accountLinkedSIWEResourceRoleText(
+        keyBytes,
+        role
+      )
+    const siwe = new SiweMessage({
+      statement: 'XMTP Account Link with Role: ' + role,
+      address: walletAddress,
+      domain: 'xmtp.org',
```
jazzz, line 353: What would it mean for a user to sign a "GrantSendPermissions" from xmtp.org?

src/crypto/PublicKey.ts
```diff
@@ -248,6 +253,63 @@ export class AccountLinkedPublicKeyV1
...
+    // if (siwe.domain !== domain) {
+    //   console.log(`Expected domain ${domain} but got ${siwe.domain}`)
+    //   return false
+    // }
+    // Rule 2: Check the resources
+    const resources = siwe.resources || []
+    const roleString =
+      SIWEWalletAccountLinkSigner.accountLinkedSIWERoleRequestText(role)
+
+    const expectedResource = `https://xmtp.org/siwe/${roleString}/secp256k1/${bytesToHex(
```
jazzz, line 288: This string is coded in two locations, here and the validation step. Unifying these will reduce the number of updates needed, and be easier to maintain

src/crypto/PrivateKeyBundle.ts
```diff
@@ -49,6 +50,25 @@ export class PrivateKeyBundleV3 implements proto.PrivateKeyBundleV3 {
     return bundle
   }
 
+  // Generate a new key bundle with the preKey signed by the accountLinkedKey.
+  // Sign the accountLinkedKey with the provided wallet as well. Same as above
+  // but uses SIWE format signature text
+  static async generateSIWE(
+    wallet: Signer,
+    role: AccountLinkedRole
+  ): Promise<PrivateKeyBundleV3> {
```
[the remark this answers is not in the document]
michaelx11: Yup, will not be called in the wild most likely. I have a comment on a subfunction explaining it but I'll also describe in this method's comment too. It's currently a default/example/helper for testing. You could imagine someone might call this, if they didn't care to reuse their SIWE for login (i.e. do a 1:1 replacement of today's CreateIdentity signature)

src/crypto/PrivateKeyBundle.ts
```diff
@@ -49,6 +50,25 @@ export class PrivateKeyBundleV3 implements proto.PrivateKeyBundleV3 {
...
+    const accountLinkedKey = await AccountLinkedPrivateKey.generate(
+      new SIWEWalletAccountLinkSigner(wallet),
+      role
```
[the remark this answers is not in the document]
michaelx11: Right now it does one role at a time per SIWE - we can definitely add multi-role in single signature support in a subsequent PR. The main thrust of this PR is to validate SIWE signatures though

src/crypto/PublicKey.ts
```diff
@@ -248,6 +253,63 @@ export class AccountLinkedPublicKeyV1
...
   }
 
+  // Return the extracted address from the SIWE signature checking role and keys, but not domain
+  private getSIWELinkedAddressWithoutDomainCheck(
+    role: AccountLinkedRole
+  ): string {
+    // The criteria for successful SIWE-signed keys with roles are:
+    // 1. The text must be a valid SIWE for the expected domain (TBD)
+    // 1b. Check expiry (TBD)
+    // 2. The resources must contain the "https://[role]/(keybytes as hex, uncompressed){64}" with no
```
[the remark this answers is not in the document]
michaelx11: yup, outdated will update

src/crypto/PublicKey.ts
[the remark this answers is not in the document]
michaelx11: sounds good, I'll switch over to base64

src/crypto/Signature.ts
michaelx11: Not much at the moment, but we need a domain placeholder to satisfy the SIWE format. I think I mention in the comment but this illustrates the flow or provides a default SIWE message. Main goal of this PR is to add the code that can validate a SIWE signature

src/crypto/PublicKey.ts
michaelx11: Yup, will consolidate

jazzz (approved):
Looks good for beta!

src/crypto/PublicKey.ts
michaelx11: Feels great to be able to this check! A solid vote in my book for SIWE

PR #317, opened 2023-03-29
Use mutex for v2 conversation listing

neekolas:
## Summary

I found a performance regression while testing v8. Previously, when calling `conversations.list()` twice, the second call would wait until the first completed because of the mutex on the `ConversationCache`. This is no longer the case, since the caching has moved to the Keystore. This change restores the previous behaviour.


michaelx11 (approved):
The gist is we could kick off two very expensive conversation list operations that would both run to completion? Whereas previously the first would run and the second queued (behind a mutex) would use the cached result? LGTM

neekolas:
Exactly

PR #318, opened 2023-03-29
Respect persistConversations flag

neekolas:
## Summary

Realized in testing that we were not consistently respecting the `persistConversations` flag across all keystore providers. This change moves the logic to a helper function so that all providers work the same. Also adds a test to avoid regressions.

### DIFF
--- a/test/keystore/providers/NetworkKeyManager.test.ts
+++ b/test/keystore/providers/NetworkKeyManager.test.ts
@@ -1,6 +1,8 @@
 import ApiClient, { ApiUrls } from '../../../src/ApiClient'
 import { PrivateKeyBundleV1 } from '../../../src/crypto/PrivateKeyBundle'
+import PrefixedPersistence from '../../../src/keystore/persistence/PrefixedPersistence'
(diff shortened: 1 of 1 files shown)


src/keystore/providers/helpers.ts
```diff
@@ -11,6 +11,9 @@ export const buildPersistenceFromOptions = async (
   opts: KeystoreProviderOptions,
   keys: PrivateKeyBundleV1 | PrivateKeyBundleV2
 ) => {
+  if (!opts.persistConversations) {
```
michaelx11, line 14: This is a helper that all existing JS keystore implementations consume, but other keystore impls have their own responsibility to process this flag correctly?

michaelx11 (approved):
LGTM

src/keystore/providers/helpers.ts
neekolas: That's right, but it's actually the `KeystoreProvider` implementations that have to respect this flag. The Keystore takes whatever persistence is passed in to it from the provider.

michaelx11: Ah gotcha

PR #320, opened 2023-03-31
PROTOTYPE: Snap Keystore

neekolas:
## Summary

- Implements a Snap Keystore, which proxies all Keystore API requests to a MM Snap
- Uses a declarative model for defining both the sending and receiving handlers. Just encodes/decodes requests/responses based on a declared mapping of API methods to proto types
- Gracefully falls back to existing behaviour if the user doesn't have a Snaps-compatible browser. 

## Todo
Listing a bunch of todos, but with no intention of actually doing them any time soon

- Package up the Snap as a NPM module
- Proper error handling
- Permissions management for applications
- Better type safety of API handlers
- Maybe update `getV2Conversations` to return a single proto object instead of an array, to make generating handlers easier
- Performance profiling
-

### DIFF
--- a/src/keystore/SnapKeystore.ts
+++ b/src/keystore/SnapKeystore.ts
@@ -1,0 +1,137 @@
+import { Keystore } from './interfaces'
+import {
+  fetcher,
+  conversationReference,
+  keystore,
+  authn,
+  publicKey,
+  signature,
+} from '@xmtp/proto'
+import { Reader, Writer } from 'protobufjs/minimal'
+const { b64Encode, b64Decode } = fetcher
+
+type Codec<T> = {
+  decode(input: Reader | Uint8Array, length?: number): T
+  encode(message: T, writer?: Writer): Writer
+}
+type ApiDefs = {
+  [k: string]: {
+    req: Codec<any> | null
+    res: Codec<any>
+  }
+}
+
+export const defaultSnapOrigin = `local:http://localhost:8080`
+
+export const apiDefs: ApiDefs = {
+  decryptV1: {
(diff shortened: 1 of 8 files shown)


src/keystore/SnapKeystore.ts
```diff
@@ -0,0 +1,133 @@
...
+        return getResponse(method as keyof Keystore, null, apiDef.res)
+      }
+      const reqBytes = apiDef.req.encode(req).finish()
+      return getResponse(method as keyof Keystore, reqBytes, apiDef.res)
+    }
+  }
+
+  return {
+    ...generatedMethods,
+    async getV2Conversations() {
```
neekolas, line 115: I knew having the response type be `ConversationReference[]` for this one was going to be annoying, but it bothers me even more than expected. Considering sneaking in a change to v8 to make this consistent with the other APIs and return a single response object with a nested array.

src/keystore/providers/SnapProvider.ts
```diff
@@ -0,0 +1,88 @@
...
+        snap.id === defaultSnapOrigin && (!version || snap.version === version)
+    )
+  } catch (e) {
+    console.warn('Failed to obtain installed snap', e)
+    return undefined
+  }
+}
+
+const initSnap = async () => {
+  await window.ethereum.request({
```
neekolas, line 52: I have a special `init` method on the Snap, which is the only time it asks for wallet signatures (and only if the keys are not available in MM storage)

neekolas:
Closing this, since we are getting closer to a non-prototype version

PR #321, opened 2023-04-02
feat: add support for ephemeral topics

nakajima:
Also adds a content type for typing notifications.


neekolas:
Given how close v8 is to release, I think this PR should be done against the beta branch.

src/codecs/TypingNotification.ts
```diff
@@ -0,0 +1,49 @@
+import { ContentTypeId, ContentCodec, EncodedContent } from '../MessageContent'
+
+// xmtp.org/typingNotification
+//
+// This content type is used for typing notifications
+export const ContentTypeTypingNotification = new ContentTypeId({
```
neekolas, line 6: We should include some strongly worded comments here that this content type should only be used on ephemeral topics, if we can't enforce or warn at runtime. 

Otherwise someone perusing the code might be inclined to use the two features in this PR separately.

src/conversations/Conversation.ts
```diff
@@ -313,6 +313,10 @@ export class ConversationV1 {
     return decoded
   }
 
+  streamEphemeral(): null {
+    return null
```
neekolas, line 317: With how the Conversation Union type works, this is going to make usage awkward. The caller doesn't know whether they have a V1 or V2 conversation without an explicit type check. So, calling `streamEphemeral` could either give you a Promise of a stream or null. Might make more sense to keep the types consistent and just throw an error. Or to figure out how we make ephemeral topics work for V1 conversations.

src/codecs/TypingNotification.ts
```diff
@@ -0,0 +1,49 @@
...
+    return ContentTypeTypingNotification
+  }
+
+  encode(content: TypingNotification): EncodedContent {
+    return {
+      type: ContentTypeTypingNotification,
+      parameters: {
+        timestamp: content.timestamp.toISOString(),
+        typerAddress: content.typerAddress,
+        isFinished: content.isFinished ? 'true' : 'false',
```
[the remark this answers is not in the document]
nakajima: >  I wonder if regularly sending a keep-alive payload with a TTL for the duration that typing is happening would be more appropriate

That's how I've had it implemented, but I was finding there were some races that would lead to the typing indicator showing up after a message had come in. This field was a way to just tell listeners "we're done here".

src/codecs/TypingNotification.ts
nakajima: I don't think there's currently a way to check at runtime. Thoughts on adding a field to `ContentCodec` like `ephemeralOnly`? That'd give us something we could check.

nakajima:
Ok, updated the target to the `beta` branch. There's a couple changes from when I forked off main included in here too. I added a comment to the typing notification content type to say it should only be used on ephemeral topics. If we want to add a field to `ContentType` to say it should be ephemeral only, we can do that in a follow up.

src/conversations/Conversation.ts
```diff
@@ -27,9 +27,29 @@ import {
...
 } from '../crypto'
+import { PreparedMessage } from '../PreparedMessage'
 import { sha256 } from '../crypto/encryption'
-import { ContentTypeText } from '../codecs/Text'
 import { buildDecryptV1Request, getResultOrThrow } from '../utils/keystore'
+import { ContentTypeText } from '../codecs/Text'
+
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+
+type ConversationV1Export = {
```
neekolas, line 37: This must be a weird merge conflict. These types are removed in beta/v8

neekolas (commented):
This looks pretty good to me. 

I do worry about the lack of support for v1 conversations. Generally the SDK doesn't require the developer to pay attention to which type of conversation they are interacting with, since we otherwise have feature parity. Developers will need to be defensive about errors whenever trying to stream ephemeral messages, and typing indicators will feel broken in random and arbitrary ways when someone hits an old V1 convo where this isn't supported.

src/codecs/TypingNotification.ts
```diff
@@ -0,0 +1,49 @@
...
+      },
+      content: new Uint8Array(),
```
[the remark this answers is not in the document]
nakajima: They're simple key/value pairs, I thought it'd be a bit easier to do it this way than worry about how to encode stuff in the content field. I don't feel super strongly about this though.

src/conversations/Conversation.ts
nakajima: Removed in https://github.com/xmtp/xmtp-js/pull/321/commits/7b1131f2c57f2d8347c59e16f9723400e1d00490.

nakajima:
> I do worry about the lack of support for v1 conversations.

You've convinced me, opened https://github.com/xmtp/xmtp-node-go/pull/265 to support v1 ephemeral topics. Added SDK support in https://github.com/xmtp/xmtp-js/pull/321/commits/6754aee77d3f5d7ca7adbbc715eadc9ef649159a.

neekolas (approved):
Impressively fast turnaround on this thing @nakajima. Love it

PR #322, opened 2023-04-03
feat: provide xmtpv3_wasm example integration in VoodooClient class

michaelx11:
## Overview

As part of prototyping with Vodozemac - this PR provides a very preliminary example of integrating with our WASM bindings library on local disk. It expects your `xmtp-js` repo to have `libxmtp` (prototype, not public yet) as a sibling in the directory hierarchy.

## Changes
- `Client.generate` initializes the WASM module and returns a new instance of a Vodozemac account + some functionality, colloquially named Voodoo (because it's magical)
- A new set of Client options governs the Voodoo integration
- We add an example `VoodooManager` class that can be discarded or used as a reference for how we can consolidate Voodoo-specific code

## Test Plan

- All tests still pass locally (not in Github Action unfortunately as `libxmtp` is unavailable)


src/Client.ts
```diff
@@ -306,6 +327,7 @@ export default class Client {
   }
 
   private async ensureUserContactPublished(legacy = false): Promise<void> {
+    // TODO (VOODOO): this should do something, but currently does nothing for Voodoo contacts
```
neekolas, line 330: I would think as part of the "green pasture" approach, we'll want to keep this function as-is and create a parallel version in the VoodooManager or somewhere else fresh.

Making changes to these core methods is going to break our test suite and existing functionality

michaelx11: It's a good callout. I was debating doing an opt-based conditional in each of the main methods or just writing a `VoodooClient`. Your opinion is a new client class would be better right?

src/Client.ts
```diff
@@ -577,6 +620,27 @@ export default class Client {
     )
   }
 
+  /**
+   * List raw message bytes of envelopes stored on a specific topic
+   *
+   * Useful for Voodoo testing
+   */
+  async listRawVoodoo(topic: string): Promise<Uint8Array[]> {
```
neekolas, line 628: This also looks a whole lot like `listEnvelopes`, just without the mapping function (which could be as simple as `(env) => env`)

src/Client.ts
```diff
@@ -497,6 +521,25 @@ export default class Client {
     }
   }
 
+  /**
+   * Low level no validation raw byte publishing for Voodoo debugging
+   */
+  async publishVoodooRaw(
```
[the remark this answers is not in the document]
michaelx11: Yeah i'll remove for now - when I did my hacky prototype I just wanted to publish bare-metal unencoded unvalidated bytes

src/Client.ts
neekolas: I do think that. Changing existing methods is going to be an endless task of fixing broken tests and keeping the type checker happy for little benefit in terms of learning.

I think a new client + the existing `apiClient` gets us the most learning the fastest.

neekolas: We'd have to copy over 2 methods from the old client to the new one `publishEnvelopes` + `listEnvelopes` and we should be able to test end to end.

src/VoodooManager.ts
```diff
@@ -0,0 +1,108 @@
+// TODO: this is a hacky wrapper class for a Voodoo contact,
```
neekolas, line 1: We should probably make a `voodoo` folder to contain this and all the other new classes we are going to be creating.

package.json
```diff
@@ -72,8 +73,10 @@
     "async-mutex": "^0.4.0",
     "elliptic": "^6.5.4",
     "ethers": "^5.5.3",
+    "libxmtp": "file:../libxmtp/bindings/wasm",
     "long": "^5.2.0",
-    "siwe": "^2.1.3"
+    "siwe": "^2.1.3",
+    "xmtpv3": "file:../libxmtp/bindings/wasm"
```
neekolas, line 79: Why two dependencies on the same package?

michaelx11: great catch - probably leftover from when I tried and failed to `npm link`

src/VoodooClient.ts
```diff
@@ -0,0 +1,117 @@
...
+    // Get the contact info which is just a handle for now
+    const contactInstance = this.contacts.get(contactAddress)?.voodooInstance
+    if (!contactInstance) {
+      throw new Error(`No contact info for ${contactAddress}`)
+    }
+    const outbound = await this.voodooInstance.createOutboundSession(
+      contactInstance,
+      initialMessage
+    )
+    return outbound
```
neekolas, line 68: Won't we need to also get the `session_id` to be able to do anything with this?

michaelx11: Yup, we need to add some session tracking when we integrate `encrypt` and `decrypt`. Otherwise all we can do is what I did in my crappy prototype: decrypt the very first message

src/VoodooManager.ts
michaelx11: Missed this, I'll make it happen

package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "prebench": "npm run build:cjs",
     "bench": "node dist/cjs/bench/index.js",
-    "build": "npm run build:cjs && npm run build:esm && npm run build:umd && npm run build:types",
+    "build": "npm run build:cjs && npm run build:esm && npm run build:umd && npm run build:types && npm run package",
```
neekolas, line 13: Do we need to run `npm run package` here? That should only be for publishing

src/voodoo/VoodooClient.ts
```diff
@@ -0,0 +1,118 @@
...
+  async processInboundSessionJson(
+    contactAddress: string,
+    inboundJson: string
+  ): Promise<string> {
+    // Get the contact info which is just a handle for now
+    const contactInstance = this.contacts.get(contactAddress)?.voodooInstance
+    if (!contactInstance) {
+      throw new Error(`No contact info for ${contactAddress}`)
+    }
+    const inboundPlaintext = await this.voodooInstance.createInboundSession(
```
neekolas, line 112: This type is going to change to be an object now that my change has been merged in

src/voodoo/VoodooClient.ts
```diff
@@ -0,0 +1,118 @@
...
+  async getOutboundSessionJson(
+    contactAddress: string,
+    initialMessage: string
+  ): Promise<string> {
+    // Get the contact info which is just a handle for now
+    const contactInstance = this.contacts.get(contactAddress)?.voodooInstance
+    if (!contactInstance) {
+      throw new Error(`No contact info for ${contactAddress}`)
+    }
+    const outbound = await this.voodooInstance.createOutboundSession(
```
neekolas, line 65: Also now an object instead of a string with my latest change

package.json
michaelx11: I can remove, I had it in there because I consumed the .tar.gz in xmtp-inbox-web

src/voodoo/VoodooClient.ts
michaelx11: I realize we're not exporting `VoodooInstance` or any other types in `xmtpv3_wasm`. I can make a PR for that so we can consume those types here.

test/Client.test.ts
```diff
@@ -291,3 +291,37 @@ describe('ClientOptions', () => {
     })
   })
 })
+
+describe('VoodooClient', () => {
```
neekolas, line 295: nit: probably want a separate test file for the voodoo stuff.

neekolas (approved):

src/voodoo/VoodooClient.ts
```diff
@@ -0,0 +1,110 @@
...
+  setContact(contactInfo: VoodooContact) {
+    this.contacts.set(contactInfo.address, contactInfo)
+  }
+
+  getContact(address: string): VoodooContact | undefined {
+    return this.contacts.get(address)
+  }
+
+  // Get the JSON from creating an outbound session with a contact
+  async getOutboundSessionJson(
```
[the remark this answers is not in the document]
michaelx11: Yup! Just the outbound encrypted message pickled as JSON - the naming sucks a bit and can be reworked. It's meant to indicate that it's a PreKey message that kicks off a session

test/Client.test.ts
michaelx11: Good call, I'll move it out into a separate file

PR #323, opened 2023-04-04
feat: publish then use voodoo identities

michaelx11:
## Overview

Uses https://github.com/xmtp/libxmtp/pull/40 to publish a contact bundle and fetch it without verification.

## Changes

- Add publishing and fetching of contacts from the network using Voodoo-topics in VoodooClient.ts

## Test Plan

- Change the roundtrip test to use the new contacts, instead of passing the contacts around inside the test itself

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "ethers": "^5.5.3",
     "long": "^5.2.0",
     "siwe": "^2.1.3",
-    "xmtpv3_wasm": "github:xmtp/libxmtp#michaelx_export_and_import_of_vodozemac_identities"
+    "xmtpv3_wasm": "github:xmtp/libxmtp#vodozemac_dev"
   },
   "devDependencies": {
     "@commitlint/cli": "^16.1.0",
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "ethers": "^5.5.3",
     "long": "^5.2.0",
     "siwe": "^2.1.3",
-    "xmtpv3_wasm": "github:xmtp/libxmtp#vodozemac_dev"
+    "xmtpv3_wasm": "github:xmtp/libxmtp#michaelx_export_and_import_of_vodozemac_identities"
   },
   "devDependencies": {
     "@commitlint/cli": "^16.1.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -74,7 +74,7 @@
     "ethers": "^5.5.3",
     "long": "^5.2.0",
     "siwe": "^2.1.3",
-    "xmtpv3_wasm": "github:xmtp/libxmtp#vodozemac_dev"
+    "xmtpv3_wasm": "github:xmtp/libxmtp#michaelx_export_and_import_of_vodozemac_identities"
```
michaelx11, line 77: Will change the branch once the other PR: https://github.com/xmtp/libxmtp/pull/40 goes in

src/voodoo/utils.ts
```diff
@@ -0,0 +1,20 @@
+import { utils } from 'ethers'
+import { buildContentTopic } from '../utils'
```
neekolas, line 2: Just to further avoid conflicts, and to make the topic restriction rules easier to work with, what do we think about bumping the version in the root of the proto to `xmtp/1/`

michaelx11: Sounds good to me, currently not using any protos which is probably questionable

michaelx11: Oh, yeah I'll make the change to the version number!

neekolas (approved):
:shipit: 

What's the next step here? Is it time to take off the training wheels (storing all identities in memory and passing the full account around with the public key?)

michaelx11:
Yup, I'm hoping to get to a point where two separate xmtp-js processes (running VoodooClient) can have a conversation over the `local` network.

My plan of attack:
- Define the crappiest of VoodooMessage types (timestamp, sender, content)
- Build out VoodooConversation and VoodooConversations. VoodoConversation stores the sessionId and a list of VoodooMessages in memory
- Run a unit test using `local` where Alice and Bob look each other up by address on the network.
1. Alice initiates a conversation where the outbound JSON blob (pickled Olm PreKey Message) goes into the voodoo invite topic.
2. Bob processes the invite which results in a VoodooConversation.
3. Both sides send a few messages back and forth.

PR #326, opened 2023-04-06
feat: implement deterministic v2 conversation topics

nakajima:
Attempting to make v2 conversation topics/keys deterministic, the goal being to reduce duplicate conversations happening.

The rough description of the approach is:
```
TopicKey = hkdf(symmetric_aes_key)
TopicName = Hmac(symmetric_aes_key, {participants, conversationId})
```

I could definitely use some help making sure I actually did that, since I'm not super familiar with a lot of the cryptographic functions we're using here.


src/keystore/InMemoryKeystore.ts
```diff
@@ -272,9 +277,43 @@ export default class InMemoryKeystore implements Keystore {
...
+            await recipient.walletSignatureAddress(),
+          ].sort(),
+        })
+      )
+
+      const topic = sha256(
+        await hmacSha256Sign(Buffer.from(secret), Buffer.from(msg))
+      )
+
+      const salt = getRandomValues(new Uint8Array(KDFSaltSize))
```
neekolas, line 304: Hrmmm. If we have random values in the salt, then repeated invites are going to be different. @jazzz thoughts on what we should use as a salt value here?

nakajima: Does it matter if the invite is different if the topic is still the same?

neekolas: Unfortunately so, since we are only going to store one key per topic in the `InviteStore`. And if we had more than one we would need to somehow choose the right one for each message.

neekolas: The salt is optional in HKDF, but I'm not sure what the practical implications are to omit

src/keystore/InMemoryKeystore.ts
```diff
@@ -272,9 +277,43 @@ export default class InMemoryKeystore implements Keystore {
...
       const recipient = toSignedPublicKeyBundle(req.recipient)
+
+      const secret = await this.v2Keys.sharedSecret(
+        recipient,
+        this.v2Keys.getCurrentPreKey().publicKey,
+        false
+      )
+
+      const msg = new TextEncoder().encode(
+        JSON.stringify({
```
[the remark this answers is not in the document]
nakajima: I like that, fixed in https://github.com/xmtp/xmtp-js/pull/326/commits/bfa9cfca7931a67248b46d701df76d2622691fa2.

src/keystore/InMemoryKeystore.ts
```diff
@@ -272,9 +277,41 @@ export default class InMemoryKeystore implements Keystore {
           'missing recipient'
         )
       }
-      const invitation = InvitationV1.createRandom(req.context)
+      // const invitation = InvitationV1.createRandom(req.context)
```
michaelx11, line 280: nit: hanging comment

src/keystore/InMemoryKeystore.ts
michaelx11: Yeah this is problematic for the same reason using AES-GCM encrypt to derive `topicKey` [was a problem](https://www.notion.so/RAPID-Dedupe-conversations-via-Determinism-18735a19ed1140fcbf95c2ea801c5422?d=f4e661a6b1d84baf83e8427dc97a79af#9fc9b372d558435cb7f7f31c779b20c9). Any randomness introduced will not be replicated by an independent client making an invite to the same recipient at the same time.

We could deterministically derive the salt as `hmac(key=secret, fixed_string=...)` but I need a second pair of eyes on that scheme.

michaelx11: Salt is optional: https://crypto.stackexchange.com/a/64154 - we should be fine with a fixed salt too. The Diffie-hellman output should provide sufficient entropy when passed through the `hkdf`.

test/keystore/InMemoryKeystore.test.ts
```diff
@@ -457,7 +457,13 @@ describe('InMemoryKeystore', () => {
       const shuffled = [...timestamps].sort(() => Math.random() - 0.5)
 
       await Promise.all(
-        shuffled.map((createdAt) => {
+        shuffled.map(async (createdAt) => {
+          let keys = await PrivateKeyBundleV1.generate(newWallet())
+
+          const recipient = SignedPublicKeyBundle.fromLegacyBundle(
```
michaelx11, line 463: I might be dumb, but it's not obvious to me how this change tests deterministic v2 conversation topics. Should we right a dedicated test with something really targeted? e.g.
```
generate Alice
generate Bob
alice generates 100 invites for bob
confirm all 100 invites have the same topicKey and same topic
confirm the invites differ in timestamp
```

michaelx11 (commented):
Looks good overall - open question about what to do with the salt and would love to see a dedicated test

mkobetic:
I haven't looked at the changes yet, just want to remark that we were also tossing around some other ideas for the topic IDs, e.g. making the topic ID a public key, and requiring anyone submitting to the topic to know the private key (maybe shared in the invite) and sign the publish request with it. That would allow nodes to execute per-topic access control without knowing anything about the participants.

test/keystore/InMemoryKeystore.test.ts
nakajima: This test change was just because it was failing due to the implementation changes. I added a new test to make sure we're using the same topic in https://github.com/xmtp/xmtp-js/pull/326/commits/551c31e2e0a18e8aab1767e62230f1aa61467e28.

I wasn't sure how to get the topic key off the conversation objects we get back from the keystore. Is that an internal thing?

src/keystore/InMemoryKeystore.ts
nakajima: Added a salt constant in https://github.com/xmtp/xmtp-js/pull/326/commits/6632dba65169b2da3f3d384abd29b895cdb481c2.

jazzz: In this case the salt should be static for domain separation, and randomness should be supplied via the info field if necessary. 

The guarantees we require from HKDF in this application don't require this, but I would suggest for completeness that the info field be set using "|".join(["0"] + addr_strings.sort()).

info="0|0x744f31bA5e8F6bAE5287fCc5c34476A2267bCaA2|0xc86aE9599A2bB5a99b19D2b8AD1BF89082023d63"

The Prefixed sequence number will allow derivation of future iterations if needed, while the Additions on the addresses adds entropy to the context.

Salt constant looks great.

jazzz (approved):

test/keystore/InMemoryKeystore.test.ts
```diff
@@ -472,6 +475,39 @@ describe('InMemoryKeystore', () => {
...
+            createdNs: dateToNs(createdAt),
+            context: undefined,
+          })
+        })
+      )
+
+      const convos = await aliceKeystore.getV2Conversations()
+
+      // Ensure all the convos have the same topic
+      expect(
```
michaelx11, line 504: I think `convos` above might return just one conversation (deduplicated off topic). A more simplistic/readable test case could be something like:
```
      // Go through CreateInviteResponses and ensure they all have the same topic
      const firstResponse = responses[0]
      const topicName = firstResponse.conversation!.topic

      expect(
        responses.filter((value, index, array) => {
          return value.conversation!.topic === topicName
        })
      ).toHaveLength(25)
```

michaelx11: I tried to think of a clean way to test `topicKey` is the same, but it's harder since the keyMaterial is not surfaced in the response

src/keystore/InMemoryKeystore.ts
```diff
@@ -24,13 +24,21 @@ import {
...
 } from './utils'
-import { nsToDate } from '../utils'
+import { nsToDate, buildDirectMessageTopicV2 } from '../utils'
 import InviteStore from './InviteStore'
 import { Persistence } from './persistence'
 import LocalAuthenticator from '../authn/LocalAuthenticator'
 import { AccountLinkedRole } from '../crypto/Signature'
+import { sha256 } from 'ethers/lib/utils'
+import { hmacSha256Sign } from '../crypto/ecies'
+import { getRandomValues } from '../crypto/utils'
```
michaelx11, line 34: nit: warnings about unused imports

michaelx11 (commented):
Looking real good, a couple comments around test and unused imports

src/keystore/InMemoryKeystore.ts
nakajima: Fixed in https://github.com/xmtp/xmtp-js/pull/326/commits/0f091a41b5acc20f15e0c42edb34a21a7e7a1fa2.

src/keystore/InMemoryKeystore.ts
```diff
@@ -272,9 +294,46 @@ export default class InMemoryKeystore implements Keystore {
...
+        this.accountAddress,
+        await recipient.walletSignatureAddress(),
+      ].sort()
+
+      const msgString =
+        (req.context?.conversationId || '') + sortedAddresses.join()
+
+      const msgBytes = new TextEncoder().encode(msgString)
+
+      const topic = sha256(
```
michaelx11, line 316: Just checking, this does a redundant `sha256` of the post `hmacSha256Sign` output, which is already run through `sha256`. Did we spec that out or otherwise can we remove it? Not a huge deal either way TBH but don't want non-spec operations sneaking in if we want to reproduce this on other platforms.

HMAC construction being: `HASH(Key XOR opad, HASH(Key XOR ipad, text))`

More details: https://security.stackexchange.com/a/79579

michaelx11 (commented):
Looks good! Just one final comment about whether we're doing a redundant `sha256` operation

src/keystore/InMemoryKeystore.ts
nakajima: Fixed in https://github.com/xmtp/xmtp-js/pull/326/commits/19a4f65de3a0e0b9d0695560a2f5d721b3600a91!

michaelx11 (approved):
Great work!

nakajima:
Just a heads up that I'm switching focus to react native for a bit but will return to this whenever that's in a better spot. I'm going to leave it open for now. Thank you all so much for your patience with the reviews here, it's been super illuminating.

src/keystore/InMemoryKeystore.ts
```diff
@@ -23,12 +23,33 @@ import {
...
+import { nsToDate, buildDirectMessageTopicV2 } from '../utils'
 import InviteStore from './InviteStore'
 import { Persistence } from './persistence'
 import LocalAuthenticator from '../authn/LocalAuthenticator'
+import { hmacSha256Sign } from '../crypto/ecies'
+import crypto from '../crypto/crypto'
 const { ErrorCode } = keystore
 
+// Constant 32 byte salt
+const inviteSalt = new TextEncoder().encode('This is my salt for your invite.')
```
neekolas, line 35: We probably want to change this to something that feels more trustworthy. Can be literally any string.

neekolas: `______________xmtp______________`?

neekolas: Long as it adds up to 32 bytes and looks reasonable, I'm down

rygine: what do you think of this? [fix: update invite salt](https://github.com/xmtp/xmtp-js/pull/326/commits/bbed5fca1fc940614069312aa2ab75c09b4883a8)

neekolas: Works for me

neekolas (approved):

Issue #275, opened 2023-02-15
Bug: Double signature is buggy on mobile wallets on iOS with walletconnect


nmalzieu:
### Describe the bug

On mobile we're using WalletConnect and WalletConnect links to mobile wallets using [Universal Links](https://developer.apple.com/ios/universal-links/) 

After discussing with WalletConnect, Universal Links only work when triggered from a **user action** (for instance clicking on a button)

When a user logs in to XMTP for the first time, the XMTP SDK asks for 2 signatures in a row : "Create Identity" & "Enable Identity". Since the second signature is not triggered by a user action but automatically, it often fails to open the mobile wallet.

I think we should have a way (probably a setting in `Client.create` or auto-detect that we are on mobile) to split those two signatures.

If the user logs in for the first time, we don't want to auto trigger a second signature but have a button to do the second signature.

Screenshot of our discussion with WalletConnect:

<img width="530" alt="Capture d’écran 2023-02-15 à 17 36 45" src="https://user-images.githubusercontent.com/2102342/219075259-c19c0630-65f1-425c-9add-38b3e6c0201f.png">

galligan:
@nmalzieu thanks for this. Am I reading this right that with WalletConnect if the second signature were initiated with a user action, this wouldn't be a problem?

nmalzieu:
@galligan exactly yes.
Apple Universal links are https links that are supposed to open the native app if it's installed. But they only do so if the link has been accessed from a recent user action

WalletConnect actually had some information about this on their website : https://docs.walletconnect.com/1.0/mobile-linking#ios-app-link-constraints

nmalzieu:
It might work sometimes but to be sure it works 100% of the time, I think splitting the 2 signatures during a signup is the way to go.

nplasterer:
I think this is related to https://github.com/xmtp/xmtp-js/issues/264.

nmalzieu:
> I think this is related to https://github.com/xmtp/xmtp-js/issues/264.

Totally! But wanted to first focus on the specific case of dual signature and the obvious fix: split them :)

neekolas:
So, because we have one JS SDK to support web, Node.js, and mobile webviews, we'd have to be careful about rolling out a change like this. It's definitely possible, but would require some thorough testing to make sure that we don't break any of the headless flows/existing apps. We would need to default to the current flow and then add an option to make it a two-phase operation.

[...]

Issue #264, opened 2023-01-24
Mobile wallets are either impossible to use or not reliable


polmaire:
We are using a webview within the Converse iOS app to connect & sign and access the XMTP network.

We encounter a lot of difficulties:

- it seems like mobile wallets do not accept two signatures in a row. The consequence is very simple: it's IMPOSSIBLE to create an XMTP account on mobile. Only people who already connected to the network on desktop can access. At least the most important wallets (let's say Metamask or Rainbow) have to change asap, if not it's a very big problem for XMTP.
- for people who already connected to the XMTP network, they generally need to try to connect and / or sign twice or three times in order to make it work
- sometimes the walletconnect popup does not show up when people try to click on "Connect" in the Converse app
- when it finally works, the deeplink to go back to the app once you have signed generally does not work, you have to come back manually

We'll be happy to provide more information if needed. It seems like both Walletconnect and wallets are having trouble doing ... the only thing that we expect from them !

PR #331, opened 2023-04-11
Pre Signature Notifications

neekolas:
## Summary

This attempts to solve two issues 

1. Developers would like to know when we are prompting for signatures, so they can update their UI to show some explanatory text
2. WebViews on mobile have some hard limitations around multiple deep-links without user interaction (see below) that make it impossible to get both the `Create Identity` and `Enable Identity` signatures without a user interaction in between.

This feature allows developers to specify two callback functions that are invoked immediately before requesting either of these signatures. The callback functions are expected to return a Promise, which will be awaited in our code. This can allow developers to force some sort of user interaction before resolving the promise if desired, or to resolve immediately but update the UI in their code.

## Issues
- https://github.com/xmtp/xmtp-js/issues/275
- https://github.com/xmtp/xmtp-js/issues/264

## Notes
I was really unsure what to call this. I was torn between `callback`, `hook`, and `notification`. I settled on `notification` because the callback function doesn't get any arguments and has limited control over what happens next (although it can abort the whole process by throwing an error or rejecting the promise).

Open to changing if others feel strongly.

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -63,6 +63,7 @@ export type SendOptions = {
 }
 
 export type XmtpEnv = keyof typeof ApiUrls
+export type PreEventNotifier = () => Promise<void>
 
 /**
  * Network startup options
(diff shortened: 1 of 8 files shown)


michaelx11 (approved):
This seems like a minimally invasive solution. LGTM!

It makes me wonder how we would structure this for libxmtp. Maybe require functions to be called to advance some "state" before the client is usable?

README.md
```diff
@@ -137,16 +137,18 @@ const xmtp = await Client.create(wallet)
...
+| appVersion                | `undefined`                                                                       | Add a client app version identifier that's included with API requests. For example, you can use the following format: `appVersion: APP_NAME + '/' + APP_VERSION`. Setting this value provides telemetry that shows which apps are using the XMTP client SDK. This information can help XMTP developers provide app support, especially around communicating important SDK updates, including deprecations and required upgrades. |
+| preCreateIdentityCallback | `undefined`                                                                       | `preCreateIdentityCallback` is a function that will be called immediately before a "Create Identity" wallet signature is requested from the user.                                                                                                                                                                                                                                                                                |
```
michaelx11, line 150: It's almost like this `callback` is used to "allow XMTP to pop a modal". Like a "showCreateIdentitySignatureModalCallback" or are the semantics slightly different?

neekolas: Opening a modal is one possible use of this, but I want to be agnostic to how the developer might use the feature. 

Maybe it's a toast message, maybe it's an analytics metric

michaelx11: I see, I was incorrectly conceptualizing as developer calls `callback` which initiates `xmtp-js` wallet signing modal popup. But the more accurate formulation is `xmtp-js` initiates by calling the `callback`, providing the developer a chance to do whatever before `xmtp-js` prompts for signature.

neekolas: I should really include an example in the README of how to use this.

Going to do another pass at documentation before release tomorrow and will include in that.

PR #332, opened 2023-04-11
feat: implement outgoing 1xN fanout and multibundle support in general

michaelx11:
## Overview

After this PR, each `VoodooClient` will publish their own contact bundle, even if an existing one for the address is present in the contact topic. When messaging another user, the client downloads all contact bundles, aggregates them into a `VoodooMultiBundle` and then messages all of them.

We also perform `MultiSession` management in `VoodooConversation`. This PR does _not_ handle self-fanout of outgoing messages. That will be done next.

**Note:** Ideally we'd do more of this bookkeeping in Rust, but after we get NxN working in xmtp-js we can decide how much logic to shift back into Rust land.

## Changes
- Add `src/voodoo/types.ts` file with all types.
- Add new types for `OneToOneSession` (device to device session), `VoodooMultiBundle` (list of contacts) and `VoodooMultiSession` (list of sessions for user<>user convo)
- Implement multibundle support throughout
- Implement a unit test for 1xN fanout and also Nx1, but there isn't NxN (no self-messaging)

## Test Plan

- Unit tests

### DIFF
--- a/src/voodoo/VoodooClient.ts
+++ b/src/voodoo/VoodooClient.ts
@@ -57,7 +57,7 @@ export default class VoodooClient {
     apiClient: ApiClient
   ): Promise<VoodooClient> {
     const xmtpWasm = await xmtpv3.XMTPWasm.initialize()
-    // TODO: STARTINGTASK: this just creates unused voodoo keys that go nowhere
+    // TODO: STARTINGTASK: this just creates unused voodoo keys that hang out in memory ephemerally
     const myVoodoo = xmtpWasm.newVoodooInstance()
     const client = new VoodooClient(address, myVoodoo, apiClient, xmtpWasm)
     await client.ensureUserContactPublished()
@@ -71,17 +71,11 @@ export default class VoodooClient {
     }
   }
 
-  // Create new Voodoo invite
-  async newVoodooInvite(
-    contactAddress: string,
+  // Create a new Voodoo invite (vmac outbound session with plaintext = topic) for a contact
+  async newVoodooInviteForContact(
+    contactInstance: VoodooContact,
     topic: string
   ): Promise<EncryptedVoodooMessage> {
-    // Get the contact info which is just a handle for now
-    const contactInstance = await this.getUserContactFromNetwork(contactAddress)
-    if (!contactInstance) {
-      throw new Error(`No contact info for ${contactAddress}`)
-    }
-
     const outboundSessionResult: SessionResult =
       await this.voodooInstance.createOutboundSession(
         contactInstance.voodooInstance,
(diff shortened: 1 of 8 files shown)


test/voodoo/conversations/VoodooConversation.test.ts
```diff
@@ -86,5 +86,65 @@ describe('conversation', () => {
         expect(ams2[i].senderAddress).toBe(bms2[i].senderAddress)
       }
     })
+
+    it('1 to N fanout', async () => {
```
michaelx11, line 90: This is the good stuff

src/voodoo/conversations/VoodooConversations.ts
```diff
@@ -46,37 +80,52 @@ export default class VoodooConversations {
   }
 
   // Must be locked by convoMutex
+  // TODO: this method currently looks up the other party's multibundle for NxN fanout
+  // we may want to carry this information in the invite itself otherwise there could be
+  // a mismatch on the initial set of devices and the set we resolve in this method
   async processInvite(
```
michaelx11, line 86: This function is currently incomplete but will addressed in the next PR. New logic will need to be:
- On invite, check if a new session needs to be added to an existing conversation
- If brand new conversation, do a `multibundle` lookup and initiate OneToOneSession's with each of those instances
- If new self-conversation, initiate `multibundle` lookup for both senderAddress and self, then initiate OneToOneSessions with each of those entities if not already in the MultiSession object of the VoodooConversation

src/voodoo/VoodooClient.ts
```diff
@@ -215,8 +194,9 @@ export default class VoodooClient {
 
   private async ensureUserContactPublished(): Promise<void> {
     const bundle = await this.getUserContactFromNetwork(this.address)
-    // TODO: validate the bundle, not just check for presence
-    if (bundle) {
+    // NOTE: other devices for this wallet could have published bundles, this is expected
+    // TODO: we only avoid republishing our own bundle if the account is the same
+    if (!!bundle && bundle.voodooInstance === this.voodooInstance) {
```
neekolas, line 199: This equality check is only validating whether the two instances point to the same memory address, not a deep equality of whether they are equivalent. 

So, it will never be true.

michaelx11: Ah good catch

michaelx11: I'll fix in the follow up PR

src/voodoo/VoodooClient.ts
```diff
@@ -256,6 +236,40 @@ export default class VoodooClient {
...
+
+      // TODO: do validation here of the address signature
+      const voodooInstance =
+        this.wasm.addOrGetPublicAccountFromJSON(voodooPublicJson)
+      listContacts.push(new VoodooContact(peerAddress, voodooInstance))
+    }
+    return {
+      address: peerAddress,
+      contacts: listContacts,
+      timestamp: new Date().getTime(),
```
neekolas, line 269: Do we actually need a timestamp here? Given that the multi-bundle spans a time range, doesn't seem particularly useful (even if we used the timestamp of the first or last contact).

src/voodoo/conversations/VoodooConversation.ts
```diff
@@ -65,30 +77,62 @@ export default class VoodooConversation {
...
+    messages.sort((a: VoodooMessage, b: VoodooMessage) => {
       return a.timestamp - b.timestamp
     })
-    return this._messages
+    this.multiSession.messages.set(sessionId, messages)
+    return messages
+  }
+
+  // Provides an aggregated and sorted list of messages for all of the device sessions
+  async messages(): Promise<VoodooMessage[]> {
```
neekolas, line 99: Having messages spread across multiple topics is going to make pagination much more annoying in the short term (given that we can't do it in a single query anymore). Once we have PDS, it probably matters less given that our goal is always to fetch everything new and put it into PDS.

michaelx11: Agreed - it's something we'll want to experiment with and learn about in this prototyping phase!

PR #333, opened 2023-04-12
feat: full NxN prototype in xmtp-js

michaelx11:
## Overview

The code isn't ready for prime-time (aka PR yet) but by looking at the test case you can see that NxN fanout is happening.

Definitions:
- VoodooConversation is now a container between two participating addresses (no conversationId concept yet). It contains multiple sessions and associated contacts (VoodooContact aka Vodozemac Accounts). There can even be duplicated contacts if two devices send invites to each other simultaneously.
- VoodooContact aka Vodozemac account aka device. This is a Vodozemac identity key pair and contact bundle that is generated by VoodooClient upon creation. There can be multiple of these per wallet address, and between two VoodooContacts there can even be multiple sessions (Vodozemac messaging sessions) for the same conversation (two invites simultaneously from both directions).

There are a lot of gotchas and sharp edges to this process:
1) Invites to all peers are only sent out when `send` is invoked. At that point, a census is taken of current open sessions for that conversation and it is compared with what was in the contact bundles before
2) On `VoodooConversations.list()` the invite topic is re-scanned and any new invites to existing Conversations are populated aka `VoodooConversation.multiSession.sessionIds` etc is updated
3) On `conversation.messages()` all messages from all sessions are aggregated. There should not be duplicates, even if we're polling from multiple sessions between the same two devices. This is because on the send side, we only ever send to one device per VoodooContact.
4) This whole system relies upon fallback keys being used. Otherwise, multiple invites to a single VoodooContact cause decryption failures etc

## Test Plan
- See the very involved NxN unit test

### DIFF
--- a/src/voodoo/conversations/VoodooConversation.ts
+++ b/src/voodoo/conversations/VoodooConversation.ts
@@ -3,6 +3,13 @@ import { VoodooMessage, VoodooMultiBundle, VoodooMultiSession } from '../types'
 import Stream from '../../Stream'
 import { messageApi } from '@xmtp/proto'
 
+/**
+ * This class represents a conversation between two users.
+ * It needs to manage various 1:1 messaging sessions between devices that
+ * comprise this conversation.
+ * - 1:1s between my current VoodooInstance and my other ones (on other devices possibly)
+ * - 1:1s between my current VoodooInstance and my peer's VoodooInstances
+ */
 export default class VoodooConversation {
   peerAddress: string
   createdAt: number
@@ -13,7 +20,8 @@ export default class VoodooConversation {
     client: VoodooClient,
     peerAddress: string,
     createdAt: number,
-    multibundle: VoodooMultiBundle,
+    myMultiBundle: VoodooMultiBundle,
+    otherMultiBundle: VoodooMultiBundle,
     sessionIds: string[],
     topics: string[]
   ) {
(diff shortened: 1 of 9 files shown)


test/voodoo/conversations/VoodooConversation.test.ts
```diff
@@ -86,5 +86,209 @@ describe('conversation', () => {
...
+        const bcs = await b.conversations.list()
+        expect(bcs).toHaveLength(1)
+        const bc = bcs[0]
+        const bms = await bc.messages()
+        expect(bms).toHaveLength(1)
+        expect(bms[0].plaintext).toBe('hi')
+      }
+    })
+
+    it('test nxn fanout', async () => {
```
michaelx11, line 191: Now this is the full NxN fanout test

src/voodoo/types.ts
```diff
@@ -43,15 +57,14 @@ export type VoodooMultiBundle = {
   address: string
   // The bundles for each device
   contacts: VoodooContact[]
-  // Last refreshed timestamp
-  timestamp: number
```
michaelx11, line 47: @neekolas - removed the timestamp per your feedback on other PR

src/voodoo/conversations/VoodooConversation.ts
```diff
@@ -168,4 +249,116 @@ export default class VoodooConversation {
...
+
+  /**
+   * Creates a new single OneToOneSession for the given contact. Does not publish the invite envelope.
+   *
+   * NOTE: we have to very explicit about what this session is. It's a session between two VoodooInstances,
+   * but the two addresses participating could be the same. So otherAddress is only used as the
+   * payload of the invite, but the invite is encrypted for contact and delivered to the contact.address
+   *
+   * Currently, we generate a random topic for the session, and send an Olm PreKey Message
+   * where the encrypted ciphertext is just the random topic. This is the "invite" message.
```
jazzz, line 261: Does this topic serve any purpose other than informing the other session where to listen? Can this topic be dynamic?

src/voodoo/types.ts
```diff
@@ -24,6 +29,15 @@ export type EncryptedVoodooMessage = {
   ciphertext: string
 }
 
+// The object to be JSON-serialized to form the body(plaintext) of the invite
+export type VoodooInvite = {
+  topic: string
+  // Addresses of the two wallets in the conversation
+  // not necessarily the two voodoo instances in a single 1:1 session
+  // e.g. Me1 sending my messages in convo to Me2
+  participantAddresses: string[]
```
jazzz, line 38: Slightly confused by this comment

In the case of Me1 -> Me2 would the participant addresses would read
['0x000ME','0x000YOU'] correct? not ['0x000ME','0x000ME']

michaelx11: Yup, that is correct. It would be ['0x000ME', '0x000YOU']. Let me reread the comment and reword if necessary

src/voodoo/conversations/VoodooConversation.ts
michaelx11: Nope and yes it can be dynamic in the sense we can constantly spawn more sessions between two Vodozemac accounts. The current code setup actually handles that nicely

michaelx11:
FYI I'm going to merge this in a few hours to cap off our two-week prototyping sprint. Thanks for playing everyone!

PR #334, opened 2023-04-12
fix: add `clientAddress` to `Conversation` type

rygine:
The `Conversation` interface does not include the `clientAddress` property, which is present in both `ConversationV1` and `ConversationV2`.


neekolas (commented):
Thanks for adding back in. I didn't realize it was actually being used by anyone anymore.

neekolas (approved):

PR #337, opened 2023-04-14
fix: ensure correct envelope is passed to prepared message

nmalzieu:
This was fixed in v7 but somehow got lost in v8


nakajima (approved):

neekolas (approved):
Thanks for catching this regression.

PR #338, opened 2023-04-19
Upgrade to node 18, refactor crypto imports

rygine:
This PR updates the version requirement of node to v18.

* Removed streams polyfill (no longer necessary)
* Centralized all crypto API imports to account for node and browser APIs
* Added ESLint rule to prevent importing from `crypto` (or `node:crypto`) directly
* Updated webpack config to remove unnecessary options
* Changed build `target` to be more aligned with node 18
* Upgraded `jest` and friends to v28

These changes are meant to modernize the code base and set us up for refactoring the build process. The main goal of the new build process is to optimize the bundle we ship for the browser.


src/crypto/crypto.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * This file is necessary to ensure that the crypto library is available
+ * in node and the browser
+ */
+
+// eslint-disable-next-line no-restricted-syntax
+import { webcrypto as nodeCrypto } from 'crypto'
```
neekolas, line 7: You've tested this in a browser to confirm this won't blow up?

rygine: i have. created a build and everything seemed to work in the Inbox app.

neekolas: Great stuff.

test/Client.test.ts
```diff
@@ -162,10 +162,10 @@ describe('encodeContent', () => {
...
       10, 18, 10, 8, 120, 109, 116, 112, 46, 111, 114, 103, 18, 4, 116, 101,
       120, 116, 24, 1, 18, 17, 10, 8, 101, 110, 99, 111, 100, 105, 110, 103, 18,
-      5, 85, 84, 70, 45, 56, 40, 0, 34, 45, 120, 1, 51, 52, 48, 209, 49, 52, 48,
-      4, 98, 11, 8, 54, 52, 212, 49, 54, 2, 82, 150, 96, 166, 161, 161, 9, 84,
-      202, 0, 44, 60, 170, 122, 84, 245, 168, 106, 218, 171, 6, 0, 139, 43, 173,
-      229,
+      5, 85, 84, 70, 45, 56, 40, 0, 34, 45, 120, 156, 51, 52, 48, 209, 49, 52,
+      48, 4, 98, 11, 8, 54, 52, 212, 49, 54, 2, 82, 150, 96, 166, 161, 161, 9,
+      84, 202, 0, 44, 60, 170, 122, 84, 245, 168, 106, 218, 171, 6, 0, 139, 43,
+      173, 229,
```
rygine, line 168: @neekolas this is one of the fixtures that i was referring to in our conversation

src/index.ts
```diff
@@ -11,7 +11,6 @@ export {
   encrypt,
   decrypt,
 } from './crypto'
-export { getRandomValues } from './crypto/utils'
```
rygine, line 14: @nakajima i looked for this being imported from the content types [repo](https://github.com/xmtp/xmtp-js-content-types), but couldn't find anything.

test/Client.test.ts
neekolas: Long as the output of the new version remains decompressible by the old, we should be safe.

rygine:
@neekolas i'm not sure how to version this update, or what semantic release will version it as. i was thinking it might be a feature release, but then again, is it considered a breaking change if the underlying node requirement changes?

test/ApiClient.test.ts
```diff
@@ -7,7 +7,7 @@ import { messageApi } from '@xmtp/proto'
 import { sleep } from './helpers'
 import { LocalAuthenticator } from '../src/authn'
 import { PrivateKey } from '../src'
-import { version } from '../package.json'
+import packageJson from '../package.json'
```
neekolas, line 10: Curious why this needed to change

neekolas (approved):
This looks good to me.

Can't wait to get rid of Webpack and start having a tree-shakeable bundle.

neekolas:
> i'm not sure how to version this update, or what semantic release will version it as. i was thinking it might be a feature release, but then again, is it considered a breaking change if the underlying node requirement changes?

I feel alright calling it a feature release, given that we aren't breaking any API contracts.

src/index.ts
neekolas: I guess technically this is a breaking change. Even though I highly doubt people are widely importing this, we might have to bring it back to make this a non-major release. We can mark it as deprecated and remove it altogether in v9.

test/ApiClient.test.ts
rygine: certain transpilers were giving a warning about using non-default imports of a json file. it still technically works, but just wanted to remove any possible issues, even though they aren't likely.

more info on why here: https://github.com/tc39/proposal-json-modules#why-dont-json-modules-support-named-exports

rygine:
> I feel alright calling it a feature release, given that we aren't breaking any API contracts.

pardon my ignorance on semantic commits, but i'm not sure what version `semantic-release` will determine based on the commits. do i have to add a `feat: ...` commit?

PR #367, opened 2023-04-20
GRPC Error Handling

neekolas:
## Summary

This is a long overdue fix to actually throw publish errors, after the auth token has been refreshed and all the retries have been exceeded.

The error swallowing is a hold-over from the bad-old-days of client side LibP2P where the error reporting was unreliable and publishing was fire and forget.

I have updated our ApiClient to wrap the errors thrown from the generated code in actual errors (they were previously plain objects) to make testing and type safe error handling easier. Async code should always reject with an actual Error.


test/Client.test.ts
```diff
@@ -266,6 +264,43 @@ describe('canMessageMultipleBatches', () => {
...
+    const envelope = {
+      contentTopic: buildUserContactTopic(c.address),
+      message: new TextEncoder().encode('hello world'),
+      timestamp: new Date(),
+    }
+
+    // `expect(promise).rejects.toThrow` fails but this works.
+    // Would love to understand why
+    try {
+      await c.publishEnvelopes([envelope])
```
rygine, line 296: what about something like this?

```
expect(await c.publishEnvelopes([envelope])).rejects.toThrow(...));
```

i'm looking at this: https://jestjs.io/docs/expect#rejects

neekolas: In what you have above the `await` will cause the entire test function to abort when there is an error. The await is evaluated first, so the `expect` can't take over and catch the error.

I tried a bunch of different variations, but keep getting `Received function did not throw` test failures.

For example, this looks like something that would work (and does work correctly in many other tests)
```ts
await expect(c.publishEnvelopes([envelope])).rejects.toThrow()
```

neekolas: I figured it out...

The generated GRPC code doesn't throw actual JS errors. It just throws objects. Jest doesn't recognize that. If I wrap the error with `throw new Error(e)` in the publish code Jest picks up on it and the boring example I have above works fine.

michaelx11 (approved):
lgtm

src/ApiClient.ts
```diff
@@ -125,22 +136,27 @@ export default class ApiClient {
   }
 
   // Raw method for querying the API
-  private _query(
+  private async _query(
```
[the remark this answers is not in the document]
neekolas: You got it

PR #368, opened 2023-04-21
build: update build process

rygine:
This PR overhauls the build process to generate single file bundles. This provides more compatibility with frontend bundlers such as vite and webpack.

**BREAKING CHANGE:** The NPM package no longer contains a mirror of the `/src` directory. Therefore, any consumers of this package that rely on importing from the `/dist` folder will break.

After these changes have merged, all externally available exports must be explicitly defined in `src/index.ts`.

Some of the benefits of this approach include:

- Strict control on what's exported from the SDK
- Extremely fast build process using [tsup](https://tsup.egoist.dev/), ~1.5 seconds for all 3 bundles versus ~16 seconds before (about 10X faster!)
- Smaller overall package size, 391kB zipped, 1.8MB unpacked versus 791kB zipped, 3.8MB unpacked (about half the size!)
- Much smaller web bundle, 95kB versus 962kB (about 10X smaller!)

Other notable changes in this PR:

- The package is now defined as a module, which means that CommonJS files must have a `.cjs` extension so that Node can process them properly
- Add `exports` keys to `package.json` to increase compatibility with other bundlers (more info [here](https://nodejs.org/api/packages.html#package-entry-points))
- Added more exports in `src/index.ts` to account for some consumers that need to be updated (see [here](https://github.com/search?q=org%3Axmtp%20%40xmtp%2Fxmtp-js%2Fdist&type=code))

These changes have been tested using the React SDK in Vite and Next.js (webpack).

With this change, it looks like we need to update [this doc](https://github.com/xmtp/example-notification-server-go/blob/b53aa01375b90831a95baee25c638b9dad38a627/docs/notifications-client-guide.md?plain=1#L95) and [this doc](https://github.com/search?q=org%3Axmtp-labs+%40xmtp%2Fxmtp-js%2Fdist&type=code). cc @jhaaaa 

@neekolas I think this update should be classified as a breaking change, so v9 would be the new version. WDYT?


neekolas:
> @neekolas I think this update should be classified as a breaking change, so v9 would be the new version. WDYT?

Does feel like a breaking change. I've seen external developers import directly from the `dist` folder and don't want to surprise them. I'm totally OK with going straight to v9. 

To bump the major version you just need a commit that includes a footer using the token `BREAKING CHANGE: `

tools/cleanup.js
rygine, line 1: is there a reason we have to remove files this way as opposed to `rm -rf` in NPM scripts?

neekolas:
@rygine I think this might warrant a merge to the beta branch, so that we can more easily test it out across different apps. Would be good to know if there are any cases where you'd need to update your own app's build config in advance of release.

I'd bet Converse would give it try. I can test it out with Lenster.

tsup.config.ts
```diff
@@ -0,0 +1,35 @@
...
+export default defineConfig((options) => {
+  const esbuildPlugins: Plugin[] = []
+
+  if (options.platform === 'browser') {
+    esbuildPlugins.push(
+      externalGlobal.externalGlobalPlugin({
+        crypto: '{ webcrypto: window.crypto }',
+      })
+    )
+  }
```
rygine, line 14: i need to add comment here about what this is doing. it's essentially replacing any imports of `crypto` with an object.

so this
```
import { webcrypto } from "crypto";
```
essentially becomes this
```
const { webcrypto } = { webcrypto: window.crypto };
```

tools/cleanup.js
neekolas: I suggested removing these scripts a while back. The pushback was that this is guaranteed to work nicely cross-platform, so it'll "just work" on Windows etc.

rygine:
> > @neekolas I think this update should be classified as a breaking change, so v9 would be the new version. WDYT?
> 
> Does feel like a breaking change. I've seen external developers import directly from the `dist` folder and don't want to surprise them. I'm totally OK with going straight to v9.
> 
> To bump the major version you just need a commit that includes a footer using the token `BREAKING CHANGE: `

Yep, this is already present in the commit message. 😅

rygine:
> 

Definitely agree on this. There's a documentation update too, which I can do in another commit on this PR. With the new web bundle, there's _some_ configuration necessary to deal with necessary polyfills.

tools/cleanup.js
rygine: gotcha, we could also use `rimraf`, which should also cover windows.

neekolas: No objections there

neekolas:
Overall, this is a terrific and much-needed change to our build system. Between the build times and the smaller bundle, this should make a big difference to developers.

Will approve if you change the PR target to `beta` (which is now up-to-date with `main`)

rygine:
> Overall, this is a terrific and much-needed change to our build system. Between the build times and the smaller bundle, this should make a big difference to developers.
> 
> Will approve if you change the PR target to `beta` (which is now up-to-date with `main`)

Done!

neekolas (approved):

tools/cleanup.js
daria-github: I know this is an outdated comment so I may just be missing context, but want to +1 using `rimraf` if this is still an issue on Windows -- any reason we're not?

rygine: added `rimraf` on [this commit](https://github.com/xmtp/xmtp-js/pull/368/commits/360c5294f3f09056b3924f9d6d0b4639cebff564)

Issue #374, opened 2023-05-01
Bug: Invalid signature


elisealix22:
### Describe the bug

Reported in Discord:
>when I try to use XMTP as a user (no matter on which app), I will always get an "Invalid Signature" error in the console. Can anyone help me understand why this is happening? Or is there anyway to reset my xmtp identity?

>1. the address: 0xbDcb6419CcB21cEe2412cCed0ac4984b6A290bF0

>2. I created this address to transfer the lens profile to a rainbow wallet. But shortly after the transfer the DMs stopped working. I was able to use DMs 2-3 times though before it stopped working.

>3. I am basically experiencing this on any app that uses XMTP, These are the apps I tried: Orb, Butterfly and Converse (the messaging app) 

We believe we found the root cause of the invalid signature being created in our Flutter SDK, however we should make the JS SDK more friendly to use those signatures so people can continue to use XMTP.

Related PR: https://github.com/xmtp/xmtp-flutter/pull/67

### Expected behavior

Allow v1 contact bundles to use either `ecdsaCompact` or `walletEcdsaCompact` signatures. The bug is that we have v1 contact bundles using `walletEcdsaSignature`s on the network and we only accept `ecdsaCompact` currently.

### Steps to reproduce the bug

1. Create a contact with a v1 bundle that has a `walletEcdsaCompact` signature (previously not allowed).
2. Ensure that contact's address can create a `Client` and use the `xmtp-js` library as expected.

elisealix22:
@yash-luna yes, I can confirm they are in the same state in our `xmtp-debug` tool. Working on a fix right now & this is my highest priority.

<img width="627" alt="Screenshot 2023-05-05 at 10 37 36 AM" src="https://user-images.githubusercontent.com/556051/236488954-dfbac026-ab03-450f-95a8-e4b5131603dc.png">

PR #376, opened 2023-05-05
fix: handle `walletEcdsaCompact` signatures for v1 keys

elisealix22:
Follow-up to https://github.com/xmtp/xmtp-flutter/pull/67.

Resolves https://github.com/xmtp/xmtp-js/issues/374.

This PR updates our SDK to accept PublicKey signatures that only have a `walletEcdsaCompact` signature. I put a fix in the `PublicKey` constructor so that we fix the signature there and then I also added a fix where we were throwing the invalid signature in case a new PublicKey isn't always re-created. I tried to add tests to recreate this state as well.

### DIFF
--- a/src/crypto/PublicKey.ts
+++ b/src/crypto/PublicKey.ts
@@ -225,7 +225,18 @@ export class PublicKey
       secp256k1Uncompressed: obj.secp256k1Uncompressed,
     })
     if (obj.signature) {
-      this.signature = new Signature(obj.signature)
+      // Handle a case where Flutter was publishing signatures with walletEcdsaCompact
+      // instead of ecdsaCompact for v1 keys.
+      if (!obj.signature.ecdsaCompact && obj.signature.walletEcdsaCompact) {
+        this.signature = new Signature({
+          ecdsaCompact: {
+            bytes: obj.signature.walletEcdsaCompact.bytes,
+            recovery: obj.signature.walletEcdsaCompact.recovery,
+          },
+        })
+      } else {
+        this.signature = new Signature(obj.signature)
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/crypto/PublicKey.ts
```diff
@@ -225,7 +225,18 @@ export class PublicKey
       secp256k1Uncompressed: obj.secp256k1Uncompressed,
     })
     if (obj.signature) {
-      this.signature = new Signature(obj.signature)
+      // Handle a case where Flutter was publishing signatures with walletEcdsaCompact
+      // instead of ecdsaCompact for v1 keys.
+      if (!obj.signature.ecdsaCompact && obj.signature.walletEcdsaCompact) {
```
mkobetic, line 230: Do we need to do this if we expand the getPublicKey functionality below?

michaelx11: It's a good question! It feels like a good move to fix the signature here just in case it propagates elsewhere and ends up serialized, but analyzing that seems tricky.

michaelx11 (approved):
LGTM - do we need a test wallet for a real world test? I can generate one.

src/crypto/PublicKey.ts
elisealix22: Yeah, I was worried about the signature assumption being made elsewhere (even in the future) and thought the best fix might be fixing the PublicKey object itself. It's a best attempt at least imo.

elisealix22:
@michaelx11 💚 that would be a huge help!!! I've just been regression testing -- would be great to test this actually resolves the issue outside of tests.

src/crypto/PublicKey.ts
elisealix22: I feel safer leaving the check in both places because I can't guarantee the new constructor will be called in all cases. I know it's a little messier, but I confirmed this does solve the issue at hand with a test wallet. I think I'm going to merge as-is unless you strongly disagree!

PR #381, opened 2023-05-22
feat: begin adding some group conversation features

nakajima:
### Creating a group

The **creator** of the group creates invites using `keystore.createInvites` which creates N invites with the same topic/keys. These invites must also have the following metadata in their context:

- `groupID` a random string identifier for the group
- `initialMembers` a comma separated list of addresses for the initial members of the group

These invites are sent to the initial members who can identify the conversation as a group chat from those properties being present.

#### UI Considerations

Most apps currently treat peer address (or the sender of an invitation) to be the identity of a conversation. With group messages they'll want to use the `groupID` or a user defined group title (covered further below).

### Sending/Receiving messages

Messages sent to the conversation's topic will be able to be received like any other messages on the network. The only difference is that the number of people sending/receiving is more than 2.


src/conversations/Conversations.ts
```diff
@@ -429,6 +439,67 @@ export default class Conversations {
     return seenPeers
   }
 
+  async newGroupConversation(
```
neekolas, line 442: Are we sure we need this in `xmtp-js`. I was kinda imagining this would all live in `xmtp-group-chat`. Are we relying on some private methods or something?

src/keystore/InMemoryKeystore.ts
```diff
@@ -251,28 +254,139 @@ export default class InMemoryKeystore implements Keystore {
...
+        )
       }
-      await this.inviteStore.add([topicData])
 
-      return keystore.CreateInviteResponse.fromPartial({
-        conversation: topicDataToConversationReference(topicData),
-        payload: sealed.toBytes(),
+      // Register an address on the network for the group
+      const sharedWallet = Wallet.createRandom()
+      const sharedClient = await Client.create(sharedWallet)
```
neekolas, line 283: I thought we were getting rid of the shared client with these changes?

nakajima: This is just to have an address for the group to use as a peer address. I'm creating a client here so that the contact bundle exists on the network.

src/conversations/Conversations.ts
nakajima: I've been thinking group chat should live here since it's such a core messaging concept and it should be as easy as possible to get a group chat. xmtp-group-chat has just been my playground for testing concepts out. If we feel strongly it should live elsewhere I can go that path though.

nakajima: Thinking about it a bit more, I think putting group chat features into a separate package increases the risk of fragmentation of group chat. Adding the invite capabilities here means that someone might just write their own implementation of group chat if they don't know about the other package. By keeping group chat here we're giving everybody the same set of tools by default.

neekolas: It's your call. Im on board for it living in the SDK. So long as we maintain the bar on documentation and stability for a core protocol feature. The main advantage to having it in its own package is our ability to say "jk! We are changing the protocol and all your old group chats are going to become unreadable". 

Maybe we can use the beta branch to at least buy us a little time to iterate.

src/keystore/InMemoryKeystore.ts
neekolas: Are we sure that's necessary with the invite approach. Couldn't the invites just come from whoever invited you to the group chat. No new contact bundles, no mysterious addresses.

src/keystore/InMemoryKeystore.ts
```diff
@@ -251,28 +254,139 @@ export default class InMemoryKeystore implements Keystore {
...
+        )
+      }
+
+      const invitation = new InvitationV1({
+        context: topicData.invitation.context,
+        topic: req.contentTopic,
+        aes256GcmHkdfSha256: topicData.invitation.aes256GcmHkdfSha256,
       })
+
+      const created = nsToDate(req.createdNs)
```
neekolas, line 341: One interesting quirk here is that the invitation timestamp and the envelope's timestamp need to match or `SealedInvitation.fromEnvelope` will fail

src/keystore/InMemoryKeystore.ts
nakajima: I'm trying to think how that would work in the UI. With a dedicated address for each group, we could just show group chats as normal entries. It's also a way to verify that you're actually in the same group as someone else, like "Oh are you in 0x1234...1234?"

src/conversations/Conversations.ts
nakajima: I dig using the beta branch to iterate in case we end up needing pull the "jk" ripcord, let's go with that.

src/keystore/InMemoryKeystore.ts
nakajima: It turns out I have a bug here and I actually am sending the invitation from the creator's address. I still think product-wise, it makes sense for groups to have their own addresses though.

neekolas: > With a dedicated address for each group, we could just show group chats as normal entries.

Why would we not be able to show group chats as normal entries if the invites came from the group creator (or inviter)?

> It's also a way to verify that you're actually in the same group as someone else, like "Oh are you in 0x1234...1234?"

Couldn't we use the topic for that? Or generate a random group ID and set that as metadata


Getting rid of the random wallet address feels like a big win for maintainability and simplicity IMO.

nakajima: > Getting rid of the random wallet address feels like a big win for maintainability and simplicity IMO.

Ok, I'm sold on this. I'll update to set a `groupID` in the metadata alongside the member list. We'll have to make sure clients know to display these differently in their conversation list.

nakajima:
Updated the PR body with a larger explanation of the approach here. Would love another round of review.

src/conversations/Conversations.ts
```diff
@@ -182,7 +184,14 @@ export default class Conversations {
     const out: ConversationV2[] = []
     for (const response of responses) {
       try {
-        const convo = this.saveInviteResponseToConversation(response)
+        let convo = this.saveInviteResponseToConversation(response)
+
+        const initialMembers = convo.context?.metadata.initialMembers.split(',')
+
+        if (initialMembers) {
```
neekolas, line 191: If the `convo.context.metadata.initialMembers` is set to `''`, this will actually return true because Javascript was invented in a long weekend and an empty array is truthy.

src/conversations/GroupConversation.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    if (!inviteResponse.conversation) {
+      throw new Error(
+        'no conversation for response: ' + JSON.stringify(inviteResponse)
+      )
+    }
+
+    const envelope = {
+      contentTopic: buildUserInviteTopic(
+        inviteResponse.conversation?.peerAddress
```
neekolas, line 81: Couldn't this just be `newMemberAddress`

src/conversations/Conversations.ts
```diff
@@ -429,6 +438,67 @@ export default class Conversations {
...
+          contact = SignedPublicKeyBundle.fromLegacyBundle(contact)
+        }
+
+        return toSignedPublicKeyBundle(contact)
+      })
+    )
+
+    const inviteResponses = await this.client.keystore.createInvites({
+      recipients: members,
+      context,
```
neekolas, line 470: I wonder if we want to enforce that the group ID has to go into the `conversationId`. It would make this compatible with the deterministic topic work. Otherwise, I can't add you to more than one group.

test/conversations/Conversations.test.ts
```diff
@@ -368,4 +370,24 @@ describe('conversations', () => {
       expect(invites).toHaveLength(1)
     })
   })
+
+  describe('newGroupConversation', () => {
+    it('sends invites to recipients', async () => {
+      const context = {
+        conversationId: '',
+        metadata: {
+          groupID: 'the-group-id',
```
neekolas, line 379: What if we moved this to the `conversationId` field. Something like `xmtp.org/group/$GROUP_ID`.

src/conversations/Conversations.ts
```diff
@@ -429,6 +438,67 @@ export default class Conversations {
     return seenPeers
   }
 
+  async newGroupConversation(
+    context: InvitationContext
```
neekolas, line 442: Are we sure we want to leave the responsibility of setting the metadata up to the caller? What if we made this method just take the list of members as an array and then constructed all the context for them? Feels less likely developers are going to mess that up

test/conversations/Conversations.test.ts
nakajima: I guess the question is what if folks want to use conversationIds with their group conversations?

neekolas: I'm ok with removing that possibility

nakajima: It sort of feels weird to me that we offer `conversationId` to developers for their own use except for in this case where we reserve it. What about a new property of Invitation context?

neekolas: We don't have the same deduping rules for group convos that we do for 1:1. 

ConversationId is useful in 1:1 because if you don't supply one, you might get an existing convo returned from `newConversation`. When creating a new group convo, you will always get a brand new convo (even with the same members). 

So, I don't think developers are missing much. 

We could let them set metadata optionally for groups, and merge it with our metadata.

src/conversations/Conversations.ts
nakajima: Fixed in https://github.com/xmtp/xmtp-js/pull/381/commits/ad6d5821c733216035c44971442c42dd15a9ec93.

src/conversations/GroupConversation.ts
nakajima: Fixed in https://github.com/xmtp/xmtp-js/pull/381/commits/0061e3dc26ac31525d64eab16f6a108883f789ff.

src/conversations/Conversations.ts
nakajima: Good call, fixed in https://github.com/xmtp/xmtp-js/pull/381/commits/6869df990910523f236b1863141e6f359f5f8dc5.

src/conversations/Conversations.ts
```diff
@@ -429,6 +439,72 @@ export default class Conversations {
...
+      }
+
+      return {
+        contentTopic: buildUserInviteTopic(response.conversation?.peerAddress),
+        message: response.payload,
+        timestamp,
+      }
+    })
+
+    await this.client.publishEnvelopes(envelopes)
```
neekolas, line 494: I'm not seeing where we invite ourselves to the group. Maybe I'm missing it

nakajima: I’ve been passing the creator’s address inside the initialMembers array, I could see just adding it here and deduping (which we should probably do anyway). Good call.

nakajima: Fixed in https://github.com/xmtp/xmtp-js/pull/381/commits/748720e486376536ef3e336ddba07440ad50f99c.

neekolas: Ideally we would just copy one of the other messages into our own invite topic. Then we don't have to worry about the security implications of self diffie Hellman.

nakajima: Aren't the sealed invitations DH'd anyway? I don't know if we can avoid self DH if we're sending the invites from the recipient?

nakajima: Sort of wondering if the shared wallet approach would actually be helpful due to this..

nakajima: Fixed in https://github.com/xmtp/xmtp-js/pull/381/commits/1746cca16a1545059ff81395320530ed1b9731d2.

neekolas (approved):

mkobetic:
Yup, this is very much like what we had in mind early on. I'm not sure what was discussed offline already, but we rejected this a few times due to its security shortcomings, e.g. no way to remove members, no assurance about who's actually a member (the initiator may lie in initialMembers, or the add member notification can be omitted). If we release this we should definitely make this very clear to the users. That said I'm not against this, I think it can still be useful in many scenarios lacking something better.

One note re groupID, I wonder if we could tie it to the invitation key somehow (just hash the key maybe?) and reduce a possibility of key/id mismatch.

mkobetic:
Or maybe why isn't the group topic the id?

nakajima:
> Or maybe why isn't the group topic the id? That is ultimately the most concrete manifestation of the conversation identity.

The way most clients handle conversation identity today is `peerAddress + conversationId`. Without using conversationId as a unique identifier here, most apps won't be able to distinguish between a group chat and a 1:1 chat with the creator of a group chat.

If we decide to do away with conversationIds for this purpose, we can update things to use the topic. That'll require some more work on the part of clients but it's always an option for the future.

There was also a world I was imagining where we could potentially implement group member removal by rotating topics/keys but keeping a stable conversation ID that clients can use to rectify on their end, but I'm not 100% about that idea.

nakajima:
@neekolas I am once again starting to think that having a shared wallet just for sending invitations for a group chat might be a good idea.

Clients that want to support _creating_ group chats will do it through our SDK which will handle the creation of the shared wallet and sending the invitations from it. Assuming they're using the most recent version of our SDK and building group chat, we shouldn't have to worry about them.

For clients that _aren't_ supporting group chat, sending invitations from the group creator means that users are just going to see a conversation with the group creator and not be able to tell that some messages might not be coming from that person (unless they're displaying the senderAddress next to the message, which would still be a confusing experience). There will be a conversationId present indicating that it's a group but that might not be handled gracefully for every client and conversationIds seem like they might be in a bit of a state of flux right now anyway.

If we generate a shared wallet that's only used for sending invitations, then any client that doesn't know about group chat will see a random other address and a message that says it's a group chat. If the client is not displaying senderAddress next to messages, the messages will be appearing to come from a random address which I think is preferable to looking like all messages are coming from the inviter.

The other advantage of a dedicated address for chats is that if we enter a world where member removal is possible, removing the creator/inviter of a chat could create confusion. The peerAddress for the chat would be a removed member.

Also, having a shared wallet also lets us avoid self diffie-hellman a bit cleaner than just copying a random invite message.

nakajima:
Ok, this is ready for a new round of review. I've introduced a new topic for group chat invitations so clients that don't support group chat just won't see anything (server updated [here](https://github.com/xmtp/xmtp-node-go/pull/272)). I also merged the content types PR into here.

src/conversations/Conversations.ts
```diff
@@ -129,10 +141,16 @@ export default class Conversations {
...
 
       // Load all conversations started after the newest conversation found
-      const newConversations = await this.updateV2Conversations(
+      let newConversations = await this.updateV2Conversations(
         latestConversation?.createdAt
       )
 
+      if (this.client.isGroupChatEnabled) {
+        newConversations = newConversations.concat(
+          await this.updateGroupConversations(latestConversation?.createdAt)
```
neekolas, line 150: If the latest 1:1 conversation is after the latest group conversation, won't this lead to a time period where group chats haven't been checked?

It would be a rare issue, since you _should_ be checking both at the same time. But if someone is flipping that `enableGroupChat` flag back and forth you can get into a weird situation. Not a blocker, but maybe something we should flag as a TODO.

neekolas: That makes sense. But you could still call conversations.list(), then turn on group chat. 

Not keeping me up at night, but if enough people start using it someone will do that and come to us with a very mysterious looking bug.

nakajima: Fixed in https://github.com/xmtp/xmtp-js/pull/381/commits/828b3dca945d5a907c45c25c85e489c703f000bb.

PR #384, opened 2023-05-25
Add group chat content types

nakajima:
The content types in this PR enable a more complete "group chat" experience like titles and nicknames.

### Adding a member

To add a member to a group, a member can call `keystore.createInviteFromTopic` to get an invite for a given recipient. They can send the invite to the recipient who will then be able to send/receive messages (this part is in this PR). The member adding should also send a message with content type `GroupChatMemberAdded` to the group so that other members can keep their member lists up to date.

### Updating the group title

Using the `GroupChatTitleChanged` content type, a group member can specify a new title for a group. Clients should keep track of the current title in their local persistence and update it accordingly when receiving these messages.

### Updating member nicknames 

Using the `GroupChatMemberNicknameChanged` content type, a group member can specify a new nickname for another address in the group. It's up to clients to verify that the sender of the message matches the address being changed (if that's the behavior that's desired.)

### DIFF
--- a/src/codecs/GroupChatMemberAdded.ts
+++ b/src/codecs/GroupChatMemberAdded.ts
@@ -1,0 +1,39 @@
+import { ContentCodec, ContentTypeId, EncodedContent } from '../index'
+
+export const ContentTypeGroupChatMemberAdded: ContentTypeId = {
+  typeId: 'groupChatMemberAdded',
+  authorityId: 'pat.xmtp.com',
+  versionMajor: 1,
+  versionMinor: 0,
+  sameAs(id) {
+    return (
+      this.typeId === id.typeId &&
+      this.authorityId === id.authorityId &&
+      this.versionMajor === id.versionMajor &&
+      this.versionMinor === id.versionMinor
+    )
+  },
+}
+
+export type GroupChatMemberAdded = {
+  member: string
+}
+
+export class GroupChatMemberAddedCodec
+  implements ContentCodec<GroupChatMemberAdded>
+{
+  contentType = ContentTypeGroupChatMemberAdded
+
+  encode(content: GroupChatMemberAdded): EncodedContent {
+    return {
+      type: ContentTypeGroupChatMemberAdded,
+      parameters: {},
+      content: new TextEncoder().encode(JSON.stringify(content)),
+    }
+  }
+
+  decode(encodedContent: EncodedContent): GroupChatMemberAdded {
+    const json = new TextDecoder().decode(encodedContent.content)
+    return JSON.parse(json)
+  }
+}
--- a/src/codecs/GroupChatTitleChanged.ts
+++ b/src/codecs/GroupChatTitleChanged.ts
@@ -1,0 +1,43 @@
+import { ContentCodec, ContentTypeId, EncodedContent } from '../index'
+
+export const ContentTypeGroupChatTitleChanged: ContentTypeId = {
+  typeId: 'groupChatTitleChanged',
+  authorityId: 'pat.xmtp.com',
+  versionMajor: 1,
+  versionMinor: 0,
+  sameAs(id) {
+    return (
+      this.typeId === id.typeId &&
+      this.authorityId === id.authorityId &&
+      this.versionMajor === id.versionMajor &&
+      this.versionMinor === id.versionMinor
+    )
+  },
+}
+
+export type GroupChatTitleChanged = {
+  // The new title
+  newTitle: string
+
+  // The old title
+  oldTitle: string
+}
+
+export class GroupChatTitleChangedCodec
+  implements ContentCodec<GroupChatTitleChanged>
+{
+  contentType = ContentTypeGroupChatTitleChanged
+
+  encode(content: GroupChatTitleChanged): EncodedContent {
+    return {
+      type: ContentTypeGroupChatTitleChanged,
+      parameters: {},
+      content: new TextEncoder().encode(JSON.stringify(content)),
+    }
+  }
+
+  decode(encodedContent: EncodedContent): GroupChatTitleChanged {
+    const json = new TextDecoder().decode(encodedContent.content)
+    return JSON.parse(json)
+  }
+}
--- a/src/conversations/Conversations.ts
+++ b/src/conversations/Conversations.ts
@@ -454,7 +454,9 @@ export default class Conversations {
     const context = {
       conversationId: `xmtp.org/groups/${groupID}`,
       metadata: {
-        initialMembers: initialMembers.join(','),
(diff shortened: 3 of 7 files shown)


nakajima:
> I can't quite formulate a clear argument, so take it for what it's worth, but something bugs me about using content types for this purpose. That said I don't really have an alternative to propose either. Figured I'd say something anyway.

I hear you here. I think content types are the best option we have at the moment. If we ever introduce a new meta-message concept or something, I think this would be a good candidate to move to that.

mkobetic:
Would resending the invite with updated member list and possibly with a title metadata field be an a viable alternative?

nakajima:
> Would resending the invite with updated member list and possibly with a title metadata field be a viable alternative?

It could work but then we'd need to keep track of all invites for a conversation to weave those types of messages into the conversation messages list. I think I'd rather wait until we have something more suitable for conversation state updates.

mkobetic:
> It could work but then we'd need to keep track of all invites for a conversation to weave those types of messages into the conversation messages list. I think I'd rather wait until we have something more suitable for conversation state updates.

I meant resending it with the same ID/key, so only the latest would be relevant. Isn't that what we already do with invites?

nakajima:
> I meant resending it with the same ID/key, so only the latest would be relevant. Isn't that what we already do with invites?

We'd still need to send messages to the group in that case so that we could display messages like "Pat changed the title to [title]" or "Pat added [address] to the group".

mkobetic:
OK, if we want actual messages to go with these updates, then content types start to make more sense. We could also stuff these updates into the EncodeContent [parameters](https://github.com/xmtp/proto/blob/main/proto/message_contents/content.proto#L35) and abuse the standard text content type that way, but that is stretching the spec a fair bit.

src/codecs/GroupChatMemberAdded.ts
```diff
@@ -0,0 +1,39 @@
...
+    return {
+      type: ContentTypeGroupChatMemberAdded,
+      parameters: {},
+      content: new TextEncoder().encode(JSON.stringify(content)),
+    }
+  }
+
+  decode(encodedContent: EncodedContent): GroupChatMemberAdded {
+    const json = new TextDecoder().decode(encodedContent.content)
+    return JSON.parse(json)
```
neekolas, line 37: We should at least do some runtime validation that the message actually looks anything like `GroupChatMemberAdded`. People can put anything in this JSON payload, and if we throw an error here it will be contained and the message will be ignored. If we wait until it gets accessed, it'll blow up later when it might just crash the whole app.

In a perfect world, these would all be protos.

src/conversations/Conversations.ts
```diff
@@ -185,15 +186,7 @@ export default class Conversations {
     const out: ConversationV2[] = []
     for (const response of responses) {
       try {
-        let convo = this.saveInviteResponseToConversation(response)
-
-        const initialMembers = convo.context?.metadata.initialMembers.split(',')
-
-        if (initialMembers && initialMembers.length > 0) {
-          convo = GroupConversation.from(convo, initialMembers)
```
neekolas, line 193: Don't we still need this code?

src/conversations/GroupChat.ts
```diff
@@ -0,0 +1,120 @@
...
+    const members =
+      this.conversation.context?.metadata.initialMembers.split(',')
+
+    if (!members) {
+      throw new Error('Conversation is not a group chat')
+    }
+
+    this._members = members
+
+    const startTime = opts?.since
```
neekolas, line 73: Does this get used anywhere right now?

src/codecs/GroupChatTitleChanged.ts
```diff
@@ -0,0 +1,43 @@
...
+    return {
+      type: ContentTypeGroupChatTitleChanged,
+      parameters: {},
+      content: new TextEncoder().encode(JSON.stringify(content)),
+    }
+  }
+
+  decode(encodedContent: EncodedContent): GroupChatTitleChanged {
+    const json = new TextDecoder().decode(encodedContent.content)
+    return JSON.parse(json)
```
neekolas, line 41: Also could use runtime validation

src/conversations/Conversations.ts
```diff
@@ -454,7 +447,9 @@ export default class Conversations {
     const context = {
       conversationId: `xmtp.org/groups/${groupID}`,
       metadata: {
-        initialMembers: initialMembers.join(','),
+        initialMembers: [this.client.address, ...new Set(initialMembers)].join(
```
neekolas, line 450:
```suggestion
        initialMembers: [..new Set(initialMembers.concat(this.client.address))].join(
```

That would make sure that if the sender is in the list of initial members they don't get duped.

src/conversations/GroupChat.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    await this.conversation.send(memberAdded, {
+      contentType: ContentTypeGroupChatMemberAdded,
+      contentFallback: `${this.memberClient.address} added ${newMemberAddress} to the group`,
+    })
+
+    this._members.push(newMemberAddress)
+
+    const conversation = GroupConversation.from(this.conversation, this.members)
+    await conversation.addMember(newMemberAddress)
```
neekolas, line 118: We should flip the order and do this before we send the message to the group. That way if I try and add someone not on the network it'll fail before I message everyone in the group.

src/codecs/GroupChatMemberAdded.ts
neekolas: Someone doing something like:
```
conversation.send({member: {derp: true}} as any, { contentType: ContentTypeGroupChatMemberAdded })
```
Would probably break everyone in the group's app.

src/codecs/GroupChatTitleChanged.ts
```diff
@@ -0,0 +1,43 @@
...
+    )
+  },
+}
+
+export type GroupChatTitleChanged = {
+  // The new title
+  newTitle: string
+
+  // The old title
+  oldTitle: string
```
neekolas, line 23: Do we actually use this anywhere?

nakajima: I thought it could be useful for clients, I don't think it's super necessary though. I'll remove it.

src/conversations/Conversations.ts
nakajima: This is unnecessary now that we have group invite topics.

nakajima (commented):
I think I addressed all the feedback here so I'm going to merge this PR into https://github.com/xmtp/xmtp-js/pull/381 so we can have all of the group chat changes in one place.

PR #385, opened 2023-05-31
fix: topics must be ascii

nakajima:
Without this change, topics were being stored as mangled utf-8 (probably?).

image

I think this got introduced during the deterministic topic changes.


nplasterer (approved):

neekolas (approved):

mkobetic:
Hm, this may be why I had to revert my [marginalia PR](https://github.com/xmtp/xmtp-node-go/pull/270)

PR #388, opened 2023-06-01
docs: update README for group chat

nakajima:
These docs offer the basics of group chat. They could probably use some refinement.

We might also want to add guidance to the README about defaulting to persisting messages/conversations locally and always driving UIs from that. It's an important thing for good client experiences, group chat notwithstanding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,10 +440,10 @@ const clientWithNoCache = await Client.create(wallet, {
 
 ## Group chat
 
-![Status](https://img.shields.io/badge/Feature_Status-Pre--Preview-red)
+![Status](https://img.shields.io/badge/Feature_Status-Alpha-red)
 
 > **Important:**  
-> This feature is in **Pre-Preview** status and ready for you to start experimenting with. We do **not** recommend using Pre-Preview features in production apps. Software in this status will change as we iterate based on feedback.
+> This feature is in **alpha** status and ready for you to start experimenting with. We do **not** recommend using alpha features in production apps. Software in this status will change as we iterate based on feedback.
 
 Use the information in this section to **experiment** with providing group chat in your app.
 
--- a/README.md
+++ b/README.md
@@ -438,26 +438,33 @@ const clientWithNoCache = await Client.create(wallet, {
 })
 ```
 
-## Group Chat (beta beta beta)
+## Group chat
 
-### How it works
+![Status](https://img.shields.io/badge/Feature_Status-Pre--Preview-red)
 
-First, you must enable group chat for your Client:
+> **Important:**  
+> This feature is in **Pre-Preview** status and ready for you to start experimenting with. We do **not** recommend using Pre-Preview features in production apps. Software in this status will change as we iterate based on feedback.
+
+Use the information in this section to **experiment** with providing group chat in your app.
+
+### Enable group chat for your Client
+
+The first step is to enable group chat for your Client:
 
 ```ts
 const creatorClient = await Client.create(yourSigner)
 creatorClient.enableGroupChat()
 ```
 
-Doing this does a couple things:
+This enables the following capabilities required to provide group chat:
 
-- The client will be able to create group conversations
-- Group conversations will be present in `client.conversations.list()`
+- The client will be able to create group chats
+- Group chats will be present in `client.conversations.list()`
 - The client will understand GroupChat codecs such as `GroupChatMemberAdded` and `GroupChatTitleChanged`
 
-### Creating a group
+### Create a group chat
 
-The creator of the group creates a group using `newGroupConversation` with member addresses.
+Enable a user to create a group chat using `newGroupConversation` and adding member addresses to it:
 
 ```ts
 const memberAddresses = [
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -438,6 +438,120 @@ const clientWithNoCache = await Client.create(wallet, {
...
+
+This enables the following capabilities required to provide group chat:
+
+- The client will be able to create group chats
+- Group chats will be present in `client.conversations.list()`
+- The client will understand GroupChat codecs such as `GroupChatMemberAdded` and `GroupChatTitleChanged`
+
+### Create a group chat
+
+Enable a user to create a group chat using `newGroupConversation` and adding member addresses to it:
```
jhaaaa, line 467: Just a thought as I was editing this - any thoughts on the difference between using `GroupChat...` and `GroupConversation...` in the code? 

Colloquially, I think we call this feature "group chat" - but I think " group conversation" might be more in line with the "conversation" code patterns in XMTP. Any thoughts? 

Whether we go with `GroupChat` or `GroupConversation` - might it be possible to use one pattern throughout? Or are we using the different patterns to indicate a difference? <3

README.md
```diff
@@ -438,6 +438,120 @@ const clientWithNoCache = await Client.create(wallet, {
...
+```ts
+import { ContentTypeGroupChatTitleChanged } from '@xmtp/xmtp-js'
+import type { GroupChatTitleChanged } from '@xmtp/xmtp-js'
+
+const
+```
+
+### Manage group state with the `GroupChat` class
+
+Use the `GroupChat` class to keep track of group state, such as the group title and member list:
```
jhaaaa, line 512: May I ask - when we say "group title" here - is this different from the group chat title? I think they are the same concept?

README.md
```diff
@@ -438,6 +438,120 @@ const clientWithNoCache = await Client.create(wallet, {
...
+```ts
+const conversations = await creatorClient.conversations.list()
+const conversation = conversations[0] // assume this conversation is a group conversation
+
+const groupChat = new GroupChat(creatorClient, conversation)
+```
+
+You can also use the `GroupChat` class to change the group chat title and member list.
+
+#### Change a group chat title
```
jhaaaa, line 523: Is this meant to be about changing the group title or the group chat title? I ask because in https://github.com/xmtp/xmtp-js/pull/388/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R499 we talk about how to change the group chat title... 

Are these the same concept - or different? Change group title vs group chat title?

README.md
```diff
@@ -438,6 +438,120 @@ const clientWithNoCache = await Client.create(wallet, {
...
+
+To add a group chat member, call `addMember` on a `GroupChat` instance:
+
+```ts
+await groupChat.addMember('0x194c31cAe1418D5256E8c58e0d08Aee1046C6Ed0')
+```
+
+This sends an invitation to the recipient address. It also send a `GroupChatMemberAdded` message that clients can display and use to update their gorup chat member lists.
+
+#### Rebuild the group state
```
jhaaaa, line 543: Just curious - when might a dev need to do this? After the group chat title or member list is updated?

README.md
```diff
@@ -438,6 +438,120 @@ const clientWithNoCache = await Client.create(wallet, {
...
+
+To rebuild the group state by replaying all messages in a group chat, such as group titles and members, call `rebuild()` on an instance of `GroupChat`:
```
jhaaaa, line 545: When we say "replaying all messages" - does this mean the actual messages sent between members as well? Just curious....

README.md
nakajima: The way I've been thinking about it is:

`GroupConversation` is similar to `ConversationV1` or `ConversationV2` from the SDK. It's just a way to send/receive messages.

`GroupChat` is a wrapper around `GroupConversation` that knows about things like group titles, keeping the member list in sync, basically any richer features beyond just sending/receiving messages.

I can see how this could lead to some confusion though. I wonder how we can best clarify it..

README.md
nakajima: Nope it's the same thing.

README.md
nakajima: Same thing, sorry I was inconsistent in my terminology here.

README.md
nakajima: They'd need to do this the first time they loaded the group chat to make sure they have everything up to date.

README.md
nakajima: The state update messages (like GroupChatMemberAdded, GroupChatTitleChanged) get sent alongside the actual messages, so in order to load the current state of the group, the entire chat history needs to be traversed at least once. This is also why it's important for people to locally persist messages.

README.md
jhaaaa: So good, @nakajima - thank you! I added this section above to help disambiguate: https://github.com/xmtp/xmtp-js/pull/388/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R450

jhaaaa (approved):
Thank you for all of your efforts on group chat - including these README updates! <3

jhaaaa:
@nakajima thank you for the note about "defaulting to persisting messages/conversations locally and always driving UIs from that. It's an important thing for good client experiences, group chat notwithstanding."

I'm going to work on text that we can add to all SDK READMEs/quickstarts.

jhaaaa:
OMG.. my mouse slipped and I clicked Close with comment instead of Comment! My apologies - reopened!

PR #390, opened 2023-06-01
Stream group conversations

nakajima:
This PR makes group conversations appear in `conversations.stream()`


neekolas (approved):
LGTM. You'll probably also need to add support in `streamAllMessages`

nakajima:
> LGTM. You'll probably also need to add support in streamAllMessages

Good call, fixed in https://github.com/xmtp/xmtp-js/pull/390/commits/23c96b682390884ffc1fb1f9c9807eb00db61990.

PR #391, opened 2023-06-05
Add Provenance Support

neekolas:
## Summary

- Publishes our NPM package using [Github NPM Package Provenance](https://github.blog/2023-04-19-introducing-npm-package-provenance/). This should increase trust in our published packages, since the package will display a badge assuring developers that the package they are installing really is built from the source code in our published repos.

## Notes
Once we get this working, we'll add provenance to our other NPM packages.


nakajima (approved):
This is neat.

PR #396, opened 2023-06-08
Block self-messaging

neekolas:
## Summary

We had a regression that allowed users to start conversations with themselves. This blocks that behaviour until we have a self-messaging protocol that has the security guarantees we are all comfortable with.


nplasterer (approved):
Thanks for fixing this.

PR #402, opened 2023-06-14
Add ContentTypeReaction

rygine:
Proposal: https://github.com/orgs/xmtp/discussions/36
Issue: https://github.com/xmtp-labs/hq/issues/1030

This PR adds the reaction content type with adjustments from the proposal as outlined in the comments of the issue.

TL;DR on the changes from the proposal:

* `emoji` => `content`, to be interpreted by client to allow for custom reactions
* added `action`, to allow for removing a reaction

### DIFF
--- a/src/codecs/Reaction.ts
+++ b/src/codecs/Reaction.ts
@@ -1,0 +1,60 @@
+import { ContentTypeId, ContentCodec, EncodedContent } from '../MessageContent'
+
+// xmtp.org/text
+//
+// This content type is used for a plain text content represented by a simple string
+export const ContentTypeReaction = new ContentTypeId({
+  authorityId: 'xmtp.org',
+  typeId: 'reaction',
+  versionMajor: 1,
+  versionMinor: 0,
+})
+
+export type Reaction = {
+  /**
+   * The message ID for the message that is being reacted to
+   */
+  reference: string
+  /**
+   * The action of the reaction
+   */
+  action: 'added' | 'removed'
+  /**
+   * The content of the reaction
+   */
+  content: string
+}
+
+export type ReactionParameters = Pick<Reaction, 'action' | 'reference'> & {
+  encoding: 'UTF-8'
+}
+
+export class ReactionCodec implements ContentCodec<Reaction> {
+  get contentType(): ContentTypeId {
+    return ContentTypeReaction
+  }
+
+  encode(content: Reaction): EncodedContent<ReactionParameters> {
+    return {
+      type: ContentTypeReaction,
+      parameters: {
+        encoding: 'UTF-8',
+        action: content.action,
+        reference: content.reference,
+      },
+      content: new TextEncoder().encode(content.content),
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


src/codecs/Reaction.ts
```diff
@@ -0,0 +1,60 @@
...
+   */
+  action: 'added' | 'removed'
+  /**
+   * The content of the reaction
+   */
+  content: string
+}
+
+export type ReactionParameters = Pick<Reaction, 'action' | 'reference'> & {
+  encoding: 'UTF-8'
```
rygine, line 29: i hardcoded this type to eliminate the need for an enum. we can add more in the future if necessary.

src/codecs/Reaction.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  decode(content: EncodedContent<ReactionParameters>): Reaction {
+    const encoding = content.parameters.encoding
+    if (encoding && encoding !== 'UTF-8') {
+      throw new Error(`unrecognized encoding ${encoding}`)
+    }
+    return {
+      action: content.parameters.action,
+      reference: content.parameters.reference,
+      content: new TextDecoder().decode(content.content),
```
neekolas, line 57: Do we want to put some limit (even a relatively large one like 1kb) on the size of the content? I can't think of a valid use-case for large content

src/codecs/Reaction.ts
```diff
@@ -0,0 +1,60 @@
...
+//
+// This content type is used for a plain text content represented by a simple string
+export const ContentTypeReaction = new ContentTypeId({
+  authorityId: 'xmtp.org',
+  typeId: 'reaction',
+  versionMajor: 1,
+  versionMinor: 0,
+})
+
+export type Reaction = {
```
rygine, line 13: wonder if we should be namespacing things like this to avoid clashes in the future. maybe something like `ContentTypeReactionContent` or similar?

src/codecs/Reaction.ts
rygine: base64 encoded SVG, images, etc... could be a thing 🤔

rygine:
> Curious why we are doing this here and not in https://github.com/xmtp/xmtp-js-content-types. Seems like having it in a separate repo would make it easier to share with RN

great call out! this is my first content type. wasn't sure where to put it.

src/codecs/Reaction.ts
neekolas: Hmmm. For something that different from an emoji, we wouldn't want some other metadata to tell the recipient "decode this like an image".

Is that really an intended use-case of this version of the content type?

rygine: AFAIK for this initial version, we're supporting the standard set of emojis via string tokens, so `smile` => 😄 (`:smile:`). this will allow clients to use their own custom emojis like Slack. i don't think the intent of this content type is to store actual content like images, SVGs, etc. so it makes sense to limit the length.

rygine:
this will be added to a different repo

PR #403, opened 2023-06-14
chore: allow for custom EncodedContent parameters

rygine:


daria-github (approved):

neekolas:
We can make the JS type more flexible, but the Protobuf is still going to expect a String->String mapping.

rygine:
> We can make the JS type more flexible, but the Protobuf is still going to expect a String->String mapping.

yep, this is for better autocomplete as well as defining string unions in the type.

PR #404, opened 2023-06-15
Fix group chat content types and streaming

nakajima:
Fixes two issues:

- Content types weren't instances of ContentTypeId, they were just conforming which made them unclonable
- Group Conversations returned from `client.conversations.stream()` didn't have `isGroup` set to `true`.


neekolas (approved):
LGTM

PR #410, opened 2023-07-05
feat: add `fromString` to ContentTypeId class

rygine:
The motivation behind this change is to allow for storing a content type in the parameters of a codec. Codec parameters are limited to string values, so this enables working with a content type stored as a string value.

This work unblocks the Reply content type, which requires storing the content type of a reply so that it can be encoded/decoded properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "test:cov": "jest --coverage --no-cache --runInBand",
     "lint": "prettier --check . && eslint .",
     "autolint": "prettier --write . && eslint --fix .",
-    "semantic-release": "semantic-release"
+    "semantic-release": "semantic-release",
+    "typecheck": "tsc"
   },
   "publishConfig": {
     "access": "public",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -21,7 +21,7 @@
     // "composite": true,                     /* Enable project compilation */
     // "tsBuildInfoFile": "./",               /* Specify file to store incremental compilation information */
     // "removeComments": true,                /* Do not emit comments to output. */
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -36,7 +36,8 @@
     "test:cov": "jest --coverage --no-cache --runInBand",
     "lint": "prettier --check . && eslint .",
     "autolint": "prettier --write . && eslint --fix .",
-    "semantic-release": "semantic-release"
+    "semantic-release": "semantic-release",
+    "typecheck": "tsc"
```
rygine, line 40: added this for convenience

src/Client.ts
```diff
@@ -110,8 +110,7 @@ export type ContentOptions = {
   /**
    * Allow configuring codecs for additional content types
    */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  codecs: ContentCodec<any>[]
+  codecs: ContentCodec[]
```
rygine, line 113: with the type updates, the `ContentCodec` generic type no longer needs an explicit value. the default is `any`.

tsconfig.json
```diff
@@ -21,7 +21,7 @@
     // "composite": true,                     /* Enable project compilation */
     // "tsBuildInfoFile": "./",               /* Specify file to store incremental compilation information */
     // "removeComments": true,                /* Do not emit comments to output. */
-    // "noEmit": true,                        /* Do not emit outputs. */
+    "noEmit": true /* Do not emit outputs. */,
```
rygine, line 24: added this so that type checking doesn't generate output

daria-github (approved):

PR #411, opened 2023-07-07
fix: capture known deterministic topic for cross-SDK checks

dmccartney:
This is helpful for verifying that generated invites match across SDKs.

e.g. https://github.com/xmtp/xmtp-flutter/pull/76


nplasterer (approved):
Nice glad you found a way to test this locally!

PR #412, opened 2023-07-07
Add validation for private key public key match

neekolas:
## Summary

Our `PrivateKey` proto types embed the associated public key. We haven't historically had a lot of need to validate `PrivateKey`s, because they are either freshly generated or are coming from an encrypted payload. 

In the context of the Snap, I would like to validate that the private keys coming from the application have matching public keys. Otherwise a malicious app could construct a malicious private key using the publicly available signed public key (from a contact topic) and a different private key. We should detect and reject those types of keys.

## Note
I haven't included a step to validate these keys when we extract them from `EncryptedPrivateKeyBundle` but we may want to do that as well just as an additional safety check. Maybe we even want to include this in the constructor. I'll leave that for another PR

### DIFF
--- a/src/crypto/PrivateKey.ts
+++ b/src/crypto/PrivateKey.ts
@@ -155,6 +155,14 @@ export class SignedPrivateKey
     return privateKey.SignedPrivateKey.encode(this).finish()
   }
 
+  validatePublicKey(): boolean {
+    const generatedPublicKey = secp.getPublicKey(this.secp256k1.bytes)
+    return equalBytes(
+      generatedPublicKey,
+      this.publicKey.secp256k1Uncompressed.bytes
+    )
+  }
+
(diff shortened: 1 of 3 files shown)


mkobetic (approved):
I don't quite understand the attack you describe, but this seems like a good hygiene regardless. At the very least it could catch some types of mistakes.

neekolas:
@mkobetic the gist is that when setting up the Snap for the first time the client uploads the private key bundle. We want automatically trust whatever origin uploaded a valid bundle, and only prompt for additional authorization when the bundle is accessed from other origins. But that means we should be sure that the bundle is actually valid, or else a malicious app could upload a bad bundle and bork your Snap. 

It's a fringe edge case, but the Snap is going into audit soon and it feels like the kind of thing an auditor would notice and flag.

PR #422, opened 2023-07-20
feat: refactor message content fallback

rygine:
BREAKING CHANGE: This changes what the SDK returns for messages without a content codec, potentially breaking implementations that rely on the `ContentTypeFallback` content type.

That being said, we could potentially treat this is a minor bump.

### Why is the current implementation a problem?

There's an ongoing effort in our React SDK to make web apps local first, which means storing conversations and messages in a local DB that is treated as the source of truth. When a client doesn't support a content type, the JS SDK returns a fallback message that doesn't expose the original content type or content. Because of this, it's impossible to re-process messages that contain an unsupported content type without re-querying the network.

With this update, we'll be able to store the original `contentType` in the local DB so that the message can be decoded and processed properly once a client adds support for its content type.

### DIFF
--- a/src/Message.ts
+++ b/src/Message.ts
@@ -240,6 +240,7 @@ export class DecodedMessage {
   content: any // eslint-disable-line @typescript-eslint/no-explicit-any
   error?: Error
   contentBytes: Uint8Array
+  contentFallback?: string
 
   constructor({
     id,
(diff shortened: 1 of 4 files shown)


nakajima (approved):
I feel like this _might_ warrant a major version change just because it'll be breaking. But I dig the approach.

PR #426, opened 2023-07-31
Move fallback content definition to content codecs

nakajima:
This will allow content types to define standard fallback content instead of forcing clients to figure it out for themselves, which should help with interoperability across different apps. Messages that have `undefined` fallback content should be hidden from users.

If this lands, we can update the packages in https://github.com/xmtp/xmtp-js-content-types to define their own fallback contents. For attachment it could include something like the filename, for replies, it could show something else, for read receipts it could just return `undefined`.


src/MessageContent.ts
```diff
@@ -57,6 +57,7 @@ export interface ContentCodec<T> {
   contentType: ContentTypeId
   encode(content: T, registry: CodecRegistry): EncodedContent
   decode(content: EncodedContent, registry: CodecRegistry): T
+  fallback(content: T): string | undefined
```
neekolas, line 60: Love this. 

I think this qualifies as a breaking change, since all existing codecs will no longer satisfy the type.

saulmc:
> If this lands, we can update the packages in https://github.com/xmtp/xmtp-js-content-types to define their own fallback contents.

@darickdang @yash-luna Let's get a discussion started about proposed fallback content for the existing content types, so that devs can give feedback on this proposal with full context about what changes to expect.

neekolas (commented):
This is a great improvement to how content types work, but definitely looks like a breaking change that should go into the next major release.

nmalzieu:
Love it! Codecs should define their own fallbacks.
I think it def fixes the UX issue for someone with good intentions.
But it doesn't change the fact that someone with bad indentions could fork content type and put malicious content in the fallback right?

src/MessageContent.ts
mkobetic: I don't understand how this is meant to work. Presumably the client that needs the fallback does not have the codec, otherwise it can decode the content instead. What am I missing?

mkobetic: nm, I see now, this is invoked during encoding, not decoding.

neekolas (approved):
Approving. But I think we should re-point this to the beta branch

neekolas:
FWIW, I have a `beta` release slated for ~next week @nakajima. Let's get this in there for v11

rygine (approved):

Issue #427, opened 2023-08-07
Feature request: Notify when stream connection lost


neekolas:
### Is your feature request related to a problem?

When streaming messages, the connection may be temporarily lost for a range of reasons. If callers were able to pass in a callback to be notified of connection failures they could fill in gaps in their history and ensure no messages are missed. 

### Describe the solution to the problem

Add an optional `onDisconnect` callback hook to all streaming methods which would be called every time the connection is closed (except for cases where the caller has explicitly disconnected)

### Describe the uses cases for the feature

Bots that strive for 100% deliverability of streamed messages

PR #435, opened 2023-08-09
Add onConnectionLost callback to all Stream methods

neekolas:
## Summary

- Adds an optional onConnectionLost callback to all streaming methods to allow for callers to be notified when the stream is reconnecting. This can be used to fill in gaps in their message history.

## Fixes
https://github.com/xmtp/xmtp-js/issues/427

### DIFF
--- a/src/ApiClient.ts
+++ b/src/ApiClient.ts
@@ -92,6 +92,8 @@ export type SubscribeCallback = NotifyStreamEntityArrival<messageApi.Envelope>
 
 export type UnsubscribeFn = () => Promise<void>
 
+export type OnConnectionLostCallback = () => void
+
 const isAbortError = (err?: Error): boolean => {
   if (!err) {
     return false
@@ -216,7 +218,8 @@ export default class ApiClient {
   // Raw method for subscribing
   private _subscribe(
     req: messageApi.SubscribeRequest,
-    cb: NotifyStreamEntityArrival<messageApi.Envelope>
+    cb: NotifyStreamEntityArrival<messageApi.Envelope>,
+    onConnectionLost?: OnConnectionLostCallback
   ): UnsubscribeFn {
     const abortController = new AbortController()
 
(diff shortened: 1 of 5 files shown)


src/ApiClient.ts
```diff
@@ -246,9 +253,14 @@ export default class ApiClient {
...
             err.toString()
           )
+
           if (new Date().getTime() - startTime < 1000) {
             await sleep(1000)
           }
+
+          if (onConnectionLost) {
+            onConnectionLost()
+          }
```
rygine, line 263: i'm not super familiar with this code, but curious why the callback is called after sleeping?

src/Stream.ts
```diff
@@ -25,17 +25,21 @@ export default class Stream<T> {
...
 
+  onConnectionLostCallback?: OnConnectionLostCallback
+
   constructor(
     client: Client,
     topics: string[],
     decoder: MessageDecoder<T>,
-    contentTopicUpdater?: ContentTopicUpdater<T>
+    contentTopicUpdater?: ContentTopicUpdater<T>,
+    onConnectionLostCallback?: OnConnectionLostCallback
```
rygine, line 35: my only nit is that we should name this argument `onConnectionLost` across the board instead of `onConnectionLostCallback`. that being said, i'm more interested in consistency than anything. there's some mixing and matching between `onConnectionLost` and `onConnectionLostCallback` (excluding the type name, which i think is fine).

src/ApiClient.ts
neekolas: The reason is that if we are sleeping (to avoid slamming the network), we want the caller to be able to fill in the message history from a time closer to when the reconnect happens. Otherwise you could handle the `onConnectionLost` by filling in missing messages by querying and still have a second missing.

src/Stream.ts
neekolas: Fair point. Fixed.

src/ApiClient.ts
neekolas: Effectively, the time it spends sleeping is all part of the time the connection was lost.

src/ApiClient.ts
```diff
@@ -237,6 +240,10 @@ export default class ApiClient {
...
```
rygine, line 246: super nit for these 2 calls: WDYT of doing `onConnectionLost?.()` instead using the `if` statement? definitely not a blocker!

rygine (approved):
LGTM

Issue #438, opened 2023-08-11
Feature request: Pluggable ApiClient implementation


neekolas:
### Is your feature request related to a problem?

The current `ApiClient` implementation uses the HTTP API, which is not very reliable for long-lived streaming sessions. Users of the SDK should be able to provide an alternative `ApiClient`. The first implementation of this pluggable `ApiClient` would be a true GRPC client, which would offer much better performance and reliable streaming.

### Describe the solution to the problem

Allow SDK users to provide their own `ApiClientFactory` function at the time of `Client` instantiation which takes in the `XmtpEnv` and returns an instance conforming to the interface of `ApiClient`.

The default would return the current version of `ApiClient`

### Describe the uses cases for the feature

Any server-side application using the XMTP network with `xmtp-js`

### Additional details

The reason for not including the GRPC client directly in the SDK is that it relies on a bunch of large libraries that are not compatible with the browser. We _could_ do some cleverness at the bundling stage to only include the GRPC client for Node.js, but I think it might be cleaner to just bundle the GRPC client as a separate package.

PR #440, opened 2023-08-21
Pluggable api client

neekolas:
## Summary

- Creates an interface type for `ApiClient` to allow for alternative implementations
- Allows the user to override the default `ApiClient` creation process with their own implementation
- Adds a bunch of exports to the package to facilitate building your own `ApiClient` implementation
- Moves all base64 decoding into the `ApiClient` instead of handling in the application
- Changes return type of `subscribe` method, to support future changes which will allow updating content topics without closing the stream

## Solves
https://github.com/xmtp/xmtp-js/issues/438

## Notes
Bit of backstory on the base64 decoding. The code generator for GRPC gateway is inconsistent with the envelope type. When sending a message, you can provide the `message` field as a Uint8Array. But when receiving messages, it returns that field as a base64 encoded string. In order to provide a consistent API that matches the new type definition, we need to consolidate that logic in the `HttpApiClient`. This also just feels tidier.

This is a breaking change and will go into the next major SDK release


mkobetic (approved):
Looks like nice clean up.

PR #444, opened 2023-08-23
Conversation List Caching Improvements Part I

neekolas:
## Summary

- Makes `persistence` mandatory for an `InMemoryKeystore`. There really aren't a lot of use-cases for persistence to be optional in the Keystore, and making the field mandatory cleans up the code quite a bit. There is a marginal performance penalty to this change for server-side apps that don't need to persist conversations, who now have to serialize the conversation data and store it in memory. But my goal with v11 is to make those types of applications a thing of the past.
- Took the opportunity to break out `InMemoryPersistence` into it's own type, as opposed to a variation of `LocalStoragePersistence` that was only available when `window.localStorage` was not set. This is more explicit and should lead to more predictable behaviour.
- Adds new `GetRefreshJob` and `SetRefreshJob` keystore methods, which will be used to store the timestamp of the last run of `conversations.list()`.

## Notes

This is all in the pursuit of fixing a nasty bug that I found in `xmtp-js`. We use the timestamp of the most recent conversation available in the Keystore as the `startTime` for subsequent calls to the network to refresh conversations. The issue is that the Keystore may also be filled by `client.conversations.stream()` (`client.conversations.streamAllMessages` should be mostly fine, since it refreshes the conversation list via query before opening the stream). In cases where someone has been offline for a period of time and then discovers a new conversation via streaming, there is a gap in the conversation history that will never be filled.


src/Client.ts
```diff
@@ -206,7 +210,10 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
     maxContentSize: MaxContentSize,
     persistConversations: true,
     skipContactPublishing: false,
-    basePersistence: new LocalStoragePersistence(),
+    basePersistence:
+      typeof localStorage === 'undefined'
```
rygine, line 214: can we use something less brittle than a magic variable on `global`/`globalThis` to determine the environment?

this seems to be a common way of determining if we're in a browser:

```
const isBrowser =
  typeof window !== "undefined" && typeof window.document !== "undefined";
```

WDYT?

src/keystore/InMemoryKeystore.ts
```diff
@@ -398,6 +410,53 @@ export default class InMemoryKeystore implements Keystore {
...
+    jobType,
+    lastRunNs,
+  }: keystore.SetRefeshJobRequest): Promise<SetRefreshJobResponse> {
+    const key = await this.buildJobStorageKey(jobType)
+    await this.jobStatePersistence.setItem(
+      key,
+      Uint8Array.from(lastRunNs.toBytes())
+    )
+
+    return {}
```
rygine, line 440: why return `{}` here?

src/Client.ts
neekolas: I'm good with that

rygine (approved):

PR #445, opened 2023-08-24
Conversation List Caching Part II

neekolas:
## Summary

- Changes the way we list conversations to deal with the bug referenced in [this issue](https://github.com/xmtp/xmtp-js/pull/444). We now keep an explicit marker of the last time `conversations.list()` has been run, instead of implicitly using the timestamp of the most recent conversation.
- Creates a new job runner that wraps up getting and setting the most recent run time
- Does a bunch of work to make it possible to cache V1 conversations in the Keystore. Previously, V1 conversations were only cached in memory and worked quite differently between V1 and V2. This will have both performance benefits and makes the code more consistent

### Why do everything in the Keystore?

Persistence needs to be encapsulated in the Keystore, since different Keystore implementations may have a persistence layer that is inaccessible from the application. If we were to store the `lastRun` timestamp from the application, and a user was taking advantage of the `SnapKeystore` we could run into a situation where the application's persistence was wiped but not the Snap's (or vice versa).


src/conversations/Conversations.ts
rygine, line 1: unused import from `async-mutex` in this file

test/conversations/Conversations.test.ts
```diff
@@ -288,13 +266,11 @@ describe('conversations', () => {
       expect(aliceConvo instanceof ConversationV1).toBeTruthy()
       await bob.publishUserContact(false)
       alice.forgetContact(bob.address)
-      await sleep(100)
```
neekolas, line 291: I removed a bunch of these sleeps, which haven't been necessary ever since we made publishing synchronous on the nodes.

PR #446, opened 2023-08-29
Snap Support

neekolas:
## Summary

- Exports request and response types for the Keystore API to generate the handlers inside the Snap
- Changes return type of `getV2Conversations` from an array of objects to `GetConversationsResponse` to conform with the standard request/response handlers in the Snap
- Adds (but does not yet enable) the new `SnapKeystoreProvider`
- Adds feature-detection for the Metamask Snap so that we know whether or not Snaps are available when we start the SDK.
- Adds types to `window.ethereum`


src/codecs/TypingNotification.ts
rygine, line 1: 🎉

PR #447, opened 2023-08-30
Remove typing notification

neekolas:
## Summary

- Removes TypingNotification from `xmtp-js` so that it can be included in `xmtp-js-content-types` (if desired). I'm not sure about the status of this one, other than that it doesn't need to be in this repo.


neekolas:
Closing as it was also part of #446

PR #449, opened 2023-08-31
Topic non-determinism fix

neekolas:
## Summary

- Fixes issue with topics not being deterministic in both directions.

### DIFF
--- a/src/keystore/InMemoryKeystore.ts
+++ b/src/keystore/InMemoryKeystore.ts
@@ -279,7 +279,7 @@ export default class InMemoryKeystore implements Keystore {
       const secret = await this.v2Keys.sharedSecret(
         recipient,
         this.v2Keys.getCurrentPreKey().publicKey,
-        myAddress > theirAddress
+        myAddress < theirAddress
       )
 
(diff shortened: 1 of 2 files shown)


src/keystore/InMemoryKeystore.ts
```diff
@@ -273,17 +273,16 @@ export default class InMemoryKeystore implements Keystore {
...
       const created = nsToDate(req.createdNs)
       const recipient = toSignedPublicKeyBundle(req.recipient)
+      const myAddress = await this.getAccountAddress()
+      const theirAddress = await recipient.walletSignatureAddress()
 
       const secret = await this.v2Keys.sharedSecret(
         recipient,
         this.v2Keys.getCurrentPreKey().publicKey,
-        false
+        myAddress > theirAddress
```
neekolas, line 282: Could do something else here. Just needs to be the same irrespective of who is sending.

nplasterer: To get the old deterministic tests to pass that expected a certain topic we will have to flip this to `<` but that shouldn't make difference as long as it's consistent across the SDKs

rygine (approved):

PR #452, opened 2023-09-08
Snaps Quality Of Life Improvements

neekolas:
## Summary

Adds four quality of life improvements for Snaps

1. Adds a static method `isSnapsReady` so that callers can know in advance that `getKeys` will fail, and that the sign in flow will be different
2. Adds support for `privateKeyOverride` even when the user is using Snaps. This can be a migration path for apps that are storing keys locally today
3. Adds a npm script to get the latest snap version and save it in the repo
4. Allows enabling Snaps with a single `ClientCreateOption`, using the version provided by `#3`.


src/Client.ts
```diff
@@ -222,6 +230,13 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
     opts.codecs = _defaultOptions.codecs.concat(opts.codecs)
   }
 
+  if (opts?.useSnaps) {
+    opts.keystoreProviders = [
+      new SnapProvider(`npm:${snapPackage}`, snapVersion),
```
richardhuaaa, line 235: Noob question, what do we do with this version eventually?

richardhuaaa (approved):

src/Client.ts
neekolas: I'm putting together a bigger doc around it, but we use it to upgrade the version of the user's installed Snap if the specified version is higher than their installed version.

PR #455, opened 2023-09-12
Strongly typed client

neekolas:
## Summary

A long-standing source of frustration with our JS SDK is the use of `any` types for message content. Sending and receiving messages has the `content` field typed as `any`.

Message content is of a finite set of types, defined by the codecs that get passed in to Client creation. If you have a codec that returns `number`, it's possible to get messages with content of type number. If you don't a number codec, it's not possible.

We just need to do some Typescript wrangling to infer those types from the codecs and pass them around to all the conversation methods.

This PR does that.

CleanShot 2023-09-12 at 16 38 37@2x

## Notes

I've structured this to be safe for existing implementers. All the existing types default to `any`. This will make the migration path smoother for developers, but also may lead devs to not take advantage of this feature.

```ts
// This will have strong typing
const goodClient = await Client.create(someWallet)
const goodConvo = await goodClient.conversations.newConversation(someAddress)
// Will cause a type error because the client is set to have a type of "string"
await goodConvo.send(1234)

// Coerce the type of Client to the default
let badClient: Client
badClient = await Client.create(someWallet)
const badConvo = await badClient.conversations.newConversation(someAddress)
// The type-checker will ignore this, because Client is set to the default of any
await badConvo.send(1234)

// This will also disable the strong typing
const otherBadConvo: Conversation = await goodClient.conversations.newConversation(someAddress)
// This will be allowed
await otherBadConvo.send(1234)
```

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -246,6 +246,7 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
  * Client class initiates connection to the XMTP network.
  * Should be created with `await Client.create(options)`
  */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export default class Client<T = any> {
   address: string
   keystore: Keystore
@@ -319,7 +320,7 @@ export default class Client<T = any> {
     const address = publicKeyBundle.walletSignatureAddress()
     apiClient.setAuthenticator(new KeystoreAuthenticator(keystore))
     const backupClient = await Client.setupBackupClient(address, options.env)
-    const client = new Client<ExtractDecodedType<[...U, TextCodec][number]>>(
+    const client = new Client(
       publicKeyBundle,
       apiClient,
       backupClient,
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -246,7 +246,7 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
  * Should be created with `await Client.create(options)`
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export default class Client<T = any> {
+export default class Client<ContentTypes = any> {
   address: string
   keystore: Keystore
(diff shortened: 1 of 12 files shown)


src/Client.ts
```diff
@@ -244,7 +246,8 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
  * Client class initiates connection to the XMTP network.
  * Should be created with `await Client.create(options)`
  */
-export default class Client {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export default class Client<T = any> {
```
neekolas, line 250: This would be better, but also much more disruptive, if I made it `<T = unknown>`

src/Message.ts
```diff
@@ -317,16 +310,16 @@ export class DecodedMessage {
...
+    content: T, // eslint-disable-line @typescript-eslint/no-explicit-any
     contentType: ContentTypeId,
     contentBytes: Uint8Array,
     contentTopic: string,
-    conversation: Conversation,
+    conversation: Conversation<T>,
     error?: Error,
     contentFallback?: string
-  ): DecodedMessage {
+  ): DecodedMessage<typeof content> {
```
rygine, line 322: i think the return type can just be `DecodedMessage<T>`

src/Message.ts
```diff
@@ -347,17 +340,17 @@ export class DecodedMessage {
...
     contentType: ContentTypeId,
     contentTopic: string,
     contentBytes: Uint8Array,
-    conversation: Conversation,
+    conversation: Conversation<T>,
     senderAddress: string,
     error?: Error,
     contentFallback?: string
-  ): DecodedMessage {
+  ): DecodedMessage<typeof content> {
```
rygine, line 353: same as above

rygine (approved):
some nits, but nothing blocking. great work, LGTM!

PR #460, opened 2023-09-26
fix: support byte array messages with viem signer

rygine:


neekolas (approved):
Would be good to have a test case using the signer with both a string and a Uint8Array

PR #462, opened 2023-10-06
Add self-encryption support

neekolas:
## Summary

- Creates a class wrapping our new WASM-backed self encryption
- The results from `encrypt` are safe to publish directly to the network


nakajima (approved):
✨

PR #467, opened 2023-10-17
feat: return DecodedMessage when sending PreparedMessage

rygine:
after sending a prepared message, there's no easy way to access the `DecodedMessage` of the sent message. this PR returns a `DecodedMessage` from the `send` method of a `PreparedMessage`.

### DIFF
--- a/src/conversations/Conversation.ts
+++ b/src/conversations/Conversation.ts
@@ -235,22 +235,24 @@ export class ConversationV1<ContentTypes>
       recipient = recipient.toLegacyBundle()
     }
 
+    const topic = options?.ephemeral ? this.ephemeralTopic : this.topic
+
     if (!this.client.contacts.has(this.peerAddress)) {
       topics = [
         buildUserIntroTopic(this.peerAddress),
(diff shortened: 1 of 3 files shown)


src/conversations/Conversation.ts
```diff
@@ -247,21 +247,31 @@ export class ConversationV1<ContentTypes>
...
           timestamp: msg.sent,
         }))
       )
+
+      return DecodedMessage.fromV1Message(
+        msg,
+        content,
+        options?.contentType || ContentTypeText,
+        payload,
+        topics[0],
```
neekolas, line 272: Think we want it to always be `this.topic`. For a V1 message, `topics[0]` will be the intro topic

rygine: ahh OK, i originally had it like that, but then saw what the `send` method is doing. will update!

neekolas: Oof. We should change which topic gets returned from send too, then.

neekolas (approved):
Thanks for fixing the bugs you spotted along the way

PR #470, opened 2023-10-25
Add PPPP support

rygine:
WIP

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -44,6 +44,7 @@ import { hasMetamaskWithSnaps } from './keystore/snapHelpers'
 import { version as snapVersion, package as snapPackage } from './snapInfo.json'
 import { ExtractDecodedType } from './types/client'
 import type { WalletClient } from 'viem'
+import { Contacts } from './Contacts'
 const { Compression } = proto
 
 // eslint-disable @typescript-eslint/explicit-module-boundary-types
@@ -194,6 +195,10 @@ export type PreEventCallbackOptions = {
   preEnableIdentityCallback?: PreEventCallback
 }
 
+export type AllowListOptions = {
+  enableAllowList: boolean
+}
+
 /**
  * Aggregate type for client options. Optional properties are used when the default value is calculated on invocation, and are computed
  * as needed by each function. All other defaults are specified in defaultOptions.
@@ -203,7 +208,8 @@ export type ClientOptions = Flatten<
     KeyStoreOptions &
     ContentOptions &
     LegacyOptions &
-    PreEventCallbackOptions
+    PreEventCallbackOptions &
+    AllowListOptions
 >
 
 /**
@@ -226,6 +232,7 @@ export function defaultOptions(opts?: Partial<ClientOptions>): ClientOptions {
     disablePersistenceEncryption: false,
     keystoreProviders: defaultKeystoreProviders(),
     apiClientFactory: createHttpApiClientFromOptions,
+    enableAllowList: false,
   }
 
   if (opts?.codecs) {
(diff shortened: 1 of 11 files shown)


src/utils/topic.ts
```diff
@@ -31,7 +31,11 @@ export const buildUserInviteTopic = (walletAddr: string): string => {
...
   return buildContentTopic(`invite-${utils.getAddress(walletAddr)}`)
 }
+
 export const buildUserPrivateStoreTopic = (addrPrefixedKey: string): string => {
   // e.g. "0x1111111111222222222233333333334444444444/key_bundle"
   return buildContentTopic(`privatestore-${addrPrefixedKey}`)
 }
+
+export const buildUserPrivatePreferencesTopic = (identifier: string) =>
+  buildUserPrivateStoreTopic(`${identifier}/allowlist`)
```
neekolas, line 41: We should get everyone aligned on what the topic identifier actually is. Pat's doc mentioned the private store topic, but that was before we created these new identifiers.

We should do something new so that this has no chance of interfering with existing private stores (or screwing up our analytics on the growth of the network).

Maybe something like `buildContentTopic(`pppp-${identifier}`)

src/Client.ts
```diff
@@ -251,7 +258,7 @@ export default class Client<ContentTypes = any> {
   address: string
   keystore: Keystore
   apiClient: ApiClient
-  contacts: Set<string> // address which we have connected to
+  contacts: Contacts
```
neekolas, line 261: We just have to be very sure that this is a drop-in replacement for the functionality we were using the Contacts set for previously. Mostly just for deciding whether to send an intro to someone in a V1 conversation.

Should be fine.

src/Contacts.ts
```diff
@@ -0,0 +1,205 @@
...
+    return entry
+  }
+
+  state(address: string) {
+    return this.entries.get(address) ?? 'unknown'
+  }
+
+  static async load(client: Client): Promise<AllowList> {
+    const allowList = new AllowList()
+    const privateKeyBundle = await client.keystore.getPrivateKeyBundle()
```
neekolas, line 71: This will break in the browser when using Snaps, since this method isn't supported in a Snap.

You're actually going to have to add the decryption methods to the Keystore and follow the Keystore release process [outlined here](https://github.com/xmtp/xmtp-js/tree/main/src/keystore).

Apologies for not warning you about that wrinkle. We can test in the `beta` branch while we wait for MetaMask to support our new Snap release, but can't release this version of the SDK until there is a corresponding Snap version.

src/utils/topic.ts
rygine: cc @nplasterer

src/Client.ts
rygine: yep, it has _very_ limited usage.

src/utils/topic.ts
nplasterer: So just to be clear it would be `"pppp-$identifier"` not  `"pppp-$identifier/allowlist"`

neekolas: My proposal is that all PPPP events (allow, block, some future feature) would go into one topic, and the client would reduce the events into a consistent state. 

Could be convinced we want to split the topics up. Just more fetching.

rygine: i think `pppp-$identifier/allowlist` works great and allows for adding more types of preferences in the future.

nplasterer: Okay that makes sense @neekolas so in that case it should just be `pppp-$identifier` we make the block and allow list from those events and serve them to them from the class. It just feels confusing to add `allowlist` to the end here when we might have pinned wallets etc... It feels like all the preferences should get loaded at the same time so no need to have separate topics for each type of preference? Unless we think that will restrict us in the future when preference may get more complicated with groups.

neekolas:
We should point this PR at the beta branch, given the size of the new dependency we are adding.

rygine:
closing this in favor of https://github.com/xmtp/xmtp-js/pull/477

PR #471, opened 2023-10-25
Update LICENSE

bwcDvorak:
Updated copyright notice

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2022 XMTP Labs (xmtp.com)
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2022 XMTP Labs (xmtp.com)
+Copyright (c) 2023 XMTP (xmtp.org)
```
nplasterer, line 3: I think it should still be 2022 right? Since the was the year it was created?

nplasterer (approved):
I don't know how licensing works so feel free to ignore the comment.

PR #477, opened 2023-10-26
feat: add PPPP support

rygine:


src/Contacts.ts
```diff
@@ -0,0 +1,203 @@
...
+            block:
+              entry.permissionType === 'block'
+                ? {
+                    walletAddresses: [entry.value],
+                  }
+                : undefined,
+          }
+          const payload =
+            privatePreferences.PrivatePreferencesAction.encode(action).finish()
+          const result = await client.keystore.selfEncrypt({
```
neekolas, line 130: Less important for encryption than decryption, but would be nice to take advantage of the batch capabilities of the API and do them all at once. At least in the context of the Snap, there's some overhead to each request.

src/Contacts.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  static async load(client: Client): Promise<AllowList> {
+    const allowList = new AllowList()
+    const identifier = await this.getIdentifier(client)
+    const envelopes = (
+      await client.listEnvelopes(
+        buildUserPrivatePreferencesTopic(identifier),
+        async ({ message }: Envelope) => {
+          if (message) {
+            const result = await client.keystore.selfDecrypt({
```
neekolas, line 80: Same as below. Would be a nice optimization to decrypt in batches. Especially as people's allowlists get longer.

src/Client.ts
```diff
@@ -731,7 +733,7 @@ export default class Client<ContentTypes = any> {
     for (const env of envelopes) {
       if (!env.message) continue
       try {
-        const res = await mapper(env)
+        const res = await mapper(env as EnvelopeWithMessage)
```
rygine, line 736: updated the types here as the mapper never receives an envelope without a message

neekolas (approved):

src/Contacts.ts
```diff
@@ -0,0 +1,211 @@
+import Client from './Client'
+import { privatePreferences } from '@xmtp/proto'
+import { EnvelopeWithMessage, buildUserPrivatePreferencesTopic } from './utils'
+
+export type AllowListPermissionType = 'allow' | 'block' | 'unknown'
+
+export type AllowListEntryType = 'address'
+
+export class AllowListEntry {
```
nplasterer, line 9: I'm not the biggest fan of this being `AllowList....` etc From the proposal it encourages the verbage `ConsentList` `ConsentState` etc... I went around the iOS Android and RN sdks to clean this up to be more consistent with the word Consent instead of Allow since it returns block, allow, and unknown.

src/Contacts.ts
```diff
@@ -0,0 +1,211 @@
+import Client from './Client'
+import { privatePreferences } from '@xmtp/proto'
+import { EnvelopeWithMessage, buildUserPrivatePreferencesTopic } from './utils'
+
+export type AllowListPermissionType = 'allow' | 'block' | 'unknown'
```
nplasterer, line 5: Per that document as well it uses `'allowed' | 'blocked' | 'unknown'`

https://github.com/orgs/xmtp/discussions/49

neekolas:
One thing that is also going to be needed is automatically allowing conversations started by you.

rygine:
> One thing that is also going to be needed is automatically allowing conversations started by you.

this should do it: https://github.com/xmtp/xmtp-js/pull/477/files#diff-a2bb4a0ed491f9b111e8a1b23156a428da7e44313e788d7edf757a40af7a2a6cR573-R576

neekolas (approved):

nplasterer (approved):
LGTM!

src/Contacts.ts
```diff
@@ -0,0 +1,209 @@
...
+    }, [] as privatePreferences.PrivatePreferencesAction[])
+
+    actions.forEach((action) => {
+      action.allow?.walletAddresses.forEach((address) => {
+        consentList.allow(address)
+      })
+      action.block?.walletAddresses.forEach((address) => {
+        consentList.block(address)
+      })
+    })
```
nplasterer, line 103: I just discovered this in my RN testing but this will override for newer entries with older ones. You need to reverse the list before iterating.

rygine: @neekolas to confirm, `client.listEnvelopes` should return these messages in ASC ordering, which is the correct order for processing?

PR #479, opened 2023-10-27
Add `startTime` option, remove auto-fetch and auto-allow consent calls

rygine:
in this PR:

- added `startTime` option when loading / refreshing a consent list to allow for smaller updates when consent is cached locally
- removed client option to enable consent
- removed consent list auto-fetch when fetching conversations

### DIFF
--- a/src/conversations/Conversations.ts
+++ b/src/conversations/Conversations.ts
@@ -49,9 +49,6 @@ export default class Conversations<ContentTypes = any> {
       this.listV2Conversations(),
     ])
 
-    // fetch allow list if enabled
-    this.client.contacts.refreshConsentList()
-
     const conversations = v1Convos.concat(v2Convos)
 
     conversations.sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime())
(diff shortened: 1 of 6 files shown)


src/conversations/Conversations.ts
```diff
@@ -570,11 +567,6 @@ export default class Conversations<ContentTypes = any> {
       },
     ])
 
-    // add peer address to allow list
-    if (this.client._enableConsentList) {
-      this.client.contacts.allow([peerAddress])
```
neekolas, line 575: Curious why we're removing this. I thought that was the desired behaviour between all implementations

rygine: currently discussing this with @nplasterer. this isn't set in stone. the idea behind removing the automatic consent calls is to allow developers to opt-in to this consent model. it feels weird to force developers into using consent... without their consent.

another option here would be to add an option to auto-allow consent to the `newConversation` method.

rygine: re-added this allow call

nplasterer: Ry and I discussed this and we decided it makes the most sense to keep it in place.

We take creating of a conversation as explicit consent that you would like to allow this person to talk to you. In the future you can of course block them if you desire. I honestly think we could take this step further that if a user sends a message to a conversation that has a state of unknown that it is also explicit consent and we should update the state to allowed. This avoids potential confusion if you are having a conversation with someone in app that doesnt have consent and then return to an app like converse those messages would show as requests. *** that being said we have put in the docs that any conversation with a message from the user should be treated as allowed but we could cut out that step more often if we did more for them in the backend.

nplasterer (approved):

PR #481, opened 2023-10-31
Convert static methods to instance methods

rygine:
in this PR:

* converted static methods of `ConsentList` class to instance methods
* added `lastSyncedAt` property to `Contacts` class to track when the consent list was last synced with the network

### DIFF
--- a/src/Contacts.ts
+++ b/src/Contacts.ts
@@ -34,11 +34,13 @@ export class ConsentListEntry {
 }
 
 export class ConsentList {
+  client: Client
   entries: Map<string, ConsentState>
-  static _identifier: string
+  private _identifier: string | undefined
 
-  constructor() {
+  constructor(client: Client) {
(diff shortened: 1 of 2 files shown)


src/Contacts.ts
```diff
@@ -191,16 +209,14 @@ export class Contacts {
   }
 
   async allow(addresses: string[]) {
-    await ConsentList.publish(
-      addresses.map((address) => this.consentList.allow(address)),
-      this.client
+    await this.consentList.publish(
+      addresses.map((address) => this.consentList.allow(address))
```
neekolas, line 213: I know this was the case before, but what happens if the publish fails. It seems like the entries will already be allowed locally.

neekolas (approved):

src/Contacts.ts
rygine: great callout! i just pushed a commit to address this. entries will now be updated after a successful publish.

PR #482, opened 2023-10-31
feat: add self encryption support

rygine:
in this PR:

* upgraded proto dependency
* added `@xmtp/ecies-bindings-wasm` dependency
* added RPC definitions and keystore methods for self encryption/decryption
* added RPC definition and keystore method for retrieving PP topic identifier

these changes are already in the `beta` branch and are being added to `main` in preparation for adding MetaMask Snaps support once the PPPP work lands in `main`. these new keystore methods will be unused until then.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,8 @@
       "license": "MIT",
       "dependencies": {
         "@noble/secp256k1": "^1.5.2",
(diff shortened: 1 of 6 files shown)


rygine:
will wait for official release

PR #490, opened 2023-11-08
Add consent list streaming

rygine:
in this PR:

- added `streamConsentList` method to allow for streaming of new consent entries
- reorganized some code for easier reuse

as the consent messages stream in, the internal consent list is updated.

### DIFF
--- a/src/Contacts.ts
+++ b/src/Contacts.ts
@@ -5,6 +5,8 @@ import {
   buildUserPrivatePreferencesTopic,
   fromNanoString,
 } from './utils'
+import Stream from './Stream'
+import { OnConnectionLostCallback } from './ApiClient'
 
 export type ConsentState = 'allowed' | 'denied' | 'unknown'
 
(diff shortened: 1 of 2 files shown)


nakajima (commented):
I wonder if we should re-emit the conversation in the conversations stream as well for these cases?

rygine:
> I wonder if we should re-emit the conversation in the conversations stream as well for these cases?

interesting thought. i've only ever used the conversation stream to add a conversation to a UI, but curious how this would be used in a backend environment. i suppose it's possible to reference a conversation with an updated consent by reading the `action` message, if necessary.

neekolas:
> I wonder if we should re-emit the conversation in the conversations stream as well for these cases?

It's not a totally crazy idea. I'd personally like that. Most developers think of the conversation stream as a "show me new conversations" stream, so it would require some re-education and updating of docs. Feels like we'd need to treat as a breaking change, even though duplicates in `conversations.stream()` has always been possible.

I'd say we hold off on including it until we're more confident it's what developers really want. Worth bringing up with some of the early adopters of the feature and seeing how they react.

nplasterer (approved):

PR #495, opened 2023-11-28
filter out conversations with invalid topics

### LINKED ISSUE #491: Bug: Conversations with invalid topics should not be included in the conversation list
### Describe the bug

There was a version of the JS SDK that generated invalid non-ascii topics that could cause issues in some clients. If those are found in the conversation list we should filter them out.

### Expected behavior

_No response_

### Steps to reproduce the bug

_No response_

rygine:
resolves #491

### DIFF
--- a/test/utils/topic.test.ts
+++ b/test/utils/topic.test.ts
@@ -1,0 +1,26 @@
+import { buildContentTopic, isValidTopic } from '../../src/utils/topic'
+
+describe('topic utils', () => {
+  describe('isValidTopic', () => {
+    it('validates topics correctly', () => {
+      expect(isValidTopic(buildContentTopic('foo'))).toBe(true)
+      expect(isValidTopic(buildContentTopic('123'))).toBe(true)
+      expect(isValidTopic(buildContentTopic('bar987'))).toBe(true)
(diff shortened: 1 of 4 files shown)


rygine:
unit tests pending, but wanted to confirm logic beforehand.

test/utils/topic.test.ts
```diff
@@ -0,0 +1,26 @@
+import { buildContentTopic, isValidTopic } from '../../src/utils/topic'
+
+describe('topic utils', () => {
+  describe('isValidTopic', () => {
+    it('validates topics correctly', () => {
+      expect(isValidTopic(buildContentTopic('foo'))).toBe(true)
```
neekolas, line 6: Appreciate the robust tests. It might be helpful to just generate a bunch of random topics (the way we do when creating invites) and ensure that those all pass as well.

neekolas (approved):

test/utils/topic.test.ts
rygine: good callout!

added random topic tests here: [test: add random topic testing](https://github.com/xmtp/xmtp-js/pull/495/commits/f3f94afbbc212683a941aae89630762ca3193fdf)

PR #519, opened 2024-01-12
Add `senderHmac` and `shouldPush` to `MessageV2`

rygine:
context: https://community.xmtp.org/t/xip-35-message-sender-identifier-in-topic/509/2

in this PR:

* added `shouldPush` to `ContentCodec` type
* added `shouldPush` method to `TextCodec` and `CompositeCodec`
* added `senderHmac` and `shouldPush` to `MessageV2`
* added encryption methods for generating, importing, and exporting HMAC keys
* added encryption methods for generating and verifying HMAC signatures
* added `getV2ConversationHmacKeys` to `InMemoryKeystore` to retrieve conversation topics and their associated HMAC keys

TODO:

- [x] confirm encryption correctness
- [x] add unit tests

### DIFF
--- a/src/crypto/encryption.ts
+++ b/src/crypto/encryption.ts
@@ -116,3 +116,26 @@ export async function generateHmac(
   const key = await hkdfHmacKey(secret, salt)
   return generateHmacWithKey(key, message)
 }
+
+export async function validateHmac(
+  key: CryptoKey,
+  signature: Uint8Array,
+  message: Uint8Array
+): Promise<boolean> {
+  return await crypto.subtle.verify('HMAC', key, signature, message)
+}
+
+export async function exportHmacKey(key: CryptoKey): Promise<Uint8Array> {
+  const exported = await crypto.subtle.exportKey('raw', key)
+  return new Uint8Array(exported)
+}
+
+export async function importHmacKey(key: Uint8Array): Promise<CryptoKey> {
+  return crypto.subtle.importKey(
+    'raw',
+    key,
+    { name: 'HMAC', hash: 'SHA-256', length: 256 },
+    true,
+    ['sign', 'verify']
+  )
+}
--- a/src/crypto/encryption.ts
+++ b/src/crypto/encryption.ts
@@ -84,7 +84,7 @@ async function hkdf(secret: Uint8Array, salt: Uint8Array): Promise<CryptoKey> {
   )
 }
 
-async function hkdfHmacKey(
+export async function hkdfHmacKey(
   secret: Uint8Array,
   salt: Uint8Array
 ): Promise<CryptoKey> {
--- a/src/crypto/encryption.ts
+++ b/src/crypto/encryption.ts
@@ -100,7 +100,7 @@ export async function hkdfHmacKey(
   )
 }
 
-async function generateHmac(
+async function generateHmacWithKey(
   key: CryptoKey,
   message: Uint8Array
 ): Promise<Uint8Array> {
@@ -108,11 +108,11 @@ async function generateHmac(
   return new Uint8Array(signed)
 }
 
-export async function generateHmacSignature(
+export async function generateHmac(
   secret: Uint8Array,
   salt: Uint8Array,
   message: Uint8Array
 ): Promise<Uint8Array> {
   const key = await hkdfHmacKey(secret, salt)
-  return generateHmac(key, message)
+  return generateHmacWithKey(key, message)
 }
--- a/src/crypto/encryption.ts
+++ b/src/crypto/encryption.ts
@@ -100,24 +100,17 @@ export async function hkdfHmacKey(
   )
 }
 
-async function generateHmacWithKey(
-  key: CryptoKey,
-  message: Uint8Array
-): Promise<Uint8Array> {
-  const signed = await crypto.subtle.sign('HMAC', key, message)
-  return new Uint8Array(signed)
-}
-
-export async function generateHmac(
+export async function generateHmacSignature(
   secret: Uint8Array,
   salt: Uint8Array,
   message: Uint8Array
 ): Promise<Uint8Array> {
   const key = await hkdfHmacKey(secret, salt)
-  return generateHmacWithKey(key, message)
+  const signed = await crypto.subtle.sign('HMAC', key, message)
+  return new Uint8Array(signed)
 }
 
-export async function validateHmac(
+export async function verifyHmacSignature(
   key: CryptoKey,
   signature: Uint8Array,
   message: Uint8Array
--- a/src/crypto/encryption.ts
+++ b/src/crypto/encryption.ts
@@ -3,6 +3,7 @@ import Ciphertext, { AESGCMNonceSize, KDFSaltSize } from './Ciphertext'
 import crypto from './crypto'
 
 const hkdfNoInfo = new ArrayBuffer(0)
+const hkdfNoSalt = new ArrayBuffer(0)
 
 // This is a variation of https://github.com/paulmillr/noble-secp256k1/blob/main/index.ts#L1378-L1388
 // that uses `digest('SHA-256', bytes)` instead of `digest('SHA-256', bytes.buffer)`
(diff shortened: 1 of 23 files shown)


src/keystore/InMemoryKeystore.ts
```diff
@@ -294,10 +295,14 @@ export default class InMemoryKeystore implements Keystore {
           )
         }
 
+        const keyMaterial = getKeyMaterial(topicData.invitation)
```
richardhuaaa, line 298: The changes here currently use the topic key as the input key material, with the topic as a salt. I think we should do one of the following:

1. The source key material for the hkdf (as opposed to the encryption) is per-user, i.e. derived from the private key in the private key bundle. This ensures different users in the topic cannot produce hmac keys for each other. The salt will include the topic, to ensure separation between different topics.

2. If the source key material is derived from the topic key as in the current code, we do not need the topic in the salt, but it needs to include the sender, so that different senders produce different hmac keys. Because all users in the topic know the topic key, they are technically able to forge the hmacs for each other.

(1) is more robust, but (2) is more deniable. We don't have deniability in v2 or v3 anyway, however in v3 it might be more difficult to sync a per-user key, so I lean slightly towards (2). Would love it if @Bren2010 and @neekolas could weigh in here!

src/keystore/InMemoryKeystore.ts
```diff
@@ -294,10 +295,14 @@ export default class InMemoryKeystore implements Keystore {
...
+        const ciphertext = await encryptV2(payload, keyMaterial, headerBytes)
+
         return {
-          encrypted: await encryptV2(
-            payload,
-            getKeyMaterial(topicData.invitation),
+          encrypted: ciphertext,
+          senderHmac: await generateHmacSignature(
+            keyMaterial,
+            new TextEncoder().encode(contentTopic),
```
richardhuaaa, line 305: For the hkdf salt/domain, I would suggest we also include some kind of time-dependent value, such as the number of months since the Unix epoch: https://github.com/xmtp/XIPs/pull/35#issuecomment-1888241997. This ensures that we have automatic key rotation

src/keystore/InMemoryKeystore.ts
```diff
@@ -294,10 +295,14 @@ export default class InMemoryKeystore implements Keystore {
...
             headerBytes
```
richardhuaaa, line 306: I think applying the hmac on the plaintext header bytes would allow for replays, if I'm understanding correctly. Would it make sense to use the ciphertext here?

src/keystore/InMemoryKeystore.ts
neekolas: My worry about the time-dependent value is the complexity of keeping the push server up to date with the correct keys. If we had it roll over at 12am on the first of the month UTC, we'd guarantee that push notifications would not be correctly filtered for some time at the start of each month. Basically would be broken until the app starts up again, which requires integrators to have some sort of wakeup scheduled to refresh keys if they aren't comfortable with extended breakages each month.

The way to fix this is to have each user pass in not just the current month's key but also the next N months keys for each topic. That's fine, but it makes things more complicated from the push server side. Need to handle expirations etc.

src/keystore/InMemoryKeystore.ts
neekolas: The header includes a timestamp which is verified against the envelope timestamp. That should limit the potential for replays. The header bytes are then used as AEAD for the message encryption, so you can't move the headers around without breaking decryption.

src/keystore/InMemoryKeystore.ts
neekolas: So, yes. We need some way of scoping it so that the HMAC only verifies correctly from one sender and not the other.

I like 2 for the compatibility with MLS. We could just include the sender's wallet address into the salt or something.

src/keystore/InMemoryKeystore.ts
neekolas: The reason for using the header is that it's already serialized to bytes. The ciphertext contains has the payload, salt, and nonce. We'd have to pick just the payload, or maybe concatenate all of them. The headerBytes is a single thing to use in the HMAC which leaves less room for bugs or mistakes.

src/keystore/InMemoryKeystore.ts
richardhuaaa: Summary of discussion:
1. We can pass (prev_hmac_key, current_hmac_key, next_hmac_key). That solves any roll over issues.
2. We can expose an API that maps (list of topics to subscribe pushes for) -> (hmac keys for topics). Integrators simply need to call it any time the topic list changes, as well as periodically, and upload it if it changes. This only needs to happen every 30 days at most.

src/crypto/encryption.ts
```diff
@@ -83,3 +83,36 @@ async function hkdf(secret: Uint8Array, salt: Uint8Array): Promise<CryptoKey> {
...
+
+async function generateHmac(
+  key: CryptoKey,
+  message: Uint8Array
+): Promise<Uint8Array> {
+  const signed = await crypto.subtle.sign('HMAC', key, message)
+  return new Uint8Array(signed)
+}
+
+export async function generateHmacSignature(
```
neekolas, line 111: I think we want to tighten up the naming here. A MAC is [similar to a signature](https://dev.to/danielkun/cryptography-what-is-the-difference-between-hashing-signing-and-mac-5dbp), but it's not quite right to call it one.

Maybe we name the function above `generateHmacWithKey` and name this one `generateHmac`?

src/crypto/encryption.ts
```diff
@@ -83,3 +83,36 @@ async function hkdf(secret: Uint8Array, salt: Uint8Array): Promise<CryptoKey> {
...
+  salt: Uint8Array
+): Promise<CryptoKey> {
+  const key = await crypto.subtle.importKey('raw', secret, 'HKDF', false, [
+    'deriveKey',
+  ])
+  return crypto.subtle.deriveKey(
+    { name: 'HKDF', hash: 'SHA-256', salt, info: hkdfNoInfo },
+    key,
+    { name: 'HMAC', hash: 'SHA-256', length: 256 },
+    false,
```
Bren2010, line 98: You have extractable here set to false, but later (successfully?) extract.

src/keystore/InMemoryKeystore.ts
```diff
@@ -294,12 +299,21 @@ export default class InMemoryKeystore implements Keystore {
...
+        const keyMaterial = getKeyMaterial(topicData.invitation)
+        const ciphertext = await encryptV2(payload, keyMaterial, headerBytes)
+        const thirtyDayPeriodsSinceEpoch = Math.floor(
+          Date.now() / 1000 / 60 / 60 / 24 / 30
+        )
+        const info = `${thirtyDayPeriodsSinceEpoch}-${this.accountAddress}`
+        const hmac = await generateHmacSignature(
+          keyMaterial,
+          new TextEncoder().encode(info),
+          headerBytes
```
Bren2010, line 311:
```suggestion
          ciphertext
```
I believe this is right

Bren2010 (approved):
One comment on the content of the HMAC, otherwise this makes sense

src/keystore/InMemoryKeystore.ts
neekolas: So, the ciphertext in this case is an object with a payload, nonces, and a salt. It doesn't get serialized until the whole message gets serialized. 

Could create the hmac on just the payload, or define an order to concat them, but the header bytes seemed the easiest because they're already a byte array. We do have other mitigations to prevent replays of headers.

src/keystore/InMemoryKeystore.ts
```diff
@@ -579,4 +593,44 @@ export default class InMemoryKeystore implements Keystore {
   lookupTopic(topic: string) {
     return this.v2Store.lookup(topic)
   }
+
+  async getV2ConversationHmacKeys(): Promise<keystore.GetConversationHmacKeysResponse> {
```
richardhuaaa, line 597: Will this be exposed to the app developer, and if so, would it make sense to allow them to supply a list of topics to fetch keys for?

rygine: yes, this will be exposed. and yes, that seems like a good option to have.

rygine: added here: [fix: allow specific topics when getting HMAC keys](https://github.com/xmtp/xmtp-js/pull/519/commits/5c59492dcbe5c4bf6bbeb9320551690e53cff3f8)

neekolas: Great improvement

PR #524, opened 2024-01-23
Update Keystore types

rygine:
in this PR:

- created new types for keystore interfaces based on their RPC
- deprecated existing `Keystore` interface

these new types allow for a single source of truth for keystore functionality and their associated types. updating a keystore's RPC definitions will automatically update its types.

i decided to export all the derivative types as well so that integrators have them. they are invaluable for creating type-safe custom keystores and their handlers.

### DIFF
--- a/src/keystore/snapHelpers.ts
+++ b/src/keystore/snapHelpers.ts
@@ -1,5 +1,11 @@
-import { keystore as keystoreProto } from '@xmtp/proto'
-import type { RPC } from './rpcDefinitions'
+import { keystore } from '@xmtp/proto'
+import type {
+  SnapKeystoreApiDefs,
+  SnapKeystoreApiMethods,
+  SnapKeystoreInterfaceRequestValues,
+  SnapKeystoreApiRequestEncoders,
+  SnapKeystoreApiResponseDecoders,
+} from './rpcDefinitions'
 import { b64Decode, b64Encode } from '../utils/bytes'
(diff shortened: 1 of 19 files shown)


src/keystore/snapHelpers.ts
```diff
@@ -24,16 +30,21 @@ type SnapParams = {
...
+  req: SnapKeystoreInterfaceRequestValues[T],
   meta: SnapMeta,
   snapId: string
-): Promise<Res> {
+) {
   let reqParam = null
-  if (codecs.req) {
-    const reqBytes = codecs.req.encode(req).finish()
+  if (rpc.req) {
+    const encoder = rpc.req.encode as SnapKeystoreApiRequestEncoders[T]
```
rygine, line 46: `tsc` needs an assist here to narrow the type

src/keystore/snapHelpers.ts
```diff
@@ -42,20 +53,21 @@ export async function snapRPC<Req, Res>(
     throw new Error('Unexpected array response')
   }
 
-  return codecs.res.decode(b64Decode(responseString))
+  const decoder = rpc.res.decode as SnapKeystoreApiResponseDecoders[T]
```
rygine, line 56: same as above: `tsc` needs an assist here to narrow the type

neekolas (approved):
This is some very impressive Typescript wrangling.

PR #528, opened 2024-02-06
Replace `ethers` with `viem`

rygine:
this PR replaces the use of `ethers` with `viem`. see [Why Viem](https://viem.sh/docs/introduction.html#why-viem) for more info on this change.

### DIFF
--- a/src/crypto/Signature.ts
+++ b/src/crypto/Signature.ts
@@ -4,8 +4,8 @@ import * as secp from '@noble/secp256k1'
 import { PublicKey, UnsignedPublicKey, SignedPublicKey } from './PublicKey'
 import { SignedPrivateKey } from './PrivateKey'
 import { Signer } from '../types/Signer'
-import { bytesToHex, equalBytes, hexToBytes, splitSignature } from './utils'
-import { Hex, hashMessage } from 'viem'
+import { bytesToHex, equalBytes, splitSignature } from './utils'
+import { Hex, hashMessage, hexToBytes } from 'viem'
 
 // ECDSA signature with recovery bit.
 export type ECDSACompactWithRecovery = {
--- a/src/crypto/Signature.ts
+++ b/src/crypto/Signature.ts
@@ -3,9 +3,9 @@ import Long from 'long'
 import * as secp from '@noble/secp256k1'
 import { PublicKey, UnsignedPublicKey, SignedPublicKey } from './PublicKey'
 import { SignedPrivateKey } from './PrivateKey'
-import { utils } from 'ethers'
 import { Signer } from '../types/Signer'
-import { bytesToHex, equalBytes, hexToBytes } from './utils'
+import { bytesToHex, equalBytes, hexToBytes, splitSignature } from './utils'
+import { Hex, hashMessage } from 'viem'
 
 // ECDSA signature with recovery bit.
 export type ECDSACompactWithRecovery = {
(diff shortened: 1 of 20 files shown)


src/crypto/Signature.ts
```diff
@@ -3,9 +3,9 @@ import Long from 'long'
 import * as secp from '@noble/secp256k1'
 import { PublicKey, UnsignedPublicKey, SignedPublicKey } from './PublicKey'
 import { SignedPrivateKey } from './PrivateKey'
-import { utils } from 'ethers'
 import { Signer } from '../types/Signer'
-import { bytesToHex, equalBytes, hexToBytes } from './utils'
+import { bytesToHex, equalBytes, hexToBytes, splitSignature } from './utils'
+import { Hex, hashMessage } from 'viem'
```
neekolas, line 8: FWIW, we do have keccak256 available in Rust-land if you need it. If that's the primary thing you're using Viem for, maybe it makes sense to just move to that and remove the Viem dependency.

neekolas (approved):

src/crypto/Signature.ts
rygine: for now, we're using `viem` for a lot more than just `keccak256`, but definitely good to know we have it in Rust if we move away from `viem` in the future.

src/crypto/utils.ts
```diff
@@ -1,19 +1,15 @@
...
+  getAddress,
+  hexToSignature,
+  keccak256,
+  hexToBytes,
+  bytesToHex as viemBytesToHex,
+} from 'viem'
 
 export const bytesToHex = secp.utils.bytesToHex
 
-export function hexToBytes(s: string): Uint8Array {
```
neekolas, line 5: Given the sensitivity of some of this stuff, I'm going to suggest we do an extra validation test and try and send some messages using this version of the SDK to an older version of the SDK.

Because the tests mostly use new wallets they can hide some sneaky issues around signature verification.

test/conversations/Conversation.test.ts
```diff
@@ -405,19 +403,13 @@ describe('conversation', () => {
...
       await expect(
         alice.conversations.newConversation(bobLower)
       ).resolves.toMatchObject({
         peerAddress: bob.address,
       })
-      await expect(
-        alice.conversations.newConversation(bobUpper)
-      ).resolves.toMatchObject({
-        peerAddress: bob.address,
-      })
```
rygine, line 420: an all uppercase address generates an error with `viem` so removed this test. the `getAddress` function runs through [this bit of code](https://github.com/wevm/viem/blob/main/src/utils/address/isAddress.ts#L21-L35) before returning the checksum address. we could keep the test and update our code to always lowercase an address before generating a checksum address.

src/crypto/utils.ts
rygine: before i start testing client compatibility, i decided to run some tests locally and they passed just fine with a million randomly generated hex values.

```ts
import { equalBytes, oldHexToBytes } from '../../src/crypto/utils'
import { hexToBytes, bytesToHex } from 'viem'
import { randomBytes, webcrypto } from 'node:crypto'

describe('hexToBytes', function () {
  it('old function generates the same as the new function (crypto)', async function () {
    const hexes = Array.from({ length: 1000000 }).map(() =>
      bytesToHex(randomBytes(32))
    )
    hexes.forEach((hex) => {
      expect(equalBytes(oldHexToBytes(hex), hexToBytes(hex)))
    })
  })

  it('old function generates the same as the new function (webcrypto)', async function () {
    const hexes = Array.from({ length: 1000000 }).map(() =>
      bytesToHex(webcrypto.getRandomValues(new Uint8Array(32)))
    )
    hexes.forEach((hex) => {
      expect(equalBytes(oldHexToBytes(hex), hexToBytes(hex)))
    })
  })
})
```

rygine: testing between different client versions seems to work!

Screenshot 2024-02-28 at 11 09 20 AM

neekolas: Appreciate you doing the extra validation here.

neekolas (approved):

PR #531, opened 2024-02-14
Upgrade to node 18.19.0

rygine:
in this PR:

- updated node version manager version to `18.19.0`
- updated node types
- upgraded `checkout` and `setup-node` actions to `v4`
- updated static compressed bytes for a test
- replaced empty `ArrayBuffer` with buffer of empty `Uint8Array`

as of node `18.17.0`...

> Web Crypto API functions' arguments are now coerced and validated as per their WebIDL definitions like in other Web Crypto API implementations. This further improves interoperability with other implementations of Web Crypto API.

link to PR for more context: https://github.com/nodejs/node/pull/46067

with this change, the `crypto.subtle.deriveKey` function started complaining that an empty `ArrayBuffer` was not actually an instance of an `ArrayBuffer`. to fix this error, i simply instantiate an empty `Uint8Array` and use its buffer as the value for the `info` parameter. the values are equivalent (both empty array buffers).


test/Client.test.ts
```diff
@@ -190,10 +190,10 @@ describe('encodeContent', () => {
...
       120, 116, 24, 1, 18, 17, 10, 8, 101, 110, 99, 111, 100, 105, 110, 103, 18,
-      5, 85, 84, 70, 45, 56, 40, 0, 34, 45, 120, 156, 51, 52, 48, 209, 49, 52,
-      48, 4, 98, 11, 8, 54, 52, 212, 49, 54, 2, 82, 150, 96, 166, 161, 161, 9,
-      84, 202, 0, 44, 60, 170, 122, 84, 245, 168, 106, 218, 171, 6, 0, 139, 43,
-      173, 229,
+      5, 85, 84, 70, 45, 56, 40, 0, 34, 48, 120, 156, 51, 52, 48, 209, 49, 52,
+      48, 212, 49, 52, 176, 128, 96, 67, 67, 29, 99, 35, 29, 67, 67, 75, 48,
+      211, 208, 208, 4, 42, 101, 0, 22, 30, 85, 61, 170, 122, 84, 53, 237, 85,
+      3, 0, 139, 43, 173, 229,
     ])
```
rygine, line 197: i'm not sure why this is necessary. i couldn't find anything in the node release notes that may have changed `CompressionStream` or `DecompressionStream`.

neekolas (approved):

src/crypto/encryption.ts
```diff
@@ -2,7 +2,7 @@ import { ciphertext } from '@xmtp/proto'
 import Ciphertext, { AESGCMNonceSize, KDFSaltSize } from './Ciphertext'
 import crypto from './crypto'
 
-const hkdfNoInfo = new ArrayBuffer(0)
+const hkdfNoInfo = new Uint8Array().buffer
```
neekolas, line 5: Weird, but OK

PR #541, opened 2024-02-26
Upgrade some dependencies, remove `js-sha3`

rygine:
# Summary

* Upgraded some dependencies
* Removed pinned versions for most dependencies
* Removed `js-sha3` dependency, used `viem` instead


neekolas (approved):
Love seeing deps go away

PR #547, opened 2024-03-12
Add path aliases, update imports

rygine:
# Summary

* Added path aliases to `tsconfig.json` and rollup/vite configs
* Updated import paths to use aliases
* Removed barrel (`index.ts`) files
* Enabled linting on test files, fixed issues
* Updated ESLint config
* Updated `.eslintignore`

These changes don't affect the exported bundles. No release necessary.


.eslintrc.cjs
```diff
@@ -20,7 +20,17 @@ module.exports = {
         fixMixedExportsWithInlineTypeSpecifier: false,
       },
     ],
+    '@typescript-eslint/no-explicit-any': 'off',
```
neekolas, line 23: This one always annoyed me, since this was a library where there were going to be some explicit anys.

neekolas (approved):

PR #599, opened 2024-04-23
feat: consent proofs

alexrisch:
Updated Protos
Added ability to add consent proofs to conversations Updated types
Added consentProof on Invitiation class

### DIFF
--- a/src/Contacts.ts
+++ b/src/Contacts.ts
@@ -1,4 +1,7 @@
-import { privatePreferences } from '@xmtp/proto'
+import { privatePreferences, type invitation } from '@xmtp/proto'
+import { hashMessage, hexToBytes } from 'viem'
+import { ecdsaSignerKey, WalletSigner } from '@/crypto/Signature'
+import { splitSignature } from '@/crypto/utils'
 import type { EnvelopeWithMessage } from '@/utils/async'
 import { fromNanoString } from '@/utils/date'
 import { buildUserPrivatePreferencesTopic } from '@/utils/topic'
@@ -252,10 +255,60 @@ export class Contacts {
     this.jobRunner = new JobRunner('user-preferences', client.keystore)
   }
 
+  private async validateConsentSignature(
+    signature: `0x${string}`,
+    timestampMs: number,
+    peerAddress: string
+  ): Promise<boolean> {
+    const signatureData = splitSignature(signature)
+    const message = WalletSigner.consentProofRequestText(
+      peerAddress,
+      timestampMs
+    )
+    const digest = hexToBytes(hashMessage(message))
+    // Recover public key
+    const publicKey = ecdsaSignerKey(digest, signatureData)
+    return publicKey?.getEthereumAddress() === this.client.address
+  }
+
+  private async handleConsentProof(
+    consentProof: invitation.ConsentProofPayload,
+    peerAddress: string
+  ): Promise<void> {
+    const { signature, timestamp } = consentProof
+    const isValid = await this.validateConsentSignature(
+      signature as `0x${string}`,
+      Number(timestamp),
+      peerAddress
+    )
+    if (!isValid) {
+      return
+    }
+    await this.client.contacts.allow([peerAddress])
+  }
+
   async loadConsentList(startTime?: Date) {
     return this.jobRunner.run(async (lastRun) => {
       // allow for override of startTime
-      return this.consentList.load(startTime ?? lastRun)
+      const entries = await this.consentList.load(startTime ?? lastRun)
+      try {
+        const conversations = await this.client.conversations.list()
+        conversations.forEach((conversation) => {
+          if (
+            conversation.consentProof &&
+            this.consentState(conversation.peerAddress) === 'unknown'
+          ) {
+            this.handleConsentProof(
+              conversation.consentProof,
+              conversation.peerAddress
+            )
+          }
+        })
+      } catch (err) {
+        console.log(err)
+      }
+
+      return entries
     })
   }
 
--- a/src/Contacts.ts
+++ b/src/Contacts.ts
@@ -260,6 +260,15 @@ export class Contacts {
     timestampMs: number,
     peerAddress: string
   ): Promise<boolean> {
+    if (!signature || !timestampMs) {
+      return false
+    }
+    if (timestampMs > Date.now()) {
+      return false
+    }
+    if (timestampMs < Date.now() - 1000 * 60 * 60 * 24 * 30) {
+      return false
+    }
     const signatureData = splitSignature(signature)
     const message = WalletSigner.consentProofRequestText(
       peerAddress,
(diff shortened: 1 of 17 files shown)


src/conversations/Conversations.ts
```diff
@@ -152,6 +160,43 @@ export default class Conversations<ContentTypes = any> {
     return this.decodeInvites(envelopes)
   }
 
+  private async validateConsentSignature(
```
alexrisch, line 163: Will move this to the signature file instead

src/conversations/Conversations.ts
```diff
@@ -152,6 +160,38 @@ export default class Conversations<ContentTypes = any> {
...
+    signature: `0x${string}`,
+    timestamp: number,
```
neekolas, line 165: What scale is timestamp at? Usually we try and add the `_ns` suffix to avoid confusion. And if it's nanoseconds, you're better off treating it as a Long, since 64 bit ints can overflow a Number and throw weird errors if someone sneaks a big one in.

src/conversations/Conversations.ts
```diff
@@ -152,6 +160,38 @@ export default class Conversations<ContentTypes = any> {
...
+    const { signature, timestamp } = consentProof
+    const isValid = await this.validateConsentSignature(
+      signature as `0x${string}`,
+      Number(timestamp),
+      peerAddress
+    )
+    if (!isValid) {
+      return
+    }
+    const consentState = await this.client.contacts.consentState(peerAddress)
```
neekolas, line 189: Are we sure that this state will always be current? `decodeInvites` can be run many times if the client is using in-memory caching (common with bots), so I wouldn't want to have repeated consent entries come out of this if you call `conversations.list()` several times with different client instances that don't share a cache.

src/conversations/Conversation.ts
```diff
@@ -4,6 +4,7 @@ import {
   type keystore,
   type messageApi,
 } from '@xmtp/proto'
+import type { ConsentProofPayload } from '@xmtp/proto/ts/dist/types/message_contents/invitation.pb'
```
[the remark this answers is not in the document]
alexrisch: Oh yeah that'll cause some issues

src/Contacts.ts
```diff
@@ -252,10 +255,60 @@ export class Contacts {
...
     return this.jobRunner.run(async (lastRun) => {
       // allow for override of startTime
-      return this.consentList.load(startTime ?? lastRun)
+      const entries = await this.consentList.load(startTime ?? lastRun)
+      try {
+        const conversations = await this.client.conversations.list()
+        conversations.forEach((conversation) => {
+          if (
+            conversation.consentProof &&
+            this.consentState(conversation.peerAddress) === 'unknown'
```
neekolas, line 299: This looks good. It'll always have the current value from the network because `this.consentList.load(...)` mutates `this.consentList` as it goes, right?

test/Contacts.test.ts
```diff
@@ -189,4 +191,116 @@ describe('Contacts', () => {
...
+        undefined,
+        consentProofPayload
+      )
+      const conversations = await alix.conversations.list()
+      const convo = conversations.find((c) => c.topic === boConvo.topic)
+      expect(convo).toBeTruthy()
+      await alix.contacts.refreshConsentList()
+      const isAllowed = await alix.contacts.isAllowed(bo.address)
+      expect(isAllowed).toBeFalsy()
+      await alix.close()
```
neekolas, line 302: I didn't even realize we still had a `close()` method on the Client. We should remove this. It's _long_ been deprecated, and does absolutely nothing right now.

neekolas (approved):

src/Contacts.ts
```diff
@@ -252,10 +255,60 @@ export class Contacts {
...
+          ) {
+            this.handleConsentProof(
+              conversation.consentProof,
+              conversation.peerAddress
+            )
+          }
```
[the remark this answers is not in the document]
alexrisch: Agreed, will update

src/Contacts.ts
```diff
@@ -252,10 +255,60 @@ export class Contacts {
...
+    const signatureData = splitSignature(signature)
+    const message = WalletSigner.consentProofRequestText(
+      peerAddress,
+      timestampMs
+    )
+    const digest = hexToBytes(hashMessage(message))
+    // Recover public key
+    const publicKey = ecdsaSignerKey(digest, signatureData)
+    return publicKey?.getEthereumAddress() === this.client.address
+  }
```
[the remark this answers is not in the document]
alexrisch: @rygine I think recent is pretty relative, but it could be within the last month or so, which could also help with a reduction in loading

test/Contacts.test.ts
alexrisch: Removed!

src/Contacts.ts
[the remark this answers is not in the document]
alexrisch: Udapted!

src/Contacts.ts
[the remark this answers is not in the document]
alexrisch: Updated!

src/Contacts.ts
```diff
@@ -260,6 +260,15 @@ export class Contacts {
...
   ): Promise<boolean> {
+    if (!signature || !timestampMs) {
+      return false
+    }
+    if (timestampMs > Date.now()) {
+      return false
+    }
+    if (timestampMs < Date.now() - 1000 * 60 * 60 * 24 * 30) {
+      return false
+    }
```
rygine, line 271: mind adding comments before these checks outlining why they are occurring?

src/Contacts.ts
```diff
@@ -260,6 +260,15 @@ export class Contacts {
     timestampMs: number,
     peerAddress: string
   ): Promise<boolean> {
```
rygine, line 262: is there a specific reason this method is async? looks like it could be sync

src/Contacts.ts
```diff
@@ -252,10 +255,84 @@ export class Contacts {
...
+          Number(proofItem.consentProof.timestamp),
+          proofItem.peerAddress
+        )
+      })
+    )
+    validationResults.forEach((result, index) => {
+      if (result.status === 'fulfilled' && result.value) {
+        validConsentProofAddresses.push(consentProofs[index].peerAddress)
+      }
+    })
```
rygine, line 303: if `validateConsentSignature` becomes sync, i think you can simplify this logic with:

```
    const validConsentProofAddresses = consentProofs.reduce(
      (result, proofItem) => {
        const isValid = this.validateConsentSignature(
          proofItem.consentProof.signature as `0x${string}`,
          Number(proofItem.consentProof.timestamp),
          proofItem.peerAddress
        )
        return isValid ? result.concat(proofItem.peerAddress) : result
      },
      [] as string[]
    )
```

alexrisch: Oh yes good catch, I I'll make it sync

src/Contacts.ts
alexrisch: Nope! good catch!

rygine (approved):

PR #600, opened 2024-05-04
fix: updated consent proof text

alexrisch:
Updated consent proof text to be more human readable


neekolas (approved):
Can't we just use the new micro-library?

alexrisch:
> Can't we just use the new micro-library?

Can do!

PR #613, opened 2024-06-05
Add MLS client

rygine:
# Summary

This PR add an MLS client wrapper around the Node MLS bindings.

# Notes

This is in alpha status and the following is missing and/or incomplete:

* Unit tests
* Code comments
* Types
* Documentation

# Publishing

Merging this won't publish the initial version, which will be handled manually. Once published, a new granular NPM token will be generated and the repository secret will be updated.

# Known issues

The unit tests are passing locally, but are not running in CI. This repo needs some additional tooling to support the updated Docker containers for running a local node.


packages/mls-client/src/AsyncStream.ts
```diff
@@ -0,0 +1,77 @@
...
+  #resolveNext: ResolveNext<T, V> | null
+  #queue: Value<T, V>[]
+  #transformValue?: TransformValue<T, V>
+
+  stopCallback: (() => void) | undefined = undefined
+
+  constructor(
+    transformValue: V extends undefined ? undefined : TransformValue<T, V>
+  ) {
+    this.#queue = []
```
neekolas, line 23: Hash prefixes are what all the cool kids in javascript are doing now?

fabriguespe (approved):

packages/mls-client/rollup.config.js
```diff
@@ -0,0 +1,59 @@
...
+  }),
+  filesize({
+    showMinifiedSize: false,
+  }),
+  json({
+    preferConst: true,
+  }),
+]
+
+export default defineConfig([
```
fabriguespe, line 30: this declaration is very cool. You think we can do something like this inside the bot package for declaring ts and build scripts

packages/mls-client/src/AsyncStream.ts
rygine: haha. it's updated syntax for [private class properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties).

packages/mls-client/src/Client.ts
```diff
@@ -0,0 +1,184 @@
...
+      await createClient(
+        host,
+        isSecure,
+        dbPath,
+        accountAddress,
+        options?.encryptionKey
+      ),
+      [
+        new MembershipChangeCodec(),
+        new GroupUpdatedCodec(),
```
neekolas, line 101: One thing to keep in mind is that `GroupUpdated` messages are illegal unless they have a `kind` of `membership_change`

packages/mls-client/src/AsyncStream.ts
rygine: `private` is a TypeScript only modifier. this new syntax is part of the JavaScript language.

packages/mls-client/src/Client.ts
neekolas: We should hide any that have a different kind at the SDK level so nobody gets bamboozled by a fake `GroupUpdated` message. I believe the mobile SDKs already do this.

rygine: i was thinking that we should prevent manually sending `MembershipChange` and `GroupUpdated` messages. i only added the encoding to match the codec signature.

WDYT?

packages/mls-client/src/codecs/MembershipChangeCodec.ts
```diff
@@ -0,0 +1,41 @@
+import { mlsTranscriptMessages } from '@xmtp/proto'
```
neekolas, line 1: Sorry about the wasted cycles. This codec is dead and we just haven't deleted it from the other repos. Fully replaced with `GroupUpdated`

packages/mls-client/src/Client.ts
neekolas: I care less about blocking sends because it doesn't change the security (a motivated attacker will just fork the package and add it back).

We could have a warning or something when we detect it. But I can see having the functionality work being useful for writing tests.

packages/mls-client/src/codecs/MembershipChangeCodec.ts
rygine: no worries, was mainly a copy/paste.

packages/mls-client/src/Client.ts
rygine: added check for invalid message in this commit: [Check for invalid group membership change messages](https://github.com/xmtp/xmtp-js/pull/613/commits/695998efec9d60e087705da9f5224a98d22f05ea)

copied [the way the iOS SDK does it](https://github.com/xmtp/xmtp-ios/blob/8f0186edc7dfd5f482725e8a3eb14c26a4173ce5/Sources/XMTPiOS/Mls/MessageV3.swift#L65-L67)

packages/mls-client/src/AsyncStream.ts
neekolas: Ahhh. Yet another reminder that Typescript only lives in our hearts and minds, but definitely not the JS engine.

PR #625, opened 2024-06-13
Add streaming callbacks

rygine:
# Summary

* Simplified `AsyncStream` types, removed value transformer
* Added `isDone` getter to `AsyncStream`
* Added optional callback parameter to all streaming functions for more flexibility in handling streams
* Added stream data transforms before calling stream callback
* Added conversations map and `get` function to `Conversations` to allow for quick access to conversations that are created, listed, or streamed during a client session
* Replaced usage of `@xmtp/xmtp-js` in favor of `@xmtp/content-type-primitives`

### DIFF
--- a/packages/mls-client/package.json
+++ b/packages/mls-client/package.json
@@ -52,9 +52,9 @@
     "typecheck": "tsc"
   },
   "dependencies": {
+    "@xmtp/content-type-primitives": "^1.0.1",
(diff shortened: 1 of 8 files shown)


fabriguespe (approved):
Thanks!

5twelve (approved):

PR #627, opened 2024-06-14
Add production environment, update content types

rygine:
# Summary

* Added `production` environment option when creating a client
* Removed dependency on `@xmtp/xmtp-js` for content types

PR #629, opened 2024-06-14
Update content type dependencies and imports

rygine:
# Summary

* Removed dependency on `@xmtp/xmtp-js` for content types


rygine:
closing in favor of https://github.com/xmtp/xmtp-js/pull/627

PR #633, opened 2024-06-15
Optimize CI workflows

rygine:
# Summary

Since the JS SDK and the MLS Client are separate packages, only run their respective workflows when their files change.

This PR also upgrades to Warp runners where possible. For some reason, there are fetch issues in some of the JS SDK tests on Warp runners so will use GitHub runners for testing.


alexrisch (approved):

.github/workflows/js-sdk.yml
```diff
@@ -0,0 +1,144 @@
+name: JS SDK
+
+on:
+  push:
+    branches:
+      - main
+
+  pull_request:
+    paths:
```
neekolas, line 9: Did not know this option was a thing

neekolas (approved):

.github/workflows/js-sdk.yml
rygine: i'm using this for the node bindings in `libxmtp` as well. i have some more optimizations in mind for `libxmtp` to help speed up the workflows using `paths`.

PR #638, opened 2024-06-24
Upgraded MLS bindings, added getters

rygine:
# Summary

* Upgraded to latest MLS node bindings
* Added `requestHistorySync` and `getInboxIdByAddress` to `Client`
* Renamed `get` to `getConversationById` in `Conversations`
* Added `getMessageById` to `Conversations`
* Removed internal group referencing
* Added/updated tests

### DIFF
--- a/.changeset/rare-pants-promise.md
+++ b/.changeset/rare-pants-promise.md
@@ -1,0 +1,8 @@
+---
+"@xmtp/mls-client": patch
+---
+
+- Upgraded to latest MLS node bindings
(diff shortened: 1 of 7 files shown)


neekolas (approved):

fabriguespe:
Tested updates:

- Removed stream at the init of the runner:

```
// start streaming groups so that they are added to the internal mapping
//const groupStream = client.conversations.stream();
```
 
- Now using `getMessageById` for getting the receiver of reaction and reply. No need for extra content type field type.
- `getConversationById` tested and working well

Thanks!

PR #677, opened 2024-10-10
Add sdks workspace, update README

rygine:
# Summary

- Added `sdks` workspace, move SDKs into it
- Update README
- Removed deprecated `launch.json`
- Removed unused setting in `settings.json`


package.json
```diff
@@ -5,6 +5,7 @@
   "workspaces": [
     "content-types/*",
     "packages/*",
```
alexrisch, line 7: Is packages/* needed anymore?

alexrisch (approved):

package.json
rygine: good callout, yep! i'm going to migrate some packages from other repos into this one and use the `packages` workspace for them.

PR #749, opened 2024-12-10
Remove support for CommonJS

rygine:
# Summary

This PR removes CommonJS exports from all packages (except the legacy JS SDK) and cleans up some `package.json` files.

### DIFF
--- a/.changeset/slow-peas-report.md
+++ b/.changeset/slow-peas-report.md
@@ -1,0 +1,17 @@
+---
+"@xmtp/content-type-transaction-reference": major
+"@xmtp/content-type-remote-attachment": major
+"@xmtp/content-type-group-updated": major
+"@xmtp/content-type-read-receipt": major
+"@xmtp/content-type-primitives": major
+"@xmtp/content-type-reaction": major
+"@xmtp/content-type-reply": major
+"@xmtp/consent-proof-signature": major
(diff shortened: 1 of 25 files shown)


insipx (approved):

malaDev:
is there any specific reason for this removal?

rygine:
> is there any specific reason for this removal?

this change is intended to simplify and streamline our build process for all of our SDKs. it also resolves an import issue with our node bindings, which are generated in ESM and not available in CJS.

since `require()` now supports loading (synchronous) ES modules in node 20 (experimental flag), 22 (latest), and 23, this _shouldn't_ be a problem for developers moving forward.

if you're having a specific issue, please let me know!

anton-prod:
Workaround for those who need to use CommonJS, like in Nest.js projects:

```typescript
// Do not import/require the package directly - use this loader: 
const loadXmtpAgentModule = () =>
  new Function('modulePath', 'return import(modulePath)')(
    '@xmtp/agent-sdk',
  )
  
export class XmtpService implements OnModuleInit {
  private agent: any // Unfortunately, I was not able to get the TypeScript types working - the `.on()` method never appears as part of the agent instance

  async onModuleInit() {
    const { Agent } = await loadXmtpAgentModule()
    this.agent = await Agent.createFromEnv()
  }
}
```

PR #793, opened 2025-01-19
feat(backup): Implement random topic and encryption key generation

crStiv:
- Add secure random bytes generation using crypto.getRandomValues()
- Create unique topic identifier with wallet address, timestamp and random id
- Store encryption key in secret field as Uint8Array
- Remove TODO comment as the feature is now implemented

### DIFF
--- a/sdks/js-sdk/src/message-backup/TopicStoreBackupClient.ts
+++ b/sdks/js-sdk/src/message-backup/TopicStoreBackupClient.ts
@@ -1,3 +1,7 @@
+/// <reference lib="dom" />
+/// <reference lib="es2017" />
+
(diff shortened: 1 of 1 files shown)


nplasterer:
👋 Hi @crStiv thanks for the contribution our `js-sdk` is deprecated in favor of our `browser-sdk` https://github.com/xmtp/xmtp-js/tree/main/sdks/browser-sdk. If you'd like to make contributions to the protocol please do so there. Let me know if it was not clear from the readme in js-sdk folder we can try to make that more clear https://github.com/xmtp/xmtp-js/tree/main/sdks/js-sdk

PR #853, opened 2025-03-02
Browser + Node SDK updates

rygine:
# Summary

- Exposed all client signature methods on both SDKs
- Refactored Node SDK client signature methods to return `undefined` instead of `null` to match the Browser SDK
- Added guard to `Client.addAccount` methods to prevent automatic reassignment of inboxes
- Renamed `syncAdmins` to `listAdmins` and `syncSuperAdmins` to `listSuperAdmins` in the Browser SDK

These updates will not be included in a release at this time.

### DIFF
--- a/sdks/browser-sdk/src/Client.ts
+++ b/sdks/browser-sdk/src/Client.ts
@@ -182,7 +182,29 @@ export class Client extends ClientWorkerClass {
     return this.sendMessage("registerIdentity", undefined);
   }
 
-  async addAccount(newAccountSigner: Signer) {
+  /**
+   * This function should be used with caution. Adding a wallet already
+   * associated with an inboxId will cause the wallet to lose access to
+   * that inbox.
+   *
+   * The `allowInboxReassign` parameter must be true to reassign an inbox
+   * already associated with a different account.
+   */
+  async addAccount(
+    newAccountSigner: Signer,
+    allowInboxReassign: boolean = false,
+  ) {
+    // check for existing inbox id
+    const existingInboxId = await this.findInboxIdByAddress(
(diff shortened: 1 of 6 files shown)


sdks/browser-sdk/src/Client.ts
```diff
@@ -112,34 +112,34 @@ export class Client extends ClientWorkerClass {
...
   }
 
-  async #createInboxSignatureText() {
+  async createInboxSignatureText() {
     return this.sendMessage("createInboxSignatureText", undefined);
   }
 
-  async #addAccountSignatureText(newAccountAddress: string) {
+  async addAccountSignatureText(newAccountAddress: string) {
     return this.sendMessage("addAccountSignatureText", {
```
nplasterer, line 120: Since this is public now I'd also add the allow reassignment Boolean here as well we really don't want people accidentally rotating inbox ids.

nplasterer (approved):
Wondering if there is a way to mark all this private functions that are now public as delicate or warn? Just don't want to cause extra confusion by exposing them. 🤔

rygine:
> Wondering if there is a way to mark all this private functions that are now public as delicate or warn? Just don't want to cause extra confusion by exposing them. 🤔

i think if we add appropriate code comments and docs, we should be good. but, if we really want to drive this point home, we should prefix these function names with something that indicates caution should be taken.

sdks/browser-sdk/src/Client.ts
rygine: thinking about this, i'm not sure we need that type of guard here since it's just returning a string and not actually doing anything. WDYT?

PR #866, opened 2025-03-11
Add identifiers to Node SDK

rygine:
# Summary

- Upgraded to latest node bindings
- Updated `Signer` type
- Replaced address parameters with inbox IDs or identifiers
- Added new methods to use with identifiers
- Added `pausedForVersion` to `Conversation`
- Updated exports

This update will not include a release. Release to follow shortly in a separate PR.

### DIFF
--- a/sdks/node-sdk/src/Group.ts
+++ b/sdks/node-sdk/src/Group.ts
@@ -1,4 +1,5 @@
 import type {
+  Identifier,
   Message,
   MetadataField,
   PermissionPolicy,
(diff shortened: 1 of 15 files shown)


insipx (approved):

sdks/node-sdk/src/Group.ts
```diff
@@ -80,16 +81,16 @@ export class Group extends Conversation {
     return this.#conversation.isSuperAdmin(inboxId);
   }
 
-  async addMembers(accountAddresses: string[]) {
-    return this.#conversation.addMembers(accountAddresses);
+  async addMembers(identifiers: Identifier[]) {
+    return this.#conversation.addMembers(identifiers);
   }
 
   async addMembersByInboxId(inboxIds: string[]) {
```
nplasterer, line 88: You forgot to swap these. addMembers and removeMembers should be inboxId and the other type should take an identifier. Can you address in a follow up. 🙏 @rygine

PR #869, opened 2025-03-12
Add identifiers to Browser SDK

rygine:
# Summary

- Upgraded to latest WASM bindings
- Updated `Signer` type
- Replaced address parameters with inbox IDs or identifiers
- Added new methods to use with identifiers
- Added `pausedForVersion` to `Conversation`
- Updated exports


nplasterer (approved):
LGTM!

PR #921, opened 2025-04-02
Add WalletSendCalls custom type and handle transaction triggering

justjkk:
This PR adds WalletSendCalls custom type proposed in [XIP 59](https://community.xmtp.org/t/xip-59-trigger-on-chain-calls-via-wallet-sendcalls/889).

The following changes are made to xmtp.chat code as well:
* Handle custom types `WalletSendCalls` and `TransactionReference` ([XIP 21](https://community.xmtp.org/t/xip-21-on-chain-transaction-reference-content-type/532) and UI for them.
* Disable click handlers on message text as this behavior conflicts with the transaction confirmation button and explorer url link UI elements.


apps/xmtp.chat/src/components/Messages/Message.tsx
```diff
@@ -1,44 +1,35 @@
...
+import { WalletSendCallsUI } from "./WalletSendCalls";
 
 export type MessageProps = {
   message: DecodedMessage;
+  sendMessage: Conversation["send"];
 };
 
-export const Message: React.FC<MessageProps> = ({ message }) => {
+export const Message: React.FC<MessageProps> = ({ message, sendMessage }) => {
   const { client } = useOutletContext<{ client: Client }>();
```
rygine, line 25: it's no longer necessary to pass down the `sendMessage` method. you should be able to access the conversation directly from the outlet context like so:

```
const { client, conversation } = useOutletContext<ConversationOutletContext>();
```

content-types/content-type-wallet-send-calls/package.json
```diff
@@ -0,0 +1,86 @@
+{
+  "name": "@xmtp/content-type-wallet-send-calls",
+  "version": "0.0.2",
```
rygine, line 3: why not `0.0.1`?

apps/xmtp.chat/src/components/Messages/TransactionReference.tsx
rygine, line 1: would you mind suffixing this with `Content` for the file name and the export?

`TransactionReferenceContent.tsx`

```
export const TransactionReferenceContent = ...
```

i'd like to have this convention for all custom message content.

apps/xmtp.chat/src/components/Messages/Message.tsx
```diff
@@ -1,44 +1,35 @@
...
-                `/conversations/${message.conversationId}/message/${message.id}`,
-              );
-            }
-          }}
-          className={classes.root}
-          onClick={() =>
-            void navigate(
-              `/conversations/${message.conversationId}/message/${message.id}`,
-            )
-          }>
```
rygine, line 41: this removes a key feature of the dev tool and needs to be preserved. i believe you should be able to prevent the click event propagation of the links in the content to not trigger the modal.

apps/xmtp.chat/src/components/Messages/WalletSendCalls.tsx
```diff
@@ -0,0 +1,57 @@
...
+        await sendMessage(
+          transactionReference,
+          ContentTypeTransactionReference,
+        );
+      }
+    })();
+  }
+
+  return (
+    <div>
```
rygine, line 48: please use Mantine components for rendering content.

`div` => `Box`
`br` => `Divider`
`p` => `Text`

this ensure the theme is applied consistently across the app

content-types/content-type-wallet-send-calls/package.json
justjkk: It was 0.0.1 earlier but as I was back-merging main I saw `content-type-transaction-reference` package version was [changed](https://github.com/HeyElsa/xmtp-js/commit/3e280789a7ee11f3af4497e848e2c9f3f6cb4aec#diff-d526eb570c5023cd812bb8b23a0dd1342cb49450fdadf5bed094a60b1b6de33cR3) from 2.0.0 to 2.0.1 and since this was copied mostly from that content-type I bumped the package version for this as well. Should this be 2.0.1 as well to be consistent with all the other content-types?

apps/xmtp.chat/src/components/Messages/WalletSendCalls.tsx
```diff
@@ -0,0 +1,57 @@
...
+  sendMessage: Conversation["send"];
+};
+
+export const WalletSendCallsUI: React.FC<WalletSendCallsProps> = ({
+  content,
+  sendMessage,
+}) => {
+  const { sendTransactionAsync } = useSendTransaction();
+
+  function handleSubmit() {
```
rygine, line 21: can you set this up as:

```
const handleSubmit = useCallback(async () => {...}, [...]);
```

this follows convention with other `async` callbacks in this app.

apps/xmtp.chat/src/components/Messages/TransactionReference.tsx
```diff
@@ -0,0 +1,36 @@
...
+import * as viemChains from "viem/chains";
+
+export type TransactionReferenceProps = {
+  content: TransactionReference;
+};
+
+export const TransactionReferenceUI: React.FC<TransactionReferenceProps> = ({
+  content,
+}) => {
+  const chains = Object.values(viemChains);
```
rygine, line 12: this should exist outside of the rendering function as it only needs to be calculated once and doesn't depending on any of the component's props/data.

apps/xmtp.chat/src/components/Messages/TransactionReference.tsx
```diff
@@ -0,0 +1,36 @@
...
+      : content.networkId;
+  const chain = chains.find((chain) => chain.id === chainId);
+  if (!chain) {
+    return (
+      <div>
+        <span>Chain Id: {content.networkId}</span>
+        <br />
+        <span>Transaction Hash: {content.reference}</span>
+        <br />
+      </div>
```
rygine, line 25: see above comment on using Mantine components

apps/xmtp.chat/src/components/Messages/TransactionReference.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export const TransactionReferenceUI: React.FC<TransactionReferenceProps> = ({
+  content,
+}) => {
+  const chains = Object.values(viemChains);
+  const chainId =
+    typeof content.networkId === "string"
+      ? parseInt(content.networkId, 16)
+      : content.networkId;
+  const chain = chains.find((chain) => chain.id === chainId);
```
rygine, line 17: this value can be wrapped in a `useMemo` like so:

```
const chain = useMemo(() => {
  const chainId =
    typeof content.networkId === "string"
      ? parseInt(content.networkId, 16)
      : content.networkId;
  return chains.find((chain) => chain.id === chainId);
}, [content.networkId]);
```

apps/xmtp.chat/src/components/Messages/Message.tsx
justjkk: ok, will try it out.

apps/xmtp.chat/src/components/Messages/TransactionReference.tsx
justjkk: Moved it to inside useMemo. Is this fine?

apps/xmtp.chat/src/components/Messages/WalletSendCalls.tsx
justjkk: Tried the above syntax but that triggers a eslint error about no-misused-promises. `Promise-returning function provided to attribute where a void return was expected.`. Updated the code to use `useCallback` and using [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) pattern inside. Please check if this is fine.

rygine: in order for ESLint to be happy, you need to update the the `onClick` handler to this:

```
<Button
  onClick={() => {
    void handleSubmit();
  }}>
  Submit
</Button>
```

humanagent (approved):

humanagent:
Tested successfully with 5 transactions in a row using CB wallet extension and Metamask

rygine (approved):

humanagent (approved):

PR #1022, opened 2025-05-14
Add seconds to date

humanagent:
Add second indicator for easily share and debug response times

CleanShot 2025-05-14 at 14 09 19@2x

CleanShot 2025-05-14 at 14 09 15@2x

### DIFF
--- a/apps/xmtp.chat/src/components/DateLabel.tsx
+++ b/apps/xmtp.chat/src/components/DateLabel.tsx
@@ -47,6 +47,20 @@ export const DateLabel: React.FC<DateLabelProps> = ({
     }
   };
 
+  // Format date and time parts separately
+  const dateFormatted = intlFormat(date, {
+    year: "numeric",
+    month: "2-digit",
+    day: "2-digit",
+  });
+
+  const timeFormatted = intlFormat(date, {
+    hour: "2-digit",
+    minute: "2-digit",
+  });
+
+  const seconds = date.getSeconds().toString().padStart(2, "0");
+
   return (
     <Tooltip
       label={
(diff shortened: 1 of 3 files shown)


apps/xmtp.chat/src/components/DateLabel.tsx
```diff
@@ -66,13 +80,10 @@ export const DateLabel: React.FC<DateLabelProps> = ({
         onClick={handleCopy}
         miw={100}
         tabIndex={0}>
-        {intlFormat(date, {
-          year: "numeric",
-          month: "2-digit",
-          day: "2-digit",
-          hour: "2-digit",
-          minute: "2-digit",
```
rygine, line 74: why not just add `second: "2-digit"` here?

apps/xmtp.chat/src/helpers/date.ts
```diff
@@ -1,3 +1,6 @@
 export function nsToDate(ns: bigint): Date {
-  return new Date(Number(ns / 1_000_000n));
+  const date = new Date(Number(ns / 1_000_000n));
+  const seconds = date.getSeconds();
+  date.setSeconds(seconds);
+  return new Date(date);
```
rygine, line 5: i don't understand this change

apps/xmtp.chat/src/components/DateLabel.tsx
humanagent: Wanted to make the have the `s` indicator so i was trying with a separate tag


        <Text component="span" ml={4} c="dimmed">
          {date.getSeconds().toString().padStart(2, "0")}s
        </Text>
        
        
        
        But that should be enough too

apps/xmtp.chat/src/helpers/date.ts
```diff
@@ -1,3 +1,4 @@
 export function nsToDate(ns: bigint): Date {
-  return new Date(Number(ns / 1_000_000n));
+  const date = new Date(Number(ns / 1_000_000n));
+  return new Date(date);
```
rygine: this is the same as before, just undo the changes in this file

rygine (approved):
LGTM

something to note. these changes won't be deployed on `xmtp.chat` until we update the bindings.

PR #1073, opened 2025-07-01
Add installation management without a client

rygine:
# Summary

- Refactored wallet and XMTP connection management
- Added a new page to `xmtp.chat` for managing installations without a client


cameronvoell (approved):
This is working great for me, well done. I was able to use it to fix one of my old accounts with too many installations.  🎉 

Some ideas for future improvements (no blockers):

1. Would be great if you could see the error status of key packages from the manage identity page (same as the messenger Identity page)
2. Would be nice if the installation management autopopulated with the ethereum address of your connected account (still with the option to search other accounts to see status)
3. Didnt immediately understand the connection flow on the new start page for messenger. Maybe adding "Step 1" before connect your wallet, and "Step 2" before connect to XMTP, maybe with some green checkmark somewhere when the step 1 is completed. (I saw browser injected with a checkmark but didnt immediately see that I needed to click connect also before I could connect to XMTP.)

PR #1118, opened 2025-07-28
Refactor streaming methods (Node SDK)

rygine:
# Summary

This PR contains **BREAKING CHANGES** for all streaming methods in the Node SDK.

### Single parameter for streaming methods

All streaming methods now accept a single parameter that defines streaming options. These new options add more callbacks and options for retrying failed streams.

```ts
type StreamOptions<T = unknown, V = T> = {
  /**
   * Called when the stream ends
   */
  onEnd?: () => void;
  /**
   * Called when a stream error occurs
   */
  onError?: (error: Error) => void;
  /**
   * Called when the stream fails
   */
  onFail?: () => void;
  /**
   * Called when the stream is restarted
   */
  onRestart?: () => void;
  /**
   * Called when the stream is retried
   */
  onRetry?: (attempts: number, maxAttempts: number) => void;
  /**
   * Called when a value is emitted from the stream
   */
  onValue?: (value: V) => void;
  /**
   * The number of times to retry the stream
   * (default: 6)
   */
  retryAttempts?: number;
  /**
   * The delay between retries (in milliseconds)
   * (default: 10000)
   */
  retryDelay?: number;
  /**
   * Whether to retry the stream if it fails
   * (default: true)
   */
  retryOnFail?: boolean;
}
```

In addition to these options, some streaming methods have more options. See their respective types for more details.

### Streams no longer end on error

When a stream error occurs, it's passed to the `onError` callback only. The stream will remain active.

### Stream types have changed

When using the `for await..of` loop, the value will never be `undefined`.

### DIFF
--- a/sdks/node-sdk/src/utils/streams.ts
+++ b/sdks/node-sdk/src/utils/streams.ts
@@ -107,16 +107,20 @@ export const createStream = async <T = unknown, V = T>(
       return;
     }
     // ensure the value is not undefined
-    if (value) {
+    if (value !== undefined) {
       try {
         // if a streamValueMutator is provided, mutate the value
         if (streamValueMutator) {
           const mutatedValue = streamValueMutator(value);
           if (isPromise(mutatedValue)) {
-            void mutatedValue.then((mutatedValue) => {
-              asyncStream.push(mutatedValue);
-              onValue?.(mutatedValue);
-            });
+            void mutatedValue
+              .then((mutatedValue) => {
+                asyncStream.push(mutatedValue);
+                onValue?.(mutatedValue);
+              })
+              .catch((error: unknown) => {
+                onError?.(error as Error);
+              });
           } else {
             asyncStream.push(mutatedValue);
             onValue?.(mutatedValue);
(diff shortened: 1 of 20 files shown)


sdks/node-sdk/src/utils/streams.ts
```diff
@@ -0,0 +1,199 @@
...
+        } else {
+          asyncStream.push(value as unknown as V);
+          onValue?.(value as unknown as V);
+        }
+      } catch (error) {
+        onError?.(error as Error);
+      }
+    }
+  };
+  const retry = async (retries: number = retryAttempts) => {
```
cameronvoell, line 137: does this have the issue where if a retry happens while there is no connection, the stream starts without detecting an issue even though it is not connected? Or is it attempting to do a sync somewhere upstream, I think you might have mentioned, for determining if we're online to get around that issue? 

I ask because I was seeing that problem when implementing retry for streams in react native, if I retried while connection was down, it thought it was working and it would never flag the stream ending.

rygine: the node SDK implementation doesn't have this issue because the stream creation methods sync to the network, which will throw if the node is still down.

cameronvoell (approved):

PR #1136, opened 2025-08-13
Send replies

bennycode:


bennycode:
@rygine I added `key={conversation.id}` to `ConversationProvider` so its subtree re-renders when the conversation changes. This resets the message preview and reply preview when switching conversations. In the future, we might replace this with proper state management to preserve message input per conversation, but for now, using the `key` improves the UX as without it, preview text carries over between conversations.

I also removed the optional modifier from `conversation` in the `useConversation` hook, since it always receives a conversation. This eliminates the need for conversation existence checks.

Additionally, I moved tracking of the reply message target into `useConversationContext`, avoiding drilling it as prop across components. Ideally, this should live in `ConversationContext` since it’s per-conversation state. I’m thinking of making that change when we introduce state management with [Zustand](https://github.com/pmndrs/zustand). WDYT?

PR #1137, opened 2025-08-15
Support sending reactions

bennycode:


apps/xmtp.chat/src/helpers/strings.ts
```diff
@@ -7,3 +9,17 @@ export const isValidInboxId = (inboxId: string): inboxId is string =>
 
 export const shortAddress = (address: string): string =>
   `${address.substring(0, 6)}...${address.substring(address.length - 4)}`;
+
+export const stringifyMessage = (message: DecodedMessage) => {
+  if ("content" in message) {
```
rygine: this should always be true

apps/xmtp.chat/src/helpers/strings.ts
```diff
@@ -7,3 +9,17 @@ export const isValidInboxId = (inboxId: string): inboxId is string =>
...
+    if (
+      message.content &&
+      typeof message.content === "object" &&
+      "content" in message.content
+    ) {
+      return String(message.content.content);
+    }
+    return String(message.content);
+  }
```
rygine: if `message.content` is `undefined`, that means its codec is not available. in that case, check for a `fallback` value to display. if there's no `fallback`, use `JSON.stringify(content, null, 2)` to display the content, like we do in `MessageContent.tsx`.

if `message.content` is defined, we know that it's been decoded. in this case, use  the `sameAs` method from the message's `contentType` field to determine how to render the content.

apps/xmtp.chat/src/helpers/strings.ts
bennycode: The value of `content` is typed as `content: ContentTypes | undefined`, so I have to check for it being defined. Also the `fallback` is optional (`fallback?: string`) but I will make sure to use it before falling back to pure strings.

rygine: yep, that makes sense, just saying that the `content` property will always exist even if it's `undefined`.

apps/xmtp.chat/src/helpers/messages.ts
```diff
@@ -0,0 +1,39 @@
...
+import {
+  ContentTypeReaction,
+  type Reaction,
+} from "@xmtp/content-type-reaction";
+import { ContentTypeReply, type Reply } from "@xmtp/content-type-reply";
+import { ContentTypeText } from "@xmtp/content-type-text";
+
+export const isReaction = (
+  m: DecodedMessage,
+): m is DecodedMessage & { content: Reaction } =>
```
rygine: this can just be `m is DecodedMessage<Reaction>`

apps/xmtp.chat/src/helpers/messages.ts
```diff
@@ -0,0 +1,39 @@
...
+  m.contentType.sameAs(ContentTypeReaction);
+
+export const isReply = (
+  m: DecodedMessage,
+): m is DecodedMessage & { content: Reply } =>
+  m.contentType.sameAs(ContentTypeReply);
+
+export const isText = (
+  m: DecodedMessage,
+): m is DecodedMessage & { content: string } =>
```
rygine: same as above for these 2

apps/xmtp.chat/src/helpers/messages.ts
```diff
@@ -0,0 +1,39 @@
...
+  m.contentType.sameAs(ContentTypeText);
+
+export const stringify = (message: DecodedMessage): string => {
+  switch (true) {
+    case isReaction(message):
+      return message.content.content;
+    case isReply(message):
+      return String(message.content.content);
```
rygine: since replies can be any content type, perhaps creating another util like `isTextReply` would be more useful here. if it's not a text reply, then use `JSON.stringify` like the `default` option.

apps/xmtp.chat/src/helpers/messages.ts
bennycode: I use the intersection type `{ content: Reaction }` to make `content` non-nullable, otherwise it can be `undefined` as `content` is typed as `content: ContentTypes | undefined` in the `DecodedMessage` class.

Problem with using `m is DecodedMessage<Reaction>`:

image

apps/xmtp.chat/src/helpers/messages.ts
bennycode: Good idea! In the future I may even implement an unwrapping function with a max-depths for reply messages.

apps/xmtp.chat/src/helpers/messages.ts
```diff
@@ -0,0 +1,43 @@
...
+import { ContentTypeText } from "@xmtp/content-type-text";
+
+export const isReaction = (
+  m: DecodedMessage,
+): m is DecodedMessage & { content: Reaction } =>
+  m.contentType.sameAs(ContentTypeReaction);
+
+export const isTextReply = (
+  m: DecodedMessage,
+): m is DecodedMessage & { content: string } => {
```
rygine: wouldn't this be `& { content: { content: string } }` or similar?

unfortunately, the `Reply` type doesn't have a generic. we can add that easily if it makes this easier.

bennycode: **Great catch!** 

It's `message.content.content` for a decoded `Reply`. Introduced `isReply` in combination with `isReplyText` that predicts `{ content: Reply & { content: string } }` to follow a consistent typing style.

Time to add some unit tests I guess! 🧪

rygine (approved):

PR #1145, opened 2025-08-27
Add remote attachment support to xmtp.chat

rygine:


apps/xmtp.chat/src/components/Conversation/AttachmentPreview.tsx
```diff
@@ -0,0 +1,89 @@
+import { ActionIcon, Box, Group, Paper, Text } from "@mantine/core";
+import { useMemo } from "react";
```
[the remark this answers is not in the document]
rygine: i'm specifically not revoking them so they're cached for a better user experience

apps/xmtp.chat/src/components/Conversation/ReplyPreview.tsx
```diff
@@ -1,54 +1,92 @@
...
+      <AttachmentDetails
+        filename={message.content.filename}
+        fileSize={formatFileSize(message.content.contentLength)}
+        align="left"
+      />
+    );
+  }
+  let textContent: string | undefined;
+  if (isTextReply(message) || isReaction(message)) {
+    textContent = message.content.content;
```
bennycode: Can't we use the `stringify` method here from `messages.ts`?

apps/xmtp.chat/src/components/Messages/Message.tsx
```diff
@@ -23,7 +28,10 @@ export const Message: React.FC<MessageProps> = ({
   const isSender = client.inboxId === message.senderInboxId;
   const align = isSender ? "right" : "left";
   const isActionable =
```
bennycode: How about we also put an `isActionable(message)` in the `helpers/messages.ts` file?

apps/xmtp.chat/src/components/Messages/MessageContent.tsx
```diff
@@ -66,6 +71,15 @@ export const MessageContent: React.FC<MessageContentProps> = ({
     return <ReadReceiptContent />;
   }
 
+  if (contentType.sameAs(ContentTypeRemoteAttachment)) {
```
bennycode, line 74: Could we do something like `isRemoteAttachment(message)` here and then avoid the `content as RemoteAttachment` type casting?

apps/xmtp.chat/src/components/Messages/RemoteAttachmentContent.tsx
```diff
@@ -0,0 +1,219 @@
...
+        <AttachmentDetails
+          filename={content.filename}
+          fileSize={fileSize}
+          align={align}
+        />
+      </Paper>
+    );
+  }
+
+  const fileType = getFileType(content.filename);
```
bennycode, line 172: Instead of having our own file type recognition, could we do something like `new File(...).type` and use the browser API to determine the file type?

bennycode (changes requested):

apps/xmtp.chat/src/components/Conversation/ReplyPreview.tsx
rygine: this component is a replacement of the `stringify` method. can definitely bring back the method and use it here, but figured it would be nice to have everything in the same place. let me know what you think, i'm good either way.

apps/xmtp.chat/src/components/Messages/Message.tsx
rygine: sounds good

apps/xmtp.chat/src/components/Messages/MessageContent.tsx
rygine: for sure, just following the established pattern.

bennycode: Yeah, I have seen that `isRemoteAttachment(message)` is already being used in `ReplyPreview.tsx`, so it should work here as well.

apps/xmtp.chat/src/components/Conversation/ReplyPreview.tsx
bennycode: I think it would be nicer to keep this text content generation in a util, so that `ReplyPreview.tsx`, as a UI component, contains less business logic and focuses primarily on rendering the stringified content.

apps/xmtp.chat/src/components/Messages/MessageContent.tsx
rygine: it actually doesn't work well here because the `DecodedMessage` is not available and passing in the values manually doesn't infer the correct type for `content`.

apps/xmtp.chat/src/components/Messages/RemoteAttachmentContent.tsx
rygine: we could possibly do this, but it does add in complexity since it would rely on the availability of `decryptedUrl`. also, the `type` field only reads the extension that is passed into `new File(...)`. i think what we have now is fine.

bennycode (approved):

PR #1158, opened 2025-08-29
fix: `useNavigate` import

zkpepe:


zkpepe:
[macroscopeapp](https://github.com/apps/macroscopeapp), 
fix useNavigate import in `AppMenu.tsx`
`useNavigate` should be imported from `react-router-dom` instead of `react-router` so navigation works correctly.

PR #1166, opened 2025-09-02
Add appVersion to xmtp.chat

humanagent:
Add appVersion to xmtp.chat


apps/xmtp.chat/src/contexts/XMTPContext.tsx
```diff
@@ -117,6 +117,7 @@ export const XMTPProvider: React.FC<XMTPProviderProps> = ({
             env,
             loggingLevel,
             dbEncryptionKey,
+            appVersion: "xmtp-chat/0.0.0",
```
rygine: can we make this just `xmtp.chat`? is the version required?

humanagent: Done. Version Is required as i understand. Could just be 0.

rygine (approved):

PR #1174, opened 2025-09-09
Add events for standard content types

bennycode:


sdks/agent-sdk/README.md
```diff
@@ -39,7 +39,7 @@ const agent = await Agent.create(signer, {
 });
 
 // 3. Respond to any incoming message
```
rygine: should this comment change to `any incoming text message`?

sdks/agent-sdk/src/core/Agent.test.ts
```diff
@@ -1,18 +1,27 @@
+import assert from "node:assert";
```
rygine: think this might have been an auto-import

sdks/agent-sdk/src/core/Agent.ts
```diff
@@ -175,13 +197,60 @@ export class Agent<ContentTypes> extends EventEmitter<
...
+        },
+        onError: (error) => {
+          return this.#runErrorChain(error, null);
+        },
+        onFail: () => {
+          return this.#runErrorChain(
+            new Error("conversations.stream failed"),
+            null,
+          );
+        },
```
rygine, line 230: might not want to trigger the error chain on fail since it can recover on its own

sdks/agent-sdk/src/core/Agent.ts
```diff
@@ -192,7 +261,15 @@ export class Agent<ContentTypes> extends EventEmitter<
...
 
-  async #processMessage(message: DecodedMessage<ContentTypes>) {
+  async #processMessage(
+    message: DecodedMessage<ContentTypes>,
+    topic: keyof EventHandlerMap<ContentTypes> = "unhandledMessage",
+  ) {
+    // Skip messages with undefined content (failed to decode)
+    if (!hasDefinedContent(message)) {
+      return;
+    }
```
rygine, line 282: not something to worry about in this PR, but in future we could potentially allow devs to capture these types of messages in the stream

bennycode: Ok, I just had to safeguard it now for the MessageContext.

sdks/agent-sdk/src/core/Agent.ts
```diff
@@ -204,15 +281,18 @@ export class Agent<ContentTypes> extends EventEmitter<
...
-    context = new AgentContext(message, conversation, this.#client);
-    await this.#runMiddlewareChain(context);
+    context = new AgentContext({ message, conversation, client: this.#client });
+    await this.#runMiddlewareChain(context, topic);
   }
 
-  async #runMiddlewareChain(context: AgentContext<ContentTypes>) {
+  async #runMiddlewareChain(
+    context: AgentContext<ContentTypes>,
+    topic: keyof EventHandlerMap<ContentTypes> = "unhandledMessage",
```
rygine: instead of repeating this `keyof` type, WDYT of defining at the top?

sdks/agent-sdk/src/core/Agent.ts
```diff
@@ -175,13 +197,60 @@ export class Agent<ContentTypes> extends EventEmitter<
...
+        },
+        onFail: () => {
+          return this.#runErrorChain(
+            new Error("conversations.stream failed"),
+            null,
+          );
+        },
+      });
+
+      const messages =
```
rygine, line 233: WDYT of doing the same thing here as with the conversation stream and have a `#messagesStream` field?

sdks/agent-sdk/src/utils/message.ts
```diff
@@ -15,6 +15,12 @@ type ContentMessage = Pick<
   "content" | "contentType" | "fallback"
 >;
 
+export const hasDefinedContent = <ContentTypes = unknown>(
+  message: DecodedMessage<ContentTypes>,
+): message is DecodedMessage<ContentTypes> & {
+  content: NonNullable<ContentTypes>;
+} => !!message.content;
```
rygine: prefer `message.content === undefined` here since some current or future content types may include types that cast to `false`

bennycode: Ok. Macroscope also wants me to check against `null`.

sdks/agent-sdk/src/core/Agent.test.ts
bennycode: I am using it in `assert(contextSend)`.

sdks/agent-sdk/src/utils/message.ts
rygine: i will only be `undefined` if the content can't be decoded

sdks/agent-sdk/src/core/Agent.test.ts
rygine: why not use `expect`?

sdks/agent-sdk/src/utils/message.ts
bennycode: Yes but there is the `NonNullable` type which hints at `null` values.

rygine: probably best to drop `NonNullable` and just use `undefined`

sdks/agent-sdk/src/core/Agent.test.ts
bennycode: Since expect isn’t an assertion function, it doesn’t guarantee that I can safely use contextSend in the following call.

bennycode: But I can use `contextSend?.("Test response")`

sdks/agent-sdk/src/core/Agent.ts
bennycode: Yeah, let's do it in the next PR.

PR #1200, opened 2025-09-12
Stabilize conversation stream test

bennycode:
### Stabilize the jest "should stream group conversations" test in [Conversations.test.ts](https://github.com/xmtp/xmtp-js/pull/1200/files#diff-66116d04f0b220afaee5094ae02adcc0c5fb2635d790dacc13e8a4f8181ba8d0)
Replace order-dependent assertions with set-based assertions in the jest "should stream group conversations" test. Introduce `expectedIds` and `receivedIds` sets to collect streamed conversation IDs, terminate the stream once both expected IDs are received, and compare set contents and size after the loop. Remove the `count` variable and related conditional checks in [Conversations.test.ts](https://github.com/xmtp/xmtp-js/pull/1200/files#diff-66116d04f0b220afaee5094ae02adcc0c5fb2635d790dacc13e8a4f8181ba8d0).

#### 📍Where to Start
Start with the jest test case "should stream group conversations" in [Conversations.test.ts](https://github.com/xmtp/xmtp-js/pull/1200/files#diff-66116d04f0b220afaee5094ae02adcc0c5fb2635d790dacc13e8a4f8181ba8d0).

#### Changes since #1200 opened

- Modified conversation stream test data handling and termination logic [9b611ef]
----

_[Macroscope](https://app.macroscope.com) summarized 9b611ef._

### DIFF
--- a/sdks/node-sdk/test/Conversations.test.ts
+++ b/sdks/node-sdk/test/Conversations.test.ts
@@ -417,8 +417,8 @@ describe("Conversations", () => {
       client3.inboxId,
     ]);
 
-    const expectedIds = new Set([conversation1.id, conversation2.id]);
-    const receivedIds = new Set<string>();
+    const expectedIds = [conversation1.id, conversation2.id];
+    const receivedIds: string[] = [];
 
     setTimeout(() => {
       void stream.end();
@@ -429,14 +429,10 @@ describe("Conversations", () => {
         break;
       }
       expect(convo).toBeDefined();
-      receivedIds.add(convo.id);
-      if (receivedIds.size === 2) {
-        void stream.end();
-        break;
-      }
+      receivedIds.push(convo.id);
     }
 
-    expect(receivedIds.size).toBe(2);
+    expect(receivedIds.length).toBe(2);
     expect(receivedIds).toEqual(expectedIds);
     expect(
       (await client3.conversations.getConversationById(conversation1.id))?.id,


sdks/node-sdk/test/Conversations.test.ts
```diff
@@ -417,25 +417,27 @@ describe("Conversations", () => {
...
       expect(convo).toBeDefined();
-      if (count === 1) {
-        expect(convo.id).toBe(conversation1.id);
-      }
-      if (count === 2) {
-        expect(convo.id).toBe(conversation2.id);
+      receivedIds.add(convo.id);
+      if (receivedIds.size === 2) {
+        void stream.end();
+        break;
```
rygine: this isn't necessary. the stream will automatically end in 2 seconds.

sdks/node-sdk/test/Conversations.test.ts
```diff
@@ -417,25 +417,27 @@ describe("Conversations", () => {
       client3.inboxId,
     ]);
 
+    const expectedIds = new Set([conversation1.id, conversation2.id]);
+    const receivedIds = new Set<string>();
```
rygine: any reason to use a set over an array?

sdks/node-sdk/test/Conversations.test.ts
bennycode: My idea was that we end the stream ourselves in cases where the test runs after than 2 seconds. I can add a `Prom.race` so we can cancel the timeout if that's the case.

sdks/node-sdk/test/Conversations.test.ts
bennycode: To avoid duplicates but maybe good to also detect duplicates (in cases where the stream may send some!). 💡

sdks/node-sdk/test/Conversations.test.ts
rygine: not sure we need to get that complex here. the timeout should handle everything for us. it also ensures that no unexpected items are in the stream.

sdks/node-sdk/test/Conversations.test.ts
rygine: yep, in my other tests i use arrays for this purpose, to ensure no duplicates.

rygine (approved):

PR #1203, opened 2025-09-14
Use filter with config object

bennycode:
Based on #1201


sdks/agent-sdk/src/core/AgentContext.ts
```diff
@@ -0,0 +1,12 @@
...
+export interface AgentBaseContext<ContentTypes = unknown> {
+  client: Client<ContentTypes>;
+  conversation: Conversation;
+  message: DecodedMessage;
+}
+
+export interface AgentErrorContext<ContentTypes = unknown>
+  extends Partial<AgentBaseContext<ContentTypes>> {
+  client: Client<ContentTypes>;
+}
```
rygine: why `interface` instead of `type`?

PR #1204, opened 2025-09-14
Expect conversation events to be out of order

bennycode:


rygine:
are these tests being flaky for you?

bennycode:
@rygine "should only stream group conversations" is definitely flaky: https://github.com/xmtp/xmtp-js/actions/runs/17702011299/job/50308896648 - That's why I am fixing it with this PR.

The main issue is that order in a stream is not guaranteed. I am also addressing "should only stream DM conversations" and "should only stream group conversation messages", since both assume that messages will be streamed in sequence.

bennycode:
@rygine I’m also sorting the ID expectations, otherwise the test could still fail if values are pushed into the array in the wrong order (as seen in https://github.com/xmtp/xmtp-js/actions/runs/17736896800/job/50400872065).

PR #1223, opened 2025-09-18
Use callbacks when streaming messages

bennycode:


sdks/agent-sdk/src/core/Agent.ts
```diff
@@ -191,58 +190,68 @@ export class Agent<ContentTypes = unknown> extends EventEmitter<
...
-              await this.#processMessage(message, "reply");
-              break;
-            case filter.isText(message):
-              await this.#processMessage(message, "text");
-              break;
-            default:
-              await this.#processMessage(message);
-              break;
+      this.#messageStream = await this.#client.conversations.streamAllMessages({
+        ...options,
```
rygine: since we're overwriting these callbacks, we should also run the ones passed in as options.

sdks/agent-sdk/src/core/Agent.ts
```diff
@@ -191,58 +190,68 @@ export class Agent<ContentTypes = unknown> extends EventEmitter<
...
-            await this.stop();
-            break;
+        },
+        onError: async (error) => {
+          await this.#handleStreamError(error, options);
+        },
+        onFail: () => {
+          if (this.#messageStream) {
+            this.#messageStream = undefined;
+            queueMicrotask(() => this.start(options));
```
rygine: this is not necessary as it's automatic in the node SDK

bennycode: Ok, shall I simply remove `onFail`?

humanagent (approved):

rygine (approved):

PR #1233, opened 2025-09-22
Add name resolver to API service

rygine:


apps/xmtp.chat-api-service/src/api/v1/resolve.router.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+    console.error(error);
+    res.status(500).json({ error: "Failed to resolve name" });
+  }
+}
+
+const resolveRouter = Router();
+resolveRouter.get("/name/:name", resolveName);
+
+// resolveRouter.post("/addresses", resolveAddresses);
```
rygine, line 174: this inactive route will be for resolving addresses to identities

apps/xmtp.chat-api-service/src/api/v1/resolve.router.ts
```diff
@@ -0,0 +1,171 @@
...
+    const fetchedProfiles = await batchFetchProfiles(addressesToResolve);
+
+    // insert new profiles into the database
+    const newProfiles = fetchedProfiles.filter(
+      (p) => p.address && missingAddresses.includes(p.address),
+    );
+    const createdProfiles = await prisma.profile.createManyAndReturn({
+      data: newProfiles.map((p) => ({
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        address: p.address!,
```
bennycode: How about using Node's `assert` here to make sure `address` exists or marking address as optional in the Prisma schema?

rygine: i added an assertion to fix this. the `address` must exist in the schema.

PR #1313, opened 2025-10-06
Add inbox store, refactor hooks

rygine:


apps/xmtp.chat/src/components/Conversations/ConversationsNavbar.tsx
```diff
@@ -6,44 +6,55 @@ import { useConversations } from "@/hooks/useConversations";
...
+    syncAll,
+  } = useConversations();
+  const stopConversationStreamRef = useRef<(() => void) | null>(null);
+  const stopAllMessagesStreamRef = useRef<(() => void) | null>(null);
 
-  const startStream = useCallback(async () => {
-    stopStreamRef.current = await stream();
-  }, [stream]);
+  const startStreams = useCallback(async () => {
+    stopConversationStreamRef.current = await stream();
```
bennycode, line 22: What happens if only one of the streams fails? Will the app then be left in an incosistent state?

apps/xmtp.chat/src/components/Messages/GroupUpdatedContent.tsx
```diff
@@ -17,18 +18,20 @@ const GroupMembersUpdatedContent: React.FC<GroupMembersAddedContentProps> = ({
...
 }) => {
-  const { members } = useConversationContext();
+  const { conversationId } = useConversationContext();
+  const { members } = useConversation(conversationId);
+  const initiatedByMember = members.get(initiatedBy);
   return (
     <Group gap="4" wrap="wrap" justify="center">
       <Identity
-        address={members.get(initiatedBy) ?? ""}
+        address={initiatedByMember?.accountIdentifiers[0].identifier ?? ""}
```
bennycode: We frequently use `.accountIdentifiers[0].identifier ?? ""` in this file. What do you think about introducing a utility function like `getIdentifier(initiatedByMember)` and using it across all identity components?

apps/xmtp.chat/src/components/Conversations/ConversationsNavbar.tsx
rygine: yes

bennycode: How about wrapping both statements in a try-catch to fail if one of them fails?

rygine: errors thrown in the `stream()` and `streamAllMessages()` function will create an error popup with the error message

apps/xmtp.chat/src/components/Messages/GroupUpdatedContent.tsx
rygine: added here: https://github.com/xmtp/xmtp-js/pull/1313/commits/e1f65bc8f4076c5c910fdf52f3d0910985267069

bennycode (approved):

apps/xmtp.chat/src/components/Conversation/ManageMetadataModal.tsx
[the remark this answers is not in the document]
rygine: this is intended

PR #1331, opened 2025-10-08
Fix `/dm` route

rygine:


apps/xmtp.chat/src/components/Conversation/LoadDM.tsx
```diff
@@ -41,8 +65,10 @@ export const LoadDM: React.FC = () => {
...
 
         if (!inboxId) {
-          setMessage("Invalid address, redirecting...");
-          setTimeout(() => {
+          setMessage(
+            "Address not registered on the XMTP network, redirecting...",
+          );
+          timeout = setTimeout(() => {
             void navigate("/");
           }, 2000);
```
bennycode: Shall we put the 2000 as a constant for navigation timeout as it is being reused in multiple places?

PR #1345, opened 2025-10-14
Update inbox store, adjust various styles

rygine:
### Update inbox store to track group `admins`, `superAdmins`, and `permissions`, and refactor `ContentLayout` to add optional borders and an aside while adjusting UI controls to `ActionIcon` and resizing the AppHeader status badge to `lg`
This change updates the inbox store to persist group conversation `admins`, `superAdmins`, and `permissions`, adds a `syncPermissions` action, and exposes new selector hooks. It refactors `ContentLayout` into header/content/footer components with optional borders and an aside panel, and standardizes several menu/copy controls to use `ActionIcon`. It also adjusts the AppHeader status badge size and removes the composer top border.

- Extend inbox store state and actions to include group conversation `admins`, `superAdmins`, and `permissions`, including refresh on group update messages and a `syncPermissions` action in [store.ts](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-572e773e875d19026abd898404b7ae68fa8a96aaf08601ef5294728d3cb9fce3) with selectors in [hooks.ts](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-a8f2befe07d48ae9d6b73859fe0f0a452ec5f2f2b8e6e24a69134958be963b89)
- Refactor `ContentLayout` into `ContentLayoutHeader`, `ContentLayoutContent`, and `ContentLayoutFooter`, add `withBorders` and `aside` support, remove focus trapping, and update related styles in [ContentLayout.tsx](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-ea4030fcbc20fa449f19e8b124df92bc61273ebbf8f909c6e1bc65929ba1e352) and [ContentLayout.module.css](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-6eedf77fda786029e7228626e6af81e372f5305944cd1c7f1035decdd1208801)
- Update UI controls to `ActionIcon` for menus and copy buttons across affected components, and resize the AppHeader badge in [AppHeader.tsx](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-d82e07c7e0f6a1ffe09de08d3ad1316fcb41390487050a3d580b8ff3e9aab208)
- Adjust `useConversation` to include `admins`, `permissions`, and `superAdmins` in [useConversation.ts](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-7d6bf0c4e42ce58d1bb76cdffeb19d380903c904ca4be18d58d3054a4dd4b9ab)
- Update main layout and conversation components styling, including removing the composer top border and disabling borders in the conversations navbar via `withBorders={false}`

#### 📍Where to Start
Start with the inbox store changes in [store.ts](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-572e773e875d19026abd898404b7ae68fa8a96aaf08601ef5294728d3cb9fce3), then review the new selector hooks in [hooks.ts](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-a8f2befe07d48ae9d6b73859fe0f0a452ec5f2f2b8e6e24a69134958be963b89), followed by the `ContentLayout` refactor in [ContentLayout.tsx](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-ea4030fcbc20fa449f19e8b124df92bc61273ebbf8f909c6e1bc65929ba1e352) and its styles in [ContentLayout.module.css](https://github.com/xmtp/xmtp-js/pull/1345/files#diff-6eedf77fda786029e7228626e6af81e372f5305944cd1c7f1035decdd1208801).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized e846044. 14 files reviewed, 16 issues evaluated, 14 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>apps/xmtp.chat/src/components/App/AppHeader.tsx — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 44](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/components/App/AppHeader.tsx#L44): Potential crash when computing `accountIdentifier`: `client.accountIdentifier?.identifier.toLowerCase()` guards only the presence of `client.accountIdentifier`, not the `identifier` property itself. If `client.accountIdentifier` exists but `identifier` is `undefined`, `null`, or any non-string, calling `.toLowerCase()` will throw at runtime. To make this safe, also guard `identifier` before lowercasing, e.g., `const id = client.accountIdentifier?.identifier; setAccountIdentifier(typeof id === 'string' ? id.toLowerCase() : null);`. <b>[ Low confidence ]</b>
- [line 46](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/components/App/AppHeader.tsx#L46): Stale `accountIdentifier` can be displayed due to an imprecise `useEffect` dependency: `[client.accountIdentifier]`. If `client.accountIdentifier` is a stable object whose `identifier` field changes, the effect won't run and the UI will show outdated data. Depend on the specific value used, e.g., `[client.accountIdentifier?.identifier]`, or compute the value on each render without an effect if cheap. <b>[ Code style ]</b>
</details>

<details>
<summary>apps/xmtp.chat/src/components/BadgeWithCopy.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 37](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/components/BadgeWithCopy.tsx#L37): Pressing keyboard activation keys on the `ActionIcon` (`Enter` or `Space`) can cause `handleCopy` to be invoked twice: once via the explicit `onKeyDown` handler in `handleKeyboardCopy` and once via the native button activation that fires `onClick`. This violates at-most-once action semantics and can lead to double application of side effects (e.g., two clipboard writes, two state updates for `copied`). To fix, remove the custom keyboard handler (Mantine/HTML button handles keyboard activation natively) or call `event.preventDefault()` and suppress the native `click` for those keys, handling copy in only one path. <b>[ Low confidence ]</b>
</details>

<details>
<summary>apps/xmtp.chat/src/components/Conversations/ConversationsNavbar.tsx — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 47](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/components/Conversations/ConversationsNavbar.tsx#L47): Unhandled promise rejection risk in `useEffect` startup sequence. In `ConversationsNavbar`, `loadConversations` is invoked via `void loadConversations()` inside a `useEffect` with an empty dependency array. If either `await sync(true)` or `await startStreams()` throws (e.g., network/client errors), the rejection is not caught. React does not handle unhandled rejections from async effects when the returned promise is ignored, leading to global unhandled promise rejection and potential app instability. Add a try/catch around the awaited calls and surface an error state or at least swallow/log the error. <b>[ Previously rejected ]</b>
- [line 76](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/components/Conversations/ConversationsNavbar.tsx#L76): Re-entrancy/race condition can create duplicate or mismatched stream lifecycles when `handleSync`/`handleSyncAll` are invoked concurrently. The menu `disabled` state is bound to `syncing`, but `handleSync` calls `sync()` without `fromNetwork=true`, so `syncing` stays false while `loading` may be true. This allows multiple rapid clicks, triggering overlapping executions of `handleSync`. Because `stopStreams()` is called at the start of each handler and `startStreams()` is called at the end, interleavings can lead to: (a) streams being stopped just before another call starts them; (b) multiple streams being started if a second call issues `stopStreams()` before the first has created the stream refs; resulting in duplicate message handling and/or leaked streams. Guard with a mutual-exclusion flag (e.g., disable on `loading` as well or add an internal `isSyncing` guard), or serialize calls to ensure at-most-once execution while prior sync is in-flight. <b>[ Previously rejected ]</b>
- [line 78](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/components/Conversations/ConversationsNavbar.tsx#L78): Unhandled promise rejection risk in menu callbacks. `handleSync` and `handleSyncAll` are `async` and are invoked via `onSync={() => void handleSync()}` and `onSyncAll={() => void handleSyncAll()}`. If either `sync()`/`syncAll()` or `startStreams()` throws, the returned promise is ignored and the rejection is not caught, causing a global unhandled promise rejection. Wrap the bodies of `handleSync`/`handleSyncAll` in try/catch or have the menu trigger a handler that manages errors. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>apps/xmtp.chat/src/components/CopyButton.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 36](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/components/CopyButton.tsx#L36): Same double-invocation issue in `CopyButton`: pressing `Enter` or `Space` on the `ActionIcon` triggers `handleCopy` both in `onKeyDown` (`handleKeyboardCopy`) and the native `click` event, causing two copy operations and duplicated side effects. Resolve by removing the custom keyboard handler or preventing default and suppressing the native click so copy happens at most once. <b>[ Low confidence ]</b>
</details>

<details>
<summary>apps/xmtp.chat/src/hooks/useConversation.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 31](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/hooks/useConversation.ts#L31): useConversation throws an Error when the conversation is not found (`if (!conversation) { throw new Error(...) }`). This causes a hard runtime crash of any component using the hook with an ID that hasn't been loaded yet or is invalid. Given there is no guard guaranteeing the conversation exists before use, this should return a defined outcome without crashing (e.g., return a status flag and allow the UI to render a loading or not-found state). <b>[ Low confidence ]</b>
</details>

<details>
<summary>apps/xmtp.chat/src/layouts/ContentLayout.tsx — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 90](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/layouts/ContentLayout.tsx#L90): Truthiness checks for rendering and styling use `footer && ...` and `!footer && classes.noFooter` (lines `90`, `175-177`). Since `footer` is typed as `React.ReactNode`, valid nodes like `0` or `""` would be treated as falsy, suppressing the footer render and possibly misapplying `noFooter` styling, leading to inconsistent behavior if such values are provided. <b>[ Code style ]</b>
- [line 97](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/layouts/ContentLayout.tsx#L97): `ScrollArea` receives `type="scroll"` (line `97`), which does not match Mantine's documented `type` prop values (typically `auto`, `always`, or `hover`). Passing an invalid value may cause the `ScrollArea` to default unexpectedly or not behave as intended, leading to missing or incorrect scrollbar behavior at runtime. <b>[ Out of scope ]</b>
- [line 156](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/layouts/ContentLayout.tsx#L156): Removal of `FocusTrap` changes externally visible behavior: focus is no longer trapped within the layout during loading or when overlays are present. Previously, the entire layout was wrapped in `FocusTrap`, preventing users from tabbing to off-layout elements while content is loading or when modal-like overlays cover the UI. Now, `LoadingOverlay` is shown (`visible={loading}`) but does not trap focus, which can allow keyboard/focus interactions with underlying or external elements. This is a contract parity regression that can lead to inconsistent/incorrect interaction during loading and accessibility issues. <b>[ Low confidence ]</b>
</details>

<details>
<summary>apps/xmtp.chat/src/stores/inbox/hooks.ts — 0 comments posted, 4 evaluated, 3 filtered</summary>

- [line 14](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/stores/inbox/hooks.ts#L14): EMPTY_ADMINS is defined as a single shared mutable array (`const EMPTY_ADMINS: string[] = [];`) and used as the fallback in `useAdmins` and `useSuperAdmins`. Returning the same array instance across calls allows accidental mutation by consumers to affect all components relying on the fallback, leading to state leakage and hard-to-debug behavior. Use an immutable/frozen array or return a fresh empty array for each invocation. <b>[ Low confidence ]</b>
- [line 79](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/stores/inbox/hooks.ts#L79): Type/shape mismatch fallback risk: `useAdmins` and `useSuperAdmins` return `EMPTY_ADMINS` (typed as `string[]`). However, the store's `admins`/`superAdmins` entries are populated via `conversation.listAdmins()`/`conversation.listSuperAdmins()`. If those methods return objects (e.g., member records) rather than strings, components may process admin entries expecting object properties and break when the fallback is a `string[]`. To avoid runtime errors, align the fallback to the actual element shape (e.g., an empty array of the correct type) or document/enforce the element type consistently. <b>[ Low confidence ]</b>
- [line 91](https://github.com/xmtp/xmtp-js/blob/e8460445bd627d8bdab938c7b67dd5f3e7a32c1b/apps/xmtp.chat/src/stores/inbox/hooks.ts#L91): usePermissions returns `undefined` when no permissions are present for a conversation (`state.permissions.get(conversationId)` may be missing). In contrast to other hooks that provide safe defaults (e.g., `EMPTY_METADATA`, `EMPTY_MEMBERS`, `EMPTY_MESSAGES`, `EMPTY_ADMINS`), this hook exposes `undefined` to consumers. Any component that dereferences fields on `permissions` without a guard will crash at runtime. To preserve invariants and avoid undefined dereferences, return a safe default (e.g., an empty permissions object or array) when the entry is missing. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/apps/xmtp.chat/src/components/App/AppHeader.tsx
+++ b/apps/xmtp.chat/src/components/App/AppHeader.tsx
@@ -66,7 +66,7 @@ export const AppHeader: React.FC<AppHeaderProps> = ({
         </Flex>
       </Flex>
       <Group align="center" gap="xs">
-        <Badge size="xl" radius="md" variant="default" p={0}>
+        <Badge size="lg" radius="md" variant="default" p={0}>
           <Group align="center" gap="xs" px="sm">
             <GlowingCircle />
             <Text size="xs" fw={700}>


apps/xmtp.chat/src/stores/inbox/store.ts
```diff
@@ -37,10 +38,16 @@ export type InboxState = {
   lastSentAt: Map<string, bigint | undefined>;
   // the members of each conversation
   members: Map<string, Map<string, SafeGroupMember>>;
+  // the admins of each conversation
+  admins: Map<string, string[]>;
```
bennycode, line 42: Is this a collection that maps a conversation ID to admin addresses? Can we add a type alias (like `type ConversatonId = string`) to make it clearer?

rygine: yes, this is conversation ID => inbox IDs.

i'll do that in a followup PR, good callout!

PR #1432, opened 2025-10-22
Allow db encryption key with 0x prefix

bennycode:
### Accept a 32-byte 0x-prefixed hex string for `sdks/node-sdk` client `utils.createClient` `options.dbEncryptionKey`
Add hex string support for the database encryption key and convert it to bytes during client creation.

- Convert `options.dbEncryptionKey` when it is a 0x-prefixed even-length hex string in `utils.createClient` and pass a byte buffer to `createNodeClient` in [createClient.ts](https://github.com/xmtp/xmtp-js/pull/1432/files#diff-384c4dfe50aecc407db2a90b34a4b35e41ddbafca70ca720f757f920439f76a3)
- Introduce `HexString` type and `isHexString` validation utility in [validation.ts](https://github.com/xmtp/xmtp-js/pull/1432/files#diff-9c61790ece8b2cba2252ca3348d149668866f50243f62b9f2e9be40c0d068ae8) and re-export from the package root in [index.ts](https://github.com/xmtp/xmtp-js/pull/1432/files#diff-a799851ddab26b071b3e6753eb77936fa9fc3da7839b22d80d9f257177cdd070)
- Update `StorageOptions.dbEncryptionKey` to `Uint8Array | HexString` and add docs in [types.ts](https://github.com/xmtp/xmtp-js/pull/1432/files#diff-5ab8d64966cdd3ef2296dd75417e657434be92abe4cd0072322fa40832ab5485)
- Add tests for client creation and hex validation in [Client.test.ts](https://github.com/xmtp/xmtp-js/pull/1432/files#diff-c00658c71eee5bdefef8b17b8d65ffb08e066b43f77200db2e049e2e6f137442) and [validation.test.ts](https://github.com/xmtp/xmtp-js/pull/1432/files#diff-ff7c582c96d218e671576e2691d1256af3e95f3af65e7bb441190186c80cbd9a)

#### 📍Where to Start
Start with the `createClient` factory in [createClient.ts](https://github.com/xmtp/xmtp-js/pull/1432/files#diff-384c4dfe50aecc407db2a90b34a4b35e41ddbafca70ca720f757f920439f76a3), focusing on the `dbEncryptionKey` handling and the use of `isHexString`.

#### Changes since #1432 opened

- Removed comment line referencing external issue URL from `vitest` configuration [433aec7]
- Added release metadata for database encryption key format support [afeaddb]
----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized afeaddb. 4 files reviewed, 5 issues evaluated, 4 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>sdks/node-sdk/src/utils/createClient.ts — 0 comments posted, 3 evaluated, 2 filtered</summary>

- [line 31](https://github.com/xmtp/xmtp-js/blob/afeaddb58f6fb9d98768d1645d08efb5ec9fe00f/sdks/node-sdk/src/utils/createClient.ts#L31): When `options.dbPath` is a function, its return value is assumed to be a string and passed to `createNodeClient` without validation. A JS consumer could return `undefined` or a non-string, leading to passing `undefined` for `dbPath` at runtime, which may cause errors in the native binding. Add a runtime check to ensure the callback returns a string, and fail fast with a clear error if not. <b>[ Low confidence ]</b>
- [line 50](https://github.com/xmtp/xmtp-js/blob/afeaddb58f6fb9d98768d1645d08efb5ec9fe00f/sdks/node-sdk/src/utils/createClient.ts#L50): `dbEncryptionKey` handling allows invalid string values to pass through to `createNodeClient`. If `options.dbEncryptionKey` is a non-hex string (e.g., `'foo'` or `'0X...'`), `isHexString` returns false and the raw string is passed to `createNodeClient`, which likely expects a `Uint8Array`. This can cause runtime errors in the native binding. The code should either reject non-hex strings with a clear error or normalize supported inputs and otherwise set the key to `undefined`/reject. <b>[ Already posted ]</b>
</details>

<details>
<summary>sdks/node-sdk/src/utils/validation.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 3](https://github.com/xmtp/xmtp-js/blob/afeaddb58f6fb9d98768d1645d08efb5ec9fe00f/sdks/node-sdk/src/utils/validation.ts#L3): `isHexString` does not enforce the required length of the encryption key. The docs and type comment specify “Encryption key for the local DB (32 bytes, hex)”, but the regex `^0x(?:[0-9a-fA-F]{2})+$` accepts any even number of hex digits. Keys shorter or longer than 32 bytes will be accepted and converted, potentially causing failures or silent misuse in the database encryption layer. Enforce exactly 32 bytes (64 hex characters after `0x`) and reject others. <b>[ Already posted ]</b>
- [line 3](https://github.com/xmtp/xmtp-js/blob/afeaddb58f6fb9d98768d1645d08efb5ec9fe00f/sdks/node-sdk/src/utils/validation.ts#L3): `isHexString` is case-sensitive on the `0x` prefix and will reject `0X...` uppercase-prefixed hex strings. When rejected, the code in `createClient` passes the raw string to `createNodeClient`, potentially causing a runtime error. Either accept both `0x` and `0X` or explicitly reject with a clear error instead of passing an invalid type downstream. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/.changeset/long-donkeys-exist.md
+++ b/.changeset/long-donkeys-exist.md
@@ -1,0 +1,6 @@
+---
+"@xmtp/agent-sdk": patch
+"@xmtp/node-sdk": patch
+---
+
+Enable hex strings as database encryption keys


sdks/node-sdk/src/utils/createClient.ts
```diff
@@ -46,13 +46,18 @@ export const createClient = async (
     ? SyncWorkerMode.disabled
     : SyncWorkerMode.enabled;
 
+  const dbEncryptionKey =
+    typeof options?.dbEncryptionKey === "string"
+      ? Buffer.from(options.dbEncryptionKey.replace(/^0x/, ""), "hex")
+      : options?.dbEncryptionKey;
```
rygine: should we validate the string here?

sdks/node-sdk/src/utils/validation.ts
```diff
@@ -0,0 +1,14 @@
...
+export function isHexString(value: unknown): asserts value is HexString {
+  if (typeof value !== "string") {
+    throw new TypeError("Value must be a string");
+  }
+
+  if (value.length % 2 !== 0) {
+    throw new TypeError("Invalid hex string length");
+  }
+
+  if (!/^0x[0-9a-fA-F]*$/.test(value)) {
```
rygine: i think the `*` should be a `+` or `{2,}` (if an even number is required and the `0x` part will be removed)

sdks/node-sdk/src/utils/createClient.ts
```diff
@@ -46,13 +47,24 @@ export const createClient = async (
     ? SyncWorkerMode.disabled
     : SyncWorkerMode.enabled;
 
+  let dbEncryptionKey: Uint8Array | undefined = undefined;
+  if (typeof options?.dbEncryptionKey === "string") {
+    isHexString(options.dbEncryptionKey);
```
rygine: is this supposed to be part of the `if` statement?

bennycode: Yes, unfortunately I cannot combine the assertion function directly with `typeof options?.dbEncryptionKey === "string"`.

sdks/node-sdk/src/utils/validation.ts
bennycode: Good catch, with `{2,}` we can require at least 2 digits to convert to bytes.

sdks/node-sdk/src/utils/validation.ts
```diff
@@ -0,0 +1,14 @@
+export type HexString = `0x${string}`;
+export function isHexString(value: unknown): asserts value is HexString {
```
rygine: why doesn't this return a `boolean`?

sdks/agent-sdk/vitest.config.ts
```diff
@@ -3,7 +3,7 @@ import { defineConfig, type Plugin } from "vitest/config";
 
 export default defineConfig({
   // @see https://github.com/aleclarson/vite-tsconfig-paths/issues/176
-  plugins: [tsconfigPaths() as Plugin],
+  plugins: [tsconfigPaths()],
```
rygine, line 5: if this assertion is no longer needed, mind removing the comment?

rygine (approved):

sdks/agent-sdk/vitest.config.ts
bennycode: Sure! ESLint removed it, I didn't even notice the comment. Thanks for mentioning it.

PR #1530, opened 2025-11-21
Make streaming options configurable

bennycode:
### Make `sdks/agent-sdk/src/core/Agent.ts::Agent.start` accept `StreamOptions` to configure conversation and message streaming
Add an optional `StreamOptions<Message, DecodedMessage<ContentTypes>>` parameter to `Agent.start` and forward it to `this.#client.conversations.stream(...)` and `this.#client.conversations.streamAllMessages(...)`, overriding any `onValue` from options with the method’s handler.

#### 📍Where to Start
Start with the `Agent.start` method in [Agent.ts](https://github.com/xmtp/xmtp-js/pull/1530/files#diff-86fb5a7aeff0525692d6db76ada7057958f615d7c746035cea44fe3745ac4cb4).

#### Changes since #1530 opened

- Constrained streaming options for Agent configuration by creating a restricted type alias and updating method signatures [ad68431]
- Added changeset file for streaming options configuration [0ed21a4]
- Simplified streaming options type system by removing generic type parameters [0018e18]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 0018e18. 1 file reviewed, 2 issues evaluated, 2 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>sdks/agent-sdk/src/core/Agent.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 298](https://github.com/xmtp/xmtp-js/blob/0018e182c29fa0d0449fbfafe3017cd12678c6d7/sdks/agent-sdk/src/core/Agent.ts#L298): start() accepts `options` and forwards them to both `conversations.stream` and `conversations.streamAllMessages`, but if an error occurs during startup or at runtime and recovery triggers a restart, the restart is invoked without the original options. In the catch block, `await this.#handleStreamError(error)` is called, and inside `#handleStreamError` the code does `queueMicrotask(() => this.start());` without passing or persisting the original `options`. This causes the agent to silently lose caller-provided streaming configuration on auto-restart. <b>[ Low confidence ]</b>
- [line 425](https://github.com/xmtp/xmtp-js/blob/0018e182c29fa0d0449fbfafe3017cd12678c6d7/sdks/agent-sdk/src/core/Agent.ts#L425): If an error occurs in start() before both streams are fully established, the catch block calls `await this.#handleStreamError(error)`. When recovery fails (`recovered` is false), `#handleStreamError` does not reset `#isLocked`. Control returns to start(), which also does not reset `#isLocked`. As a result, `#isLocked` remains true, permanently preventing subsequent calls to `start()` from doing anything. This leaves the Agent in a locked state after an unrecoverable startup error. <b>[ Out of scope ]</b>
</details>

</details>

### DIFF
--- a/sdks/agent-sdk/src/core/Agent.ts
+++ b/sdks/agent-sdk/src/core/Agent.ts
@@ -25,7 +25,6 @@ import {
   type CreateGroupOptions,
   type DecodedMessage,
   type HexString,
-  type Message,
   type StreamOptions,
   type XmtpEnv,
 } from "@xmtp/node-sdk";
@@ -108,10 +107,7 @@ export type AgentErrorMiddleware<ContentTypes = unknown> = (
   next: (err?: unknown) => Promise<void> | void,
 ) => Promise<void> | void;
 
-export type AgentStreamingOptions<ContentTypes> = Omit<
-  StreamOptions<Message, DecodedMessage<ContentTypes>>,
-  "onValue" | "onError"
->;
+export type AgentStreamingOptions = Omit<StreamOptions, "onValue" | "onError">;
 
 export type StreamAllMessagesOptions<ContentTypes> = Parameters<
   Client<ContentTypes>["conversations"]["streamAllMessages"]
@@ -299,7 +295,7 @@ export class Agent<ContentTypes = unknown> extends EventEmitter<
     }
   }
 
-  async start(options?: AgentStreamingOptions<ContentTypes>) {
+  async start(options?: AgentStreamingOptions) {
     if (this.#isLocked || this.#conversationsStream || this.#messageStream)
       return;
 


sdks/agent-sdk/src/core/Agent.ts
```diff
@@ -292,14 +294,15 @@ export class Agent<ContentTypes = unknown> extends EventEmitter<
     }
   }
 
-  async start() {
+  async start(options?: StreamOptions<Message, DecodedMessage<ContentTypes>>) {
```
rygine: i think we can remove the generic values and the `onValue` field for this since we're providing it.

so something like this:
```
export type AgentStreamOptions = Omit<StreamOptions, "onValue">;
```

WDYT?

rygine (approved):

sdks/agent-sdk/src/core/Agent.ts
```diff
@@ -106,6 +108,11 @@ export type AgentErrorMiddleware<ContentTypes = unknown> = (
   next: (err?: unknown) => Promise<void> | void,
 ) => Promise<void> | void;
 
+export type AgentStreamingOptions<ContentTypes> = Omit<
+  StreamOptions<Message, DecodedMessage<ContentTypes>>,
```
rygine: the generic values of `StreamOptions` are not necessary here as they are only used in the `onValue` callback.

PR #1547, opened 2025-11-23
feat: add inline actions to agent-sdk

humanagent:
### Add inline actions to agent-sdk and emit `actions` and `intent` events in `Agent` with codecs `ActionsCodec` and `IntentCodec`
Introduce a public `./inline-actions` export, add codecs and content types for actions and intent, wire message filters and `Agent` event dispatch, and provide an `inlineActionsMiddleware` plus builder/utilities for sending and handling action menus.

#### 📍Where to Start
Start with codec definitions in [`types/ActionsContent.ts`](https://github.com/xmtp/xmtp-js/pull/1547/files#diff-34b23c8c776a904a35a22b951e3c7992454ed931423969f88f4c99fc7613effb) and [`types/IntentContent.ts`](https://github.com/xmtp/xmtp-js/pull/1547/files#diff-c9946f3bfb05cbee3dbd7917a7ccf0680a92c2fef94db3673d7a00b7799941f4), then follow integration in [`src/core/Agent.ts`](https://github.com/xmtp/xmtp-js/pull/1547/files#diff-86fb5a7aeff0525692d6db76ada7057958f615d7c746035cea44fe3745ac4cb4) and predicates in [`src/core/filter.ts`](https://github.com/xmtp/xmtp-js/pull/1547/files#diff-cd25531655110eba5f4fcf8b590aaf542f61efec22ade311a7ee83972225b376).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized ff73769. 6 files reviewed, 26 issues evaluated, 25 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts — 0 comments posted, 15 evaluated, 14 filtered</summary>

- [line 18](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L18): `lastSentActionMessage` is a module-global mutable variable used to track the "last" message across executions. In any environment with concurrent or interleaved requests, this can leak state across sessions/users and produce race conditions: the value may be overwritten by a different request before a consumer reads it, or a consumer may read a stale value from an earlier request. This violates at-most-once and provenance invariants for "last" and can cause incorrect behavior in multi-context usage. <b>[ Already posted ]</b>
- [line 18](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L18): `lastSentActionMessage` can retain a reference to arbitrary `unknown` objects indefinitely at module scope, creating memory retention across requests. Without explicit per-context cleanup on all exit paths (success, error, cancellation), large messages or cyclic graphs may be kept alive longer than intended, causing memory growth/leaks until overwritten or the process restarts. <b>[ Low confidence ]</b>
- [line 21](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L21): `lastShownMenu` is a module-global mutable variable tracking a `{ config: AppConfig; menuId: string }`. In concurrent or multi-session environments, this can leak a menu/config from one session into another and cause races where one request overwrites the "last" menu while another expects to re-show its own menu, resulting in wrong navigation for users and broken invariants. <b>[ Already posted ]</b>
- [line 23](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L23): Global `actionHandlers` registry allows silent handler overwrites based on duplicate `actionId`s across menus or features. `registerAction` warns but proceeds to overwrite. During `initializeAppFromConfig`, multiple menus can define actions with the same `id`, or user-defined IDs may conflict with auto-registered ones like `"main-menu"`, `"help"`, `"back-to-main"`. This can lead to incorrect handler execution at runtime when an intent arrives for a duplicated ID. <b>[ Low confidence ]</b>
- [line 23](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L23): Unbounded growth of `actionHandlers` leads to a memory leak. Functions like `sendConfirmation` and `sendSelection` register per-interaction action IDs (e.g., using `Date.now()` or caller-provided IDs) via `registerAction`, but there is no corresponding deregistration after the intent is processed. Over time, the global `actionHandlers` `Map` accumulates entries and is never cleared (except manually via `clearAllActions`), causing memory usage to grow with user interactions. <b>[ Low confidence ]</b>
- [line 31](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L31): Global `lastSentActionMessage` is shared across all conversations and invocations, causing races and cross-conversation/state bleed. Concurrent sends will overwrite the value, and `getLastSentActionMessage()` returns whichever message was last sent globally rather than the one relevant to the current `ctx`. This violates at-most-once and uniqueness/association invariants and can surface the wrong message to callers. <b>[ Low confidence ]</b>
- [line 48](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L48): `showLastMenu` claims to "Fallback to main menu" in the comment, but when no last menu is tracked it only sends a text message `"Returning to main menu..."` and does not actually show the main menu or any actions. This contradicts the documented intent and yields a degraded UX where users cannot navigate. <b>[ Low confidence ]</b>
- [line 56](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L56): `inlineActionsMiddleware` detects intents via `ctx.message.contentType?.typeId === "intent"`. This brittle check can miss valid intents or misclassify messages if the intent content type differs by `authorityId` or versioning semantics. The SDK typically compares via `sameAs(ContentTypeIntent)`, which accounts for the full content type identifier. Using only `typeId` risks contract mismatches across versions. <b>[ Low confidence ]</b>
- [line 97](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L97): `ActionBuilder.add` allows duplicate `id`s within a single `ActionsContent` without validation. Inline actions typically require unique IDs per menu to correctly route intents. Duplicates create ambiguity and can silently misroute or ignore user selections, violating sequence uniqueness constraints. <b>[ Low confidence ]</b>
- [line 115](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L115): `lastSentActionMessage` is set in `ActionBuilder.send`, which will overwrite the global value on every send. Under concurrent or multi-conversation use, this causes races and wrong association when retrieved via `getLastSentActionMessage()`. Each send should either return the message to the caller or store per-conversation/per-user state to avoid cross-talk. <b>[ Low confidence ]</b>
- [line 124](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L124): `sendActions` also sets the same global `lastSentActionMessage`, producing the same cross-conversation/race issues. Multiple send paths compete to set this shared value unexpectedly. <b>[ Low confidence ]</b>
- [line 176](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L176): Both validators use `input.trim()` for validation but do not expose or return the normalized value. This can cause a mismatch where an input with leading/trailing whitespace (e.g., `' 0xabc...'`) is deemed valid, yet downstream code consuming the original untrimmed value may fail. To preserve invariants, either return the trimmed value alongside `valid`, or require callers to supply already-normalized input and avoid trimming here. <b>[ Low confidence ]</b>
- [line 182](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L182): `ethereumAddress` validator accepts any `0x` + 40 hex characters and does not enforce EIP-55 checksum casing. If downstream components require checksum addresses, this validator will incorrectly mark non-checksummed addresses as valid, enabling malformed inputs to pass validation. <b>[ Low confidence ]</b>
- [line 290](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/inline-actions.ts#L290): `showNavigationOptions` can send an `ActionsContent` with zero `actions` if neither `customActions` is provided nor a `main-menu` exists in `config`. This produces an unusable navigation message and may violate the target contract if empty action lists are not allowed, yielding a poor or invalid user artifact without fallback. <b>[ Low confidence ]</b>
</details>

<details>
<summary>sdks/agent-sdk/src/content-types/inline-actions/types/ActionsContent.ts — 0 comments posted, 6 evaluated, 6 filtered</summary>

- [line 70](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/types/ActionsContent.ts#L70): `decode` blindly reads `content.parameters.encoding` without checking that `content.parameters` exists. If `parameters` is undefined (which is possible for `EncodedContent` produced by other codecs or sources), this causes a runtime `TypeError: Cannot read properties of undefined (reading 'encoding')`. Add a guard, e.g., `const encoding = content.parameters?.encoding;`. <b>[ Low confidence ]</b>
- [line 71](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/types/ActionsContent.ts#L71): `decode` only accepts exactly `"UTF-8"` for the `encoding` parameter and throws otherwise. Other producers may emit variants like `"utf-8"`, `"UTF8"`, or omit the parameter. The strict case-sensitive check can cause unnecessary runtime errors and interoperability issues. Normalizing case or accepting common aliases would avoid failures. <b>[ Low confidence ]</b>
- [line 138](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/types/ActionsContent.ts#L138): `validateContent` rejects `action.style === "secondary"` even though the `Action.style` type includes `"secondary"`. The check `if (action.style && !["primary", "danger"].includes(action.style))` omits `"secondary"`, and the error message itself claims allowed values are `primary, secondary, danger`. This will throw on valid inputs and contradicts the declared type and message. <b>[ Already posted ]</b>
- [line 152](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/types/ActionsContent.ts#L152): `validateContent` rejects the valid `Action.style` value `"secondary"`. The type `Action` declares `style?: "primary" | "secondary" | "danger"`, and the error message says it must be one of `primary, secondary, danger`, but the check at `validateContent` only allows `["primary", "danger"]`. This will cause a runtime error when `style` is `"secondary"`. Update the inclusion check to allow `"secondary"` as declared. <b>[ Already posted ]</b>
- [line 171](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/types/ActionsContent.ts#L171): `isValidISO8601` is overly strict: it creates `new Date(timestamp)` and requires `date.toISOString() === timestamp`. This rejects many valid ISO-8601 strings (e.g., those with timezone offsets `+01:00`, without milliseconds, or lowercase `z`). As a result, valid `expiresAt` values for actions and content are incorrectly rejected, causing unnecessary runtime errors in `validateContent`. Consider a real ISO-8601 parser or a more permissive check. <b>[ Already posted ]</b>
- [line 188](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/types/ActionsContent.ts#L188): `isValidISO8601` uses `new Date(timestamp)` and compares `date.toISOString() === timestamp`. This is overly strict and will reject many valid ISO-8601/RFC 3339 timestamps (e.g., those without milliseconds like `"2023-08-01T12:00:00Z"`, or with timezone offsets like `"2023-08-01T12:00:00+02:00"`). Given the function is labeled "Basic ISO-8601 timestamp validation", this strict equality causes avoidable runtime failures on valid inputs. Consider a more permissive validation (e.g., parsing and checking for validity without requiring exact `toISOString()` equivalence). <b>[ Already posted ]</b>
</details>

<details>
<summary>sdks/agent-sdk/src/content-types/inline-actions/types/IntentContent.ts — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 49](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/types/IntentContent.ts#L49): `decode` does not verify that `content.type` matches `ContentTypeIntent` before decoding. If a mismatched content type is passed, it may produce confusing errors or incorrectly accept malformed data. Add a type guard to ensure only `ContentTypeIntent` is decoded. <b>[ Low confidence ]</b>
- [line 50](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/types/IntentContent.ts#L50): `decode` dereferences `content.parameters` without a null/undefined guard: `const encoding = content.parameters.encoding;`. If `parameters` is absent (which is allowed in `EncodedContent`), this throws at runtime. Add a safe access or default before reading `encoding`. <b>[ Low confidence ]</b>
- [line 51](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/content-types/inline-actions/types/IntentContent.ts#L51): `decode` enforces a case-sensitive check for `encoding` equal to `"UTF-8"` and rejects other common variants like `"utf-8"`. This can cause valid content from other senders to be rejected unnecessarily. Normalize or case-insensitively compare the `encoding` value. <b>[ Low confidence ]</b>
</details>

<details>
<summary>sdks/agent-sdk/src/core/Agent.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 436](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/core/Agent.ts#L436): `start()` sets `#isLocked = true` at entry and, on failure, calls `this.#handleStreamError(error)` at line 436. If the error chain does not indicate recovery (`recovered === false` inside `#handleStreamError`), the lock is never cleared and no restart is scheduled, leaving the agent permanently locked with no active streams. Subsequent `start()` calls will early-return due to `#isLocked` being `true`. The lock must be cleared on all non-recovery paths or `stop()` invoked to restore a usable state. <b>[ Out of scope ]</b>
- [line 436](https://github.com/xmtp/xmtp-js/blob/ff73769b7f54f400b83b01a04771b082268a39d2/sdks/agent-sdk/src/core/Agent.ts#L436): When `start()` fails and delegates to `#handleStreamError` at line 436, the automatic restart path inside `#handleStreamError` uses `queueMicrotask(() => this.start())` without forwarding the original `AgentStreamingOptions`. This silently drops caller-provided streaming options on recovery, changing externally visible behavior (contract parity) after an initialization error. <b>[ Out of scope ]</b>
</details>

</details>

### DIFF
--- a/sdks/agent-sdk/src/core/filter.ts
+++ b/sdks/agent-sdk/src/core/filter.ts
@@ -37,10 +37,10 @@ import {
   type DecodedMessage,
 } from "@xmtp/node-sdk";
 import {
-  ActionsCodec,
   ContentTypeActions,
   ContentTypeIntent,
-  IntentCodec,
+  type ActionsCodec,
+  type IntentCodec,
 } from "@/inline-actions.js";
 
 export type DecodedMessageWithContent<ContentTypes = unknown> =


bennycode:
@humanagent there's a `sendIntent` method now on the `Conversation` class. We are exposing it in the Agent SDK v2. Are you okay with closing this PR in favor of https://github.com/xmtp/xmtp-js/pull/1682/files?

humanagent:
very nice! @bennycode , feel free to close it.

PR #1576, opened 2025-12-04
Add auto-revoke command to xmtp-cli.

humanagent:
### Add auto-revoke support to `packages/xmtp-cli/commands/revoke.ts` and fix debug conversation counts in `packages/xmtp-cli/commands/debug.ts` to improve xmtp-cli debug tools
Introduce `--all` and optional `[inbox-id]` to `revoke` with auto-detection of the current inbox; add `debug.getConversationBreakdown` and use SDK `filter.isDM` for accurate totals; update README revoke docs and remove unused CLI scripts.

#### 📍Where to Start
Start with `revoke.runRevokeCommand` in [packages/xmtp-cli/commands/revoke.ts](https://github.com/xmtp/xmtp-js/pull/1576/files#diff-9bc9792c557ff928c0f684926e6bb0ec05571d321197e6ab60d836241694c8be), then review `debug.getConversationBreakdown` in [packages/xmtp-cli/commands/debug.ts](https://github.com/xmtp/xmtp-js/pull/1576/files#diff-88a3ed080eddef730dc110e26b309780a27660a3033565dc8f610e256b314aae).

#### Changes since #1576 opened

- Removed automatic inbox resolution functionality and --all flag from revoke command [942119b]
- Changed default installation retention behavior when no installations are specified [942119b]
- Updated environment resolution to default to dev environment instead of exiting on missing configuration [942119b]
- Improved .env file parsing logic [942119b]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 942119b. 2 files reviewed, 8 issues evaluated, 7 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>packages/xmtp-cli/commands/debug.ts — 0 comments posted, 4 evaluated, 3 filtered</summary>

- [line 478](https://github.com/xmtp/xmtp-js/blob/942119b76e0f9dcc3d0a805caa3b06248282a48d/packages/xmtp-cli/commands/debug.ts#L478): `listAllConversations` prints aggregate stats from `getConversationBreakdown(agent)` and then separately fetches `conversations` again (`agent.client.conversations.list()`). Between these two calls, the set of conversations can change, leading to inconsistent totals vs. the detailed list. Reuse the same snapshot (e.g., return the conversations from `getConversationBreakdown` or avoid a second fetch) to ensure the stats and details match. <b>[ Low confidence ]</b>
- [line 490](https://github.com/xmtp/xmtp-js/blob/942119b76e0f9dcc3d0a805caa3b06248282a48d/packages/xmtp-cli/commands/debug.ts#L490): `listAllConversations` selects the "last" message as `messages[0]` without any guarantee of sort order from `conv.messages()`. If messages are returned oldest-first, this will pick the first (oldest) message instead of the most recent one, skewing `lastMessage` content and time and the subsequent sorting by `sentAt`. Use `messages[messages.length - 1]` or request/rely on a documented newest-first order. <b>[ Low confidence ]</b>
- [line 542](https://github.com/xmtp/xmtp-js/blob/942119b76e0f9dcc3d0a805caa3b06248282a48d/packages/xmtp-cli/commands/debug.ts#L542): `timeAgo` can be negative if `stat.lastMessage.sentAt` is in the future or local clock skew exists. This yields outputs like `"-5m ago"`. Clamp at zero or handle future timestamps (e.g., display `in Xm`). <b>[ Low confidence ]</b>
</details>

<details>
<summary>packages/xmtp-cli/commands/revoke.ts — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 77](https://github.com/xmtp/xmtp-js/blob/942119b76e0f9dcc3d0a805caa3b06248282a48d/packages/xmtp-cli/commands/revoke.ts#L77): Platform-specific bug: `exampleName` is derived using `process.cwd().split("/").pop()` which fails on Windows paths (backslashes), yielding the entire path rather than the directory name. Use `path.basename(process.cwd())` for cross-platform correctness. <b>[ Out of scope ]</b>
- [line 151](https://github.com/xmtp/xmtp-js/blob/942119b76e0f9dcc3d0a805caa3b06248282a48d/packages/xmtp-cli/commands/revoke.ts#L151): Possible unsafe array access: `inboxState[0].installations` is accessed without verifying that `inboxState` contains at least one element. If `Client.inboxStateFromInboxIds([inboxId], env)` returns an empty array (e.g., unknown `inboxId` or environment), `inboxState[0]` is `undefined` and accessing `.installations` will throw. The same pattern occurs later with `finalInboxState[0].installations`. Add explicit length checks and produce a clear error instead of relying on the broad `catch`. <b>[ Low confidence ]</b>
- [line 189](https://github.com/xmtp/xmtp-js/blob/942119b76e0f9dcc3d0a805caa3b06248282a48d/packages/xmtp-cli/commands/revoke.ts#L189): Misleading contract: when no `--keep` is provided, the code states it will keep "the current installation" (`console.log` at `lines 189-191` and `line 138`), but it actually keeps `currentInstallations[0].id` (`line 206`). Unless the first element is guaranteed to be the current installation (no such guarantee is shown), the implementation contradicts the message and may revoke the current installation while keeping another one. Either determine the actual current installation explicitly or change the user-facing text. <b>[ Low confidence ]</b>
- [line 257](https://github.com/xmtp/xmtp-js/blob/942119b76e0f9dcc3d0a805caa3b06248282a48d/packages/xmtp-cli/commands/revoke.ts#L257): Possible unsafe array access (second occurrence): `finalInboxState[0].installations.length` is accessed without ensuring `finalInboxState` is non-empty. If the inbox disappears or the API returns an empty array, this will throw. Add a length check and handle gracefully. <b>[ Already posted ]</b>
</details>

</details>

### DIFF
--- a/packages/xmtp-cli/commands/revoke.ts
+++ b/packages/xmtp-cli/commands/revoke.ts
@@ -1,36 +1,25 @@
 import { existsSync } from "node:fs";
 import { readFile } from "node:fs/promises";
 import { join } from "node:path";
-import {
-  createIdentifier,
-  createSigner,
-  createUser,
-} from "@xmtp/agent-sdk/user";
-import { Client, getInboxIdForIdentifier, type XmtpEnv } from "@xmtp/node-sdk";
+import { createSigner, createUser } from "@xmtp/agent-sdk/user";
+import { Client, type XmtpEnv } from "@xmtp/node-sdk";
 import type { Argv } from "yargs";
 
 export interface RevokeOptions {
   keep?: string;
   env?: string;
-  all?: boolean;
 }
 
 export function registerRevokeCommand(yargs: Argv) {
   return yargs.command(
-    "revoke [inbox-id]",
+    "revoke <inbox-id>",
     "Revoke XMTP installations for an inbox",
     (yargs: Argv) => {
       return yargs
         .positional("inbox-id", {
           type: "string",
           description: "64-character hex inbox ID",
-        })
-        .option("all", {
-          type: "boolean",
-          alias: "a",
-          description:
-            "Revoke all installations for current inbox (gets inboxId automatically)",
-          default: false,
+          demandOption: true,
         })
         .option("keep", {
           type: "string",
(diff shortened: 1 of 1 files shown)


packages/xmtp-cli/commands/revoke.ts
```diff
@@ -28,21 +39,104 @@ export function registerRevokeCommand(yargs: Argv) {
...
+    const envVars: Record<string, string> = {};
+
+    envContent.split("\n").forEach((line) => {
+      const [key, value] = line.split("=");
+      if (key && value && !key.startsWith("#")) {
+        envVars[key.trim()] = sanitizeEnvValue(value);
+      }
+    });
+
+    const env = (options.env ||
```
bennycode: What's the difference here between `envVars` & `process.env`?

humanagent: env vars are cli variables like --inboxid, or --env, they override .env file (if specified)
much faster to tweak env vars than modifying files. also for parallel work

PR #1563, opened 2025-11-24
Normalize ens input names to lowercase

humanagent:
Normalize ENS/Base names to lowercase for case-insensitive matching in xmtp.chat input fields.

This ensures that user-provided ENS/Base names, regardless of casing, correctly resolve to the intended profile when creating direct messages, group conversations, or adding members.

---
[Slack Thread](https://xmtp-labs.slack.com/archives/C09UWPZCGQ6/p1764007061478729?thread_ts=1764007061.478729&cid=C09UWPZCGQ6)

<a href="https://cursor.com/background-agent?bcId=bc-c18179c6-8f81-402f-b726-ec1c9e8fd243"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-c18179c6-8f81-402f-b726-ec1c9e8fd243"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>

PR #1582, opened 2025-12-08
Support ENS resolution in DM links

bennycode:
### Add ENS name resolution to DM loading in `LoadDM` to support ENS in direct message links
Introduce `resolveAddress` in `LoadDM` to resolve ENS names via `resolveNameQuery`, route non-`0x` inputs through resolution with status messaging and redirects on failure, and use the resolved lowercased address for `client.findInboxIdByIdentifier` and `client.conversations.newDmWithIdentifier`. Centralize redirect messaging with `navigateToHome`.

#### 📍Where to Start
Start in the `LoadDM` component at the ENS resolution path and helpers in [LoadDM.tsx](https://github.com/xmtp/xmtp-js/pull/1582/files#diff-f999c8ef5e116330d863dbca66126681e618fa2e7c75744d86705106e220bc74), beginning with `resolveAddress` and its calls within the DM load flow.

#### Changes since #1582 opened

- Normalized name resolution and storage to use lowercase identities [df87bac]
- Replaced heuristic address validation in `LoadDM.resolveAddress` with `isValidEthereumAddress()` function and gated ENS/name resolution attempts behind `isValidName()` validation check [749dfa2]
- Modified `LoadDM.resolveAddress` to return null on all failure cases without triggering navigation side-effects [749dfa2]
- Added input sanitization to the `useMemberId` hook's `setMemberId` function [409ea03]
- Refactored `setMemberId` state setter in `useMemberId` hook by removing `useCallback` memoization and inlining whitespace-stripping logic previously contained in `sanitizeMemberId` helper function [86bc148]
- Prevented space character input via keyboard in `memberId` input fields [bf31aa3]
- Modified `setMemberId` function in `useMemberId` hook to preserve internal whitespace [bf31aa3]
- Modified profile lookup flow to handle case-insensitive ENS name resolution [cfd9bc1]
- Moved name normalization from client to server by implementing lowercase conversion in the `resolveName` handler before schema validation while removing the local lowercase normalization from the `resolveName` helper [13ae527]
- Added documentation clarifying the service acts as a server-side intermediary to external services and enumerated supported features [365e38c]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 365e38c. 6 files reviewed, 2 issues evaluated, 2 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>apps/xmtp.chat/src/components/Conversation/LoadDM.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 120](https://github.com/xmtp/xmtp-js/blob/365e38cb93bbc57855645b8e067c5254a705fd5e/apps/xmtp.chat/src/components/Conversation/LoadDM.tsx#L120): The `useEffect` dependency array on line 120 is missing `addConversation` and `navigate`. While these are typically stable references, omitting them violates React hooks rules and could theoretically cause stale closure issues if the references ever change during the component lifecycle. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>apps/xmtp.chat/src/components/Conversations/CreateDmModal.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 59](https://github.com/xmtp/xmtp-js/blob/365e38cb93bbc57855645b8e067c5254a705fd5e/apps/xmtp.chat/src/components/Conversations/CreateDmModal.tsx#L59): The `useMemo` dependency array on line 59 is missing `memberIdError` and `inboxId`, which are used in the `disabled` prop on line 52. When these values change, the memoized `footer` will not re-compute, causing the button's disabled state to be stale. Users may see an enabled button when it should be disabled or vice versa. <b>[ Out of scope ]</b>
</details>

</details>

### DIFF
--- a/apps/xmtp.chat-api-service/README.md
+++ b/apps/xmtp.chat-api-service/README.md
@@ -1,6 +1,16 @@
 # xmtp.chat API service
 
-This package contains code for the xmtp.chat API service.
+This package contains code for the xmtp.chat API service, a server-side intermediary between the xmtp.chat frontend and external services (web3.bio, Pinata, etc.).
+
+## Featues
+
+The xmtp.chat API service is a backend service that provides the following:
+
+- **Name Resolution**: Resolves ENS and Base names to Ethereum addresses and profile information using [web3.bio](https://web3.bio/)
+- **Profile Caching**: Caches resolved profiles in a PostgreSQL database to reduce external API calls
+- **File Storage**: Integrates with [Pinata](https://pinata.cloud/) for IPFS file uploads and management
+- **Rate Limiting**: Protects endpoints from abuse with configurable rate limits
+- **CORS Support**: Handles cross-origin requests from the xmtp.chat frontend
 
 ### Get started
 


apps/xmtp.chat/src/components/Conversation/LoadDM.tsx
```diff
@@ -18,24 +19,62 @@ export const LoadDM: React.FC = () => {
...
+      setMessage(message);
+      timeout = setTimeout(() => {
+        void navigate(`/${environment}`);
+      }, REDIRECT_TIMEOUT);
+    };
+
+    const resolveAddress = async (
+      addressOrENS: string,
+    ): Promise<string | null> => {
+      if (!addressOrENS.startsWith("0x")) {
```
rygine: we should make sure it also ends with `.eth` or `.base.eth` like [this function](https://github.com/xmtp/xmtp-js/blob/main/apps/xmtp.chat-api-service/src/helpers/web3.bio.ts#L9-L11) we use to validate names in the API service.

bennycode: Ok! I will also simplify the resolver so it doesn't redirect because that is already done by the caller of it.

bennycode:
PR also includes work from #1563

apps/xmtp.chat/src/components/Conversation/AddMembers.tsx
```diff
@@ -99,6 +99,10 @@ export const AddMembers: React.FC<AddMembersProps> = ({
             setMemberId(event.target.value);
           }}
           onKeyDown={(event) => {
+            if (event.key === " ") {
+              event.preventDefault();
+              return;
+            }
```
rygine, line 105: i don't think these are necessary. as soon as a space is typed, the input will become invalid. also, this doesn't prevent pasting a string with spaces.

apps/xmtp.chat/src/hooks/useMemberId.ts
```diff
@@ -19,6 +19,10 @@ export const useMemberId = () => {
   const [error, setError] = useState<string | null>(null);
   const { environment } = useSettings();
 
+  const setMemberId = (value: string) => {
+    setMemberIdState(value.trim());
+  };
+
```
rygine, line 25: this is not necessary as the `memberId` will not be valid with spaces

apps/xmtp.chat/src/helpers/names.ts
```diff
@@ -23,14 +23,18 @@ export const resolveName = async (name: string, force: boolean = false) => {
     return null;
   }
 
+  const normalizedName = name.toLowerCase();
```
rygine: this logic can be moved into the `getProfilesByName` method

apps/xmtp.chat/src/helpers/names.ts
```diff
@@ -23,14 +23,18 @@ export const resolveName = async (name: string, force: boolean = false) => {
...
+  const cachedProfiles = profilesStore
+    .getState()
+    .getProfilesByName(normalizedName);
   if (!force && cachedProfiles.length > 0) {
     return cachedProfiles;
   }
 
   const response = await fetch(
-    `${import.meta.env.VITE_API_SERVICE_URL}/api/v2/resolve/name/${window.encodeURIComponent(name)}`,
+    `${import.meta.env.VITE_API_SERVICE_URL}/api/v2/resolve/name/${window.encodeURIComponent(normalizedName)}`,
```
rygine: the normalization can occur on the API service. doing it here doesn't really matter as we shouldn't trust clients.

apps/xmtp.chat/src/stores/profiles.ts
```diff
@@ -59,7 +59,7 @@ export const profilesStore = createStore<ProfilesState & ProfilesActions>()(
       const existingProfiles = state.profiles.get(profile.address) ?? [];
       newProfiles.set(profile.address, [...existingProfiles, profile]);
       if (profile.identity) {
-        newNames.set(profile.identity, profile.address);
+        newNames.set(profile.identity.toLowerCase(), profile.address);
```
rygine, line 62: we can also handle this case in the API service, but i'm not sure that would be necessary if ENS/base names are always lowercase.

apps/xmtp.chat/src/components/Conversation/AddMembers.tsx
bennycode: The input will become invalid as soon as a space is typed but I won't to prevent the users from even running into this issue. I have seen this happening when people type from their mobile phones (as the phone's auto correct is adding spaces).

apps/xmtp.chat/src/stores/profiles.ts
bennycode: I want to fix this UX issue:

image

bennycode:
Closing this PR in favor of https://github.com/xmtp/xmtp-js/pull/1595

Will do the DM link feature in a follow-up PR.

PR #1589, opened 2025-12-09
Resolve non-accessible links

ylmin:
Resolve non-accessible links


bennycode:
Hi @ylmin, thank you for this contribution. 👍

PR #1618, opened 2026-01-06
1.8 release

rygine:
### Release 1.8 by refactoring SDK APIs to use `CodecRegistry`, renaming conversation and preferences methods across `sdks/node-sdk` and `sdks/browser-sdk`, and adding an OPFS worker with `opfs.*` actions
Introduce `CodecRegistry` and shift encoding/decoding to codecs, rename core APIs (`conversations.create*`, `Client.fetchInboxIdByIdentifier`, `Preferences.fetchInboxStates`), add granular send methods, and implement an OPFS worker with typed `opfs.*` actions; update barrels, enums, and tests to new forms.

#### 📍Where to Start
Start with the API surface changes in `Client` and conversation layers: review `sdks/browser-sdk/src/Client.ts` (worker bridge, getters, renamed methods) and `sdks/node-sdk/src/Client.ts` (codec registry, renamed fetch methods). Then examine `sdks/browser-sdk/src/workers/client.ts` for action routing and `sdks/browser-sdk/src/workers/opfs.ts` for OPFS worker initialization and actions.

#### Changes since #1618 opened

- Updated error handling in `DecodedMessage` constructor's custom content decoding catch block across both browser and node SDKs to conditionally log error messages [1734f12]
- Removed module-level initialization caching from `inboxState` module in `browser-sdk` package [ef6ef12]
- Added decoding logic for custom reply content in `DecodedMessage` class [3e92db1]
- Added `DecodeFailureCodec` test helper class [3e92db1]
- Expanded test coverage for custom content decoding behavior [3e92db1]
- Fixed reply message decoding to use the embedded custom content type from the reply payload instead of the outer message content type [8606200]
- Added async `metadataFieldName` utility function to `@xmtp/browser-sdk` [8606200]
- Added default case with never assertion to `getContentFromDecodedMessageContent` switch statement [8606200]
- Updated bindings dependencies from version 1.8.0 to 1.8.1 [8606200]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 3e92db1. 35 files reviewed, 21 issues evaluated, 19 issues filtered, 1 comment posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>sdks/browser-sdk/src/Conversations.ts — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 132](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/Conversations.ts#L132): In `getDmByInboxId` at line 132, the `Dm` class is instantiated without the generic type parameter (`new Dm(...)` instead of `new Dm<ContentTypes>(...)`), unlike all other instantiations in this class which use the generic. While TypeScript will infer the type, this inconsistency could cause type inference issues in consuming code. <b>[ Low confidence ]</b>
- [line 423](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/Conversations.ts#L423): The `convertConversation` function at line 406-427 throws an `Error` for unknown conversation types in the `default` case, but other methods handling conversation types (`list` at lines 163-184, `getConversationById` at lines 84-101) return `undefined` for unknown types instead. This inconsistency means streams will crash with an unhandled exception when encountering an unknown conversation type, while list operations will silently filter them out. <b>[ Low confidence ]</b>
- [line 568](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/Conversations.ts#L568): The `stream` function in `streamMessageDeletions` does not accept the `onFail` callback parameter that `createStream` passes to it. This means when the stream fails in the worker (sending a `stream.fail` action), the retry mechanism in `createStream` will never be triggered. Compare lines 568-580 with lines 382-404 where `onFail` is properly accepted and passed to `handleStreamMessage`. The stream function signature should be `async (callback: StreamCallback<string>, onFail: () => void)` and the `handleStreamMessage` call should include `{ ...options, onFail }` instead of just `options`. <b>[ Already posted ]</b>
</details>

<details>
<summary>sdks/browser-sdk/src/DecodedMessage.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 66](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/DecodedMessage.ts#L66): The switch statement in `getContentFromDecodedMessageContent` (lines 19-65) is not exhaustive - if `DecodedMessageContent` adds a new type variant that isn't handled, the function silently returns `null as T` (line 66) instead of properly indicating an unhandled case. This could cause subtle runtime bugs where content appears as `null` rather than failing explicitly. <b>[ Low confidence ]</b>
- [line 136](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/DecodedMessage.ts#L136): When handling a reply message with custom inner content (line 135-147), the codec lookup at line 136 uses `this.contentType` which is the outer message's content type (set at line 114 from `message.contentType`). For a reply message, this would be the reply content type, not the content type of the custom content nested inside the reply. This means the wrong codec will be looked up, causing decoding to fail or produce incorrect results. <b>[ Already posted ]</b>
</details>

<details>
<summary>sdks/browser-sdk/src/utils/createClient.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 51](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/utils/createClient.ts#L51): The `LogOptions` construction changed from `new LogOptions(...)` (class constructor) to an object literal with type assertion `({...} as LogOptions)`. If `LogOptions` from `@xmtp/wasm-bindings` is a class with methods or specific internal state rather than a plain interface, the WASM bindings will receive an incompatible object at runtime, potentially causing crashes or undefined behavior when the bindings attempt to use it. <b>[ Low confidence ]</b>
</details>

<details>
<summary>sdks/browser-sdk/src/utils/installations.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 39](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/utils/installations.ts#L39): The new `revokeInstallationsSignatureText` function does not compute or pass an `isSecure` parameter to `revokeInstallationsSignatureRequest`, while the similar `getInboxIdForIdentifier` call in `createClient` was updated to include `isSecure`. If `revokeInstallationsSignatureRequest` also requires this parameter for proper TLS handling (as part of the same API update), the function signature may be incorrect and could fail at runtime. <b>[ Low confidence ]</b>
</details>

<details>
<summary>sdks/browser-sdk/src/workers/client.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 542](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/workers/client.ts#L542): In `conversations.streamAllMessages` (lines 541-553) and `conversation.stream` (lines 937-949), the async call to `client.conversations.getMessageById(value.id)` uses `void` to discard the promise but has no `.catch()` handler. If `getMessageById` throws an error, it will result in an unhandled promise rejection rather than being reported through `postStreamMessageError`. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>sdks/browser-sdk/src/workers/opfs.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 49](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/workers/opfs.ts#L49): The `await init()` call on line 49 is outside the try-catch block. If the WASM module fails to initialize (e.g., network error loading WASM, browser incompatibility), the error is not caught and no error response is sent via `postMessageError`. The caller will hang waiting for a response that never arrives. <b>[ Already posted ]</b>
- [line 69](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/browser-sdk/src/workers/opfs.ts#L69): When `opfs.init` is called after the OPFS has already been initialized, no response is sent to the caller. The first `if` block (line 53-62) is skipped because `initialized` is `true`, then the second `if` at line 69 returns without posting any message. This will cause the caller to hang indefinitely waiting for a response. <b>[ Already posted ]</b>
</details>

<details>
<summary>sdks/node-sdk/src/Client.ts — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 52](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/node-sdk/src/Client.ts#L52): The `CodecRegistry` is initialized with only user-provided codecs (`options?.codecs ?? []`), but the previous implementation always included default codecs (`GroupUpdatedCodec` and `TextCodec`). This means clients created without explicit codecs will have an empty registry, causing message encoding/decoding operations that rely on these default codecs to fail. <b>[ Low confidence ]</b>
- [line 266](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/node-sdk/src/Client.ts#L266): In `unsafe_addSignature`, the check at line 266-268 throws `SignerUnavailableError` if `this.#signer` is undefined, even when a valid `signer` parameter is provided. The condition should be `if (!signer && !this.#signer)` to allow the optional `signer` parameter to be used when the instance signer is not set. <b>[ Out of scope ]</b>
- [line 772](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/node-sdk/src/Client.ts#L772): In `Client.canMessage`, accessing `identifier.identifier.toLowerCase()` will throw a `TypeError` if `identifier.identifier` is `undefined` or `null`. The code does not validate that `identifier.identifier` exists before calling `.toLowerCase()` on it. <b>[ Low confidence ]</b>
</details>

<details>
<summary>sdks/node-sdk/src/Conversation.ts — 0 comments posted, 3 evaluated, 2 filtered</summary>

- [line 155](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/node-sdk/src/Conversation.ts#L155): The `stream` method calls `this.#client.conversations.getMessageById(value.id)` at line 155-156, but this method is not visible in the provided `Client` or `Conversations` class definitions. If `getMessageById` does not exist on the `conversations` object, this will throw a `TypeError` at runtime when the stream receives a message. <b>[ Low confidence ]</b>
- [line 158](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/node-sdk/src/Conversation.ts#L158): The `convertMessage` function in `stream()` returns `enrichedMessage ?? value` where `value` is a raw `Message` type. If `getMessageById` returns `undefined` or `null`, the stream will emit raw `Message` objects instead of `DecodedMessage` objects. Consumers iterating over the stream may receive inconsistent object structures (sometimes `DecodedMessage` with decoded `content`, sometimes raw `Message` with encoded content), which could cause runtime errors if code accesses properties like `content` expecting the decoded format. <b>[ Low confidence ]</b>
</details>

<details>
<summary>sdks/node-sdk/src/DecodedMessage.ts — 1 comment posted, 2 evaluated, 1 filtered</summary>

- [line 130](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/node-sdk/src/DecodedMessage.ts#L130): On line 130, `message.content.reply` is cast to `EnrichedReply` without null checking. If `message.content.reply` is `null` or `undefined` when `message.content.type === DecodedMessageContentType.Reply`, subsequent access to `reply.content` on line 131-132 will throw a runtime error. <b>[ Low confidence ]</b>
</details>

<details>
<summary>sdks/node-sdk/src/index.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 15](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/node-sdk/src/index.ts#L15): The type `EncodedContent` was previously exported but has been removed from the exports. If any consuming code imports `EncodedContent` from this package, it will fail with a module resolution error at build time. <b>[ Already posted ]</b>
- [line 15](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/node-sdk/src/index.ts#L15): The type `ContentTypeId` was previously exported but has been removed from the exports. If any consuming code imports `ContentTypeId` from this package, it will fail with a module resolution error at build time. <b>[ Low confidence ]</b>
</details>

<details>
<summary>sdks/node-sdk/test/helpers.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 45](https://github.com/xmtp/xmtp-js/blob/3e92db1144b7ccc0d78dc85fdaa61555aa035464/sdks/node-sdk/test/helpers.ts#L45): The `createSigner` function signature changed from `(user: User): Signer` to `(): { address, identifier, signer, user }`. Any existing call sites that pass a `user` argument will have it silently ignored (the passed user is never used), and code expecting the return value to be a `Signer` directly will fail at runtime when trying to call methods like `.signMessage()` or `.getIdentifier()` on the returned object, since those methods now live on the `.signer` property instead. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/sdks/browser-sdk/src/index.ts
+++ b/sdks/browser-sdk/src/index.ts
@@ -6,6 +6,7 @@ export { Dm } from "./Dm";
 export { Group } from "./Group";
 export { DecodedMessage } from "./DecodedMessage";
 export { generateInboxId, getInboxIdForIdentifier } from "./utils/inboxId";
+export { metadataFieldName } from "./utils/metadata";
 export { ApiUrls, HistorySyncUrls } from "./constants";
 export type * from "./types/options";
 export * from "./utils/conversions";
--- a/sdks/browser-sdk/src/utils/metadata.ts
+++ b/sdks/browser-sdk/src/utils/metadata.ts
@@ -1,0 +1,15 @@
+import init, {
+  metadataFieldName as wasmMetadataFieldName,
+  type MetadataField,
+} from "@xmtp/wasm-bindings";
+
+/**
+ * Gets the name of a metadata field
+ *
+ * @param field - The metadata field to get the name for
+ * @returns The name of the metadata field
+ */
+export const metadataFieldName = async (field: MetadataField) => {
+  await init();
+  return wasmMetadataFieldName(field);
+};
--- a/sdks/browser-sdk/test/Dm.test.ts
+++ b/sdks/browser-sdk/test/Dm.test.ts
@@ -3,12 +3,12 @@ import {
   ContentType,
   ConversationType,
   MetadataField,
-  metadataFieldName,
   type GroupUpdated,
   type MessageDisappearingSettings,
 } from "@xmtp/wasm-bindings";
 import { describe, expect, it } from "vitest";
 import type { DecodedMessage } from "@/DecodedMessage";
+import { metadataFieldName } from "@/utils/metadata";
 import { createRegisteredClient, createSigner, sleep } from "@test/helpers";
 
 describe("Dm", () => {
@@ -291,7 +291,7 @@ describe("Dm", () => {
     expect(fieldChange1.content?.metadataFieldChanges).toBeDefined();
     expect(fieldChange1.content?.metadataFieldChanges.length).toBe(1);
     expect(fieldChange1.content?.metadataFieldChanges[0].fieldName).toBe(
-      metadataFieldName(MetadataField.MessageExpirationFromNs),
+      await metadataFieldName(MetadataField.MessageExpirationFromNs),
     );
     expect(fieldChange1.content?.metadataFieldChanges[0].oldValue).toBe("1");
     expect(fieldChange1.content?.metadataFieldChanges[0].newValue).toBe("0");
@@ -300,7 +300,7 @@ describe("Dm", () => {
     expect(fieldChange2.content?.metadataFieldChanges).toBeDefined();
     expect(fieldChange2.content?.metadataFieldChanges.length).toBe(1);
     expect(fieldChange2.content?.metadataFieldChanges[0].fieldName).toBe(
-      metadataFieldName(MetadataField.MessageExpirationInNs),
+      await metadataFieldName(MetadataField.MessageExpirationInNs),
     );
     expect(fieldChange2.content?.metadataFieldChanges[0].oldValue).toBe(
       "2000000000",
--- a/sdks/browser-sdk/test/Group.test.ts
+++ b/sdks/browser-sdk/test/Group.test.ts
@@ -6,7 +6,6 @@ import {
   encodeText,
   GroupMessageKind,
   MetadataField,
-  metadataFieldName,
   ReactionAction,
   ReactionSchema,
   SortDirection,
@@ -19,6 +18,7 @@ import {
   contentTypeGroupUpdated,
   contentTypeLeaveRequest,
 } from "@/utils/contentTypes";
+import { metadataFieldName } from "@/utils/metadata";
 import {
   createRegisteredClient,
   createSigner,
@@ -210,7 +210,7 @@ describe("Group", () => {
     const message = messages[0] as DecodedMessage<GroupUpdated>;
     expect(message.content!.metadataFieldChanges).toHaveLength(1);
     expect(message.content!.metadataFieldChanges[0].fieldName).toBe(
-      metadataFieldName(MetadataField.GroupName),
+      await metadataFieldName(MetadataField.GroupName),
     );
     expect(message.content!.metadataFieldChanges[0].oldValue).toBe("");
     expect(message.content!.metadataFieldChanges[0].newValue).toBe(newName);
@@ -229,7 +229,7 @@ describe("Group", () => {
     const message = messages[0] as DecodedMessage<GroupUpdated>;
     expect(message.content!.metadataFieldChanges).toHaveLength(1);
     expect(message.content!.metadataFieldChanges[0].fieldName).toBe(
-      metadataFieldName(MetadataField.GroupImageUrlSquare),
+      await metadataFieldName(MetadataField.GroupImageUrlSquare),
     );
     expect(message.content!.metadataFieldChanges[0].oldValue).toBe("");
     expect(message.content!.metadataFieldChanges[0].newValue).toBe(imageUrl);
@@ -248,7 +248,7 @@ describe("Group", () => {
     const message = messages[0] as DecodedMessage<GroupUpdated>;
     expect(message.content!.metadataFieldChanges).toHaveLength(1);
     expect(message.content!.metadataFieldChanges[0].fieldName).toBe(
-      metadataFieldName(MetadataField.Description),
+      await metadataFieldName(MetadataField.Description),
     );
     expect(message.content!.metadataFieldChanges[0].oldValue).toBe("");
     expect(message.content!.metadataFieldChanges[0].newValue).toBe(
@@ -269,7 +269,7 @@ describe("Group", () => {
     const message = messages[0] as DecodedMessage<GroupUpdated>;
     expect(message.content!.metadataFieldChanges).toHaveLength(1);
     expect(message.content!.metadataFieldChanges[0].fieldName).toBe(
-      metadataFieldName(MetadataField.AppData),
+      await metadataFieldName(MetadataField.AppData),
     );
     expect(message.content!.metadataFieldChanges[0].oldValue).toBe("");
     expect(message.content!.metadataFieldChanges[0].newValue).toBe(appData);
@@ -809,7 +809,7 @@ describe("Group", () => {
     expect(fieldChange1.content?.metadataFieldChanges).toBeDefined();
     expect(fieldChange1.content?.metadataFieldChanges.length).toBe(1);
     expect(fieldChange1.content?.metadataFieldChanges[0].fieldName).toBe(
-      metadataFieldName(MetadataField.MessageExpirationFromNs),
+      await metadataFieldName(MetadataField.MessageExpirationFromNs),
     );
     expect(fieldChange1.content?.metadataFieldChanges[0].oldValue).toBe("1");
     expect(fieldChange1.content?.metadataFieldChanges[0].newValue).toBe("0");
@@ -818,7 +818,7 @@ describe("Group", () => {
     expect(fieldChange2.content?.metadataFieldChanges).toBeDefined();
     expect(fieldChange2.content?.metadataFieldChanges.length).toBe(1);
     expect(fieldChange2.content?.metadataFieldChanges[0].fieldName).toBe(
-      metadataFieldName(MetadataField.MessageExpirationInNs),
+      await metadataFieldName(MetadataField.MessageExpirationInNs),
     );
     expect(fieldChange2.content?.metadataFieldChanges[0].oldValue).toBe(
       "2000000000",


sdks/browser-sdk/src/Conversation.ts
```diff
@@ -305,51 +464,51 @@ export class Conversation<ContentTypes = unknown> {
...
-      const streamId = v4();
+      const streamId = uuid();
       if (!options?.disableSync) {
         // sync the conversation
         await this.sync();
       }
       // start the stream
-      await this.#client.sendMessage("conversation.stream", {
+      await this.#worker.action("conversation.stream", {
         groupId: this.#id,
```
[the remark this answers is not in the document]
rygine: this is intentional

sdks/browser-sdk/src/Conversations.ts
```diff
@@ -477,13 +566,13 @@ export class Conversations<ContentTypes = unknown> {
...
   ) {
     const stream = async (callback: StreamCallback<string>) => {
-      const streamId = v4();
+      const streamId = uuid();
       // start the stream
-      await this.#client.sendMessage("conversations.streamMessageDeletions", {
+      await this.#worker.action("conversations.streamMessageDeletions", {
         streamId,
       });
       // handle stream messages
```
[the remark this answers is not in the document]
rygine: this stream intentionally does not have a retry mechanism

sdks/browser-sdk/src/types/options.ts
```diff
@@ -87,14 +103,39 @@ export type OtherOptions = {
...
   OtherOptions;
+
+export type Reply<T = unknown, U = unknown> = {
+  referenceId: EnrichedReply["referenceId"];
+  content: T;
+  inReplyTo: DecodedMessage<U> | null;
+};
+
+export type BuiltInContentTypes =
+  | string // text, markdown
```
bennycode, line 120: How about using type aliases here to make this clear in code?

```ts
type Markdown = string;
```

sdks/node-sdk/src/Conversation.ts
```diff
@@ -147,47 +161,177 @@ export class Conversation<ContentTypes = unknown> {
...
-        : // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          this.#client.prepareForSend(content, contentType!);
+  /**
+   * Sends a text message
+   *
+   * @param text - The text to send
+   * @param optimistic - Whether to send the message optimistically
+   * @returns Promise that resolves with the message ID after it has been sent
+   */
+  async sendText(text: string, optimistic?: boolean) {
```
bennycode: According to the SDK style guide this should become `isOptimistic`

sdks/node-sdk/src/Conversation.ts
```diff
@@ -147,47 +161,177 @@ export class Conversation<ContentTypes = unknown> {
...
 
-    return this.#conversation.sendOptimistic(encodedContent, sendOptions);
+  /**
+   * Sends a markdown message
+   *
+   * @param markdown - The markdown to send
+   * @param optimistic - Whether to send the message optimistically
+   * @returns Promise that resolves with the message ID after it has been sent
+   */
+  async sendMarkdown(markdown: string, optimistic?: boolean) {
```
bennycode: Let's name it `isOptimistic`

sdks/node-sdk/src/Conversation.ts
```diff
@@ -147,47 +161,177 @@ export class Conversation<ContentTypes = unknown> {
...
+   * @param reaction - The reaction to send
+   * @param optimistic - Whether to send the message optimistically
    * @returns Promise that resolves with the message ID after it has been sent
-   * @throws {MissingContentTypeError} if content type is required but not provided
    */
-  async send(content: ContentTypes, contentType?: ContentTypeId) {
-    if (typeof content !== "string" && !contentType) {
-      throw new MissingContentTypeError();
-    }
+  async sendReaction(reaction: Reaction, optimistic?: boolean) {
```
bennycode: `isOptimistic`

sdks/node-sdk/test/Conversation.test.ts
```diff
@@ -984,30 +941,27 @@ describe("Conversation", () => {
...
     await sleep(10);
     const timestamp2 = Date.now() * 1_000_000;
     await sleep(10);
-    await conversation.send("text 3");
+    await conversation.sendText("text 3");
 
     // Test different filters against the same message set
     // Total: 5 messages (1 group creation + 4 sent)
-    expect(await conversation.countMessages()).toBe(5);
+    expect(await conversation.countMessages()).toBe(4);
```
bennycode: The comment talks about 5 messages but the test looks for 4?

bennycode (changes requested):

sdks/node-sdk/test/Conversation.test.ts
rygine: this was on an outdated test in an older commit

sdks/node-sdk/src/Conversation.ts
rygine: https://github.com/xmtp/xmtp-js/pull/1618/commits/8dabe384240e38c1549006eaa1fe509306253ad2

sdks/node-sdk/src/Conversation.ts
rygine: https://github.com/xmtp/xmtp-js/pull/1618/commits/8dabe384240e38c1549006eaa1fe509306253ad2

sdks/node-sdk/src/Conversation.ts
rygine: https://github.com/xmtp/xmtp-js/pull/1618/commits/8dabe384240e38c1549006eaa1fe509306253ad2

sdks/browser-sdk/src/types/options.ts
rygine: since markdown can only be a string, i don't think this is necessary. most devs won't see this type spelled out.

sdks/browser-sdk/src/WorkerConversation.ts
```diff
@@ -95,7 +92,7 @@ export class WorkerConversation {
     return this.#group.createdAtNs();
   }
 
-  async lastMessage() {
+  async lastMessage(): Promise<DecodedMessage | undefined> {
```
bennycode: I think the explicit return type isn't necessary anmore.

sdks/node-sdk/scripts/groups.ts
```diff
@@ -59,18 +59,19 @@ console.log(`Created ${groups.length} groups`);
...
 
 while (accounts.length > 0) {
-  const dmGroup = await primaryAccountClient.conversations.newDmWithIdentifier({
-    identifierKind: IdentifierKind.Ethereum,
-    identifier: (accounts.pop() as Account).address,
-  });
+  const dmGroup =
+    await primaryAccountClient.conversations.createDmWithIdentifier({
+      identifierKind: IdentifierKind.Ethereum,
+      identifier: (accounts.pop() as Account).address,
```
bennycode, line 73: `identifier: accounts[accounts.length - 1].address`

sdks/node-sdk/src/utils/date.ts
```diff
@@ -1,3 +1,3 @@
-export function nsToDate(ns: number): Date {
-  return new Date(ns / 1_000_000);
+export function nsToDate(ns: bigint): Date {
```
bennycode, line 1: By the new guidelines this should become `nsToDateRounded`

sdks/node-sdk/scripts/groups.ts
rygine: this is intentional so that the `while` loop completes when there are no items left

sdks/node-sdk/src/utils/date.ts
rygine: makes sense, but this function is for internal use only and not exported from either SDK. should it still be renamed?

sdks/browser-sdk/src/WorkerConversation.ts
rygine: https://github.com/xmtp/xmtp-js/pull/1618/commits/4159182e74c55692698083f561620c21bc579ee4

sdks/browser-sdk/src/workers/client.ts
```diff
@@ -585,12 +538,16 @@ self.onmessage = async (
...
-            postStreamMessage({
-              action: "stream.message",
-              streamId: data.streamId,
-              result: value ? toSafeMessage(value) : undefined,
-            });
+          } else if (value) {
+            void client.conversations
+              .getMessageById(value.id)
+              .then((enrichedMessage) => {
+                postStreamMessage({
```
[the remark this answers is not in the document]
rygine: this code path should only produce a defined value for `enrichedMessage`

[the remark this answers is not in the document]
rygine: while you're technically correct, this code path only results from the message existing in the local database, which will be returned when calling `getMessageById`.

sdks/browser-sdk/test/Opfs.test.ts
```diff
@@ -0,0 +1,139 @@
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { Opfs } from "@/Opfs";
+import { uuid } from "@/utils/uuid";
+import { createRegisteredClient, createSigner } from "./helpers";
+
+describe.skip("Opfs", () => {
```
bennycode, line 6: Can we run the tests in a CI environment? What makes them flaky?

sdks/node-sdk/src/DecodedMessage.ts
```diff
@@ -1,97 +1,157 @@
...
         break;
-      case DeliveryStatus.Failed:
-        this.deliveryStatus = "failed";
+      }
+      case DecodedMessageContentType.Custom: {
+        const customContent = message.content.custom;
+        if (customContent !== null) {
+          const codec = codecRegistry.getCodec<ContentTypes>(this.contentType);
+          if (codec) {
+            this.content = codec.decode(customContent);
```
bennycode: How about we wrap this in a try-catch in case the content causes an error to the decoder?

sdks/browser-sdk/src/workers/opfs.ts
```diff
@@ -0,0 +1,127 @@
...
+  const { action, id, data } = event.data;
+
+  // initialize WASM module
+  await init();
+
+  try {
+    // init is a special action that initializes the client
+    if (action === "opfs.init" && !initialized) {
+      try {
+        await opfsInit();
```
bennycode, line 55: Shall we set an initializing flag before awaiting the `opfsInit` here so that concurrent calls don't pass the same check?

sdks/node-sdk/src/DecodedMessage.ts
```diff
@@ -1,97 +1,157 @@
...
+      return content.actions as T;
+    }
+    case DecodedMessageContentType.Intent: {
+      return content.intent as T;
+    }
+    case DecodedMessageContentType.Custom: {
+      return content.custom as T;
+    }
+  }
+  return null as T;
```
bennycode: Can we turn this into a `default` case?

sdks/browser-sdk/src/DecodedMessage.ts
```diff
@@ -1,89 +1,155 @@
...
+      return content.content as T;
+    }
+    case "actions": {
+      return content.content as T;
+    }
+    case "custom": {
+      return content.content as T;
+    }
+  }
+  return null as T;
```
bennycode: Can we turn this into a `default` case?

sdks/node-sdk/src/DecodedMessage.ts
```diff
@@ -1,97 +1,157 @@
...
-        break;
-      case DeliveryStatus.Published:
-        this.deliveryStatus = "published";
+    this.content =
+      getContentFromDecodedMessageContent<ContentTypes>(message.content) ??
+      undefined;
+
+    switch (message.content.type) {
+      case DecodedMessageContentType.Reply: {
+        const reply = message.content.reply as EnrichedReply;
```
bennycode, line 132: In VS Code it shows me that reply can be `EnrichedReply | null`. I think we can avoid the type assertion but doing an existence check for reply (through an `if` statement for example).

sdks/browser-sdk/src/DecodedMessage.ts
```diff
@@ -143,23 +132,20 @@ export class DecodedMessage<ContentTypes = unknown> {
...
+        const codec = codecRegistry.getCodec<ContentTypes>(this.contentType);
         if (codec) {
-          const encodedContent: EncodedContent = {
-            type: this.contentType,
-            parameters: customContent.parameters,
-            fallback: customContent.fallback,
-            compression: customContent.compression,
-            content: customContent.content,
-          };
           this.content = codec.decode(encodedContent);
```
bennycode: Shall we catch decoder errors?

sdks/node-sdk/scripts/groups.ts
bennycode: Would a simple `for (const account of accounts)` be sufficient in this case?

sdks/browser-sdk/test/contentTypes.test.ts
```diff
@@ -0,0 +1,1137 @@
...
+    expect(message).toBeDefined();
+    expect(message?.content).toEqual({ test: "test" });
+    expect(message?.contentType).toEqual(testCodec.contentType);
+  });
+
+  it("should have undefined content when receiving custom content without codec", async () => {
+    const { signer: signer1 } = createSigner();
+    const { signer: signer2 } = createSigner();
+    const testCodec = new TestCodec();
+    const client1 = await createRegisteredClient(signer1, {
```
bennycode: Let’s name the clients based on their roles (for example `clientWithCodec`) instead of using `client1` and `client2`. Makes the test easier to follow.

sdks/browser-sdk/src/utils/installations.ts
```diff
@@ -0,0 +1,97 @@
...
+  applySignatureRequest,
+  revokeInstallationsSignatureRequest,
+  type Identifier,
+  type SignatureRequestHandle,
+} from "@xmtp/wasm-bindings";
+import { ApiUrls } from "@/constants";
+import type { XmtpEnv } from "@/types/options";
+import type { Signer } from "@/utils/signer";
+
+const initPromise = init();
```
bennycode: Should this file also make use of `await init()` (like here https://github.com/xmtp/xmtp-js/pull/1618/commits/c4b593000f8210d8186b7b69a43ea65c25c1afa1)?

sdks/browser-sdk/src/utils/inboxState.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @param env - Optional XMTP environment configuration (default: "dev")
+ * @param gatewayHost - Optional gateway host override
+ * @returns The inbox state for the specified inbox IDs
+ */
+export const inboxStateFromInboxIds = async (
+  inboxIds: string[],
+  env?: XmtpEnv,
+  gatewayHost?: string,
+) => {
+  await initPromise;
```
bennycode: Should this file also make use of `await init()` (like here https://github.com/xmtp/xmtp-js/pull/1618/commits/c4b593000f8210d8186b7b69a43ea65c25c1afa1)?

bennycode (changes requested):

sdks/node-sdk/src/DecodedMessage.ts
rygine: this is due to the wonkiness of the type in node bindings. if `message.content.type` is `DecodedMessageContentType.Reply`, then `message.content.reply` is guaranteed to be `EnrichedReply`.

sdks/node-sdk/src/DecodedMessage.ts
rygine: why a default case instead?

sdks/browser-sdk/test/Opfs.test.ts
rygine: they do pass in CI, but i ran into issues when running them with the rest of the tests. i disabled for now. can re-enable in a follow up, just didn't want to spend more time on it.

sdks/node-sdk/scripts/groups.ts
rygine: i'm not sure i understand. this is just a test script. i'm using a `while` loop above as i go through the test accounts to create groups. this bit of code is just following the same pattern.

sdks/browser-sdk/src/workers/opfs.ts
rygine: typically, this won't be the case as a worker is spawned for each instance of the `Opfs` class. and when init'd by `Opfs.create`, the `opfs.init` call is handled automatically.

i don't think it's necessary to have an initializing flag here.

sdks/node-sdk/src/DecodedMessage.ts
bennycode: My IDE indicates that the code path is unreachable because all cases are already covered. The `null` return would only occur if new cases are introduced, right? How about using an exhaustiveness check using `never`, so we get notified by the compiler if a new state is unhandled?

image

sdks/browser-sdk/src/workers/opfs.ts
bennycode: Ok

PR #1638, opened 2026-01-15
Add /new route to xmtp.chat

rygine:
### Add '/new' route in `App` to render `New` for XMTP connection setup and redirect to `/{environment}`
Introduce a top-level `'/new'` route in [apps/xmtp.chat/src/components/App/App.tsx](https://github.com/xmtp/xmtp-js/pull/1638/files#diff-b9e02bfb94bc72dd8e56f99a17137708c479ac80cb68b7f7972b6685f28dd7e5) that renders `New`, which manages XMTP client detection, optional wallet connect, ephemeral account toggling, SCW disabling, and redirects when a client exists.

#### 📍Where to Start
Start with the `New` component logic in [apps/xmtp.chat/src/components/App/New.tsx](https://github.com/xmtp/xmtp-js/pull/1638/files#diff-72efae8bc3aa08c33f479e08063d93f961235b445f9e9b6678c890fe73df27f4), then see its route registration in [apps/xmtp.chat/src/components/App/App.tsx](https://github.com/xmtp/xmtp-js/pull/1638/files#diff-b9e02bfb94bc72dd8e56f99a17137708c479ac80cb68b7f7972b6685f28dd7e5).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized bae78ed.

### DIFF
--- a/apps/xmtp.chat/src/components/App/New.tsx
+++ b/apps/xmtp.chat/src/components/App/New.tsx
@@ -1,0 +1,60 @@
+import { useEffect } from "react";
+import { useNavigate } from "react-router";
+import { LoadingMessage } from "@/components/LoadingMessage";
+import { useXMTP } from "@/contexts/XMTPContext";
+import { useConnectWallet } from "@/hooks/useConnectWallet";
+import { useConnectXmtp } from "@/hooks/useConnectXmtp";
+import { useSettings } from "@/hooks/useSettings";
+
+export const New = () => {
+  const { isConnected } = useConnectWallet();
+  const {
+    environment,
+    ephemeralAccountEnabled,
+    setEphemeralAccountEnabled,
+    autoConnect,
+    setUseSCW,
+  } = useSettings();
+  // if autoConnect is true, this hook will automatically connect to XMTP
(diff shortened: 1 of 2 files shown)


apps/xmtp.chat/src/components/App/New.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  useEffect(() => {
+    // redirect if there's already a client
+    if (client) {
+      void navigate(`/${environment}`);
+      return;
+    }
+
+    // wallet is already connected without autoConnect, connect to XMTP
+    if (isConnected && !autoConnect) {
```
bennycode, line 32: Do we actually need to explicitly check `isConnected`? If it isn’t connected, won’t it still call `connect` in the final block that only checks `!autoConnect`?

bennycode (approved):

apps/xmtp.chat/src/components/App/New.tsx
rygine: the connection mechanism is wonky. these explicit checks are for edge cases that will never happen in reality.

PR #1644, opened 2026-01-16
Upgrade Agent SDK

rygine:
# Summary

- Upgraded Node SDK to latest version
- Refactored Agent SDK to address breaking changes and updates
- Refactored testing to use actual clients instead of mocks
- Updated build process to use rollup

### DIFF
--- a/sdks/agent-sdk/src/core/Agent.ts
+++ b/sdks/agent-sdk/src/core/Agent.ts
@@ -29,6 +29,7 @@ import {
 import { version as appVersion } from "~/package.json";
 import { filter } from "@/core/filter";
 import { getInstallationInfo } from "@/debug";
+import { getValidLogLevels, parseLogLevel } from "@/debug/log";
 import { createSigner, createUser } from "@/user/User";
 import { AgentError, AgentStreamingError } from "./AgentError";
 import { ClientContext } from "./ClientContext";
@@ -166,9 +167,18 @@ export class Agent<ContentTypes = unknown> extends EventEmitter<
     initializedOptions.appVersion ??= `agent-sdk/${appVersion}`;
     initializedOptions.disableDeviceSync ??= true;
 
-    if (process.env.XMTP_FORCE_DEBUG) {
-      const loggingLevel = process.env.XMTP_FORCE_DEBUG_LEVEL || LogLevel.Warn;
-      initializedOptions.loggingLevel = loggingLevel as LogLevel;
+    if (process.env.XMTP_FORCE_DEBUG_LEVEL) {
+      const rawLevel = process.env.XMTP_FORCE_DEBUG_LEVEL;
+      const logLevel = parseLogLevel(rawLevel);
+
+      if (logLevel) {
+        initializedOptions.loggingLevel = logLevel;
+      } else {
+        console.warn(
+          `[WARNING] Invalid XMTP_FORCE_DEBUG_LEVEL "${rawLevel}". Defaulting to "${LogLevel.Warn}". Valid values are: ${getValidLogLevels().join(", ")}`,
+        );
+        initializedOptions.loggingLevel = LogLevel.Warn;
+      }
       initializedOptions.structuredLogging = true;
     }
 
--- a/sdks/agent-sdk/README.md
+++ b/sdks/agent-sdk/README.md
@@ -73,10 +73,9 @@ const agent = await Agent.createFromEnv();
 
 Agents can also recognize the following environment variables:
 
-| Variable                 | Purpose                                                                     | Example                        |
-| ------------------------ | --------------------------------------------------------------------------- | ------------------------------ |
-| `XMTP_FORCE_DEBUG`       | [Activate debugging logs](https://docs.xmtp.org/agents/deploy/debug-agents) | `XMTP_FORCE_DEBUG=true`        |
-| `XMTP_FORCE_DEBUG_LEVEL` | Specify the logging level (defaults to `"Warn"`)                            | `XMTP_FORCE_DEBUG_LEVEL=Debug` |
+| Variable                 | Purpose                                                                                            | Example                        |
+| ------------------------ | -------------------------------------------------------------------------------------------------- | ------------------------------ |
+| `XMTP_FORCE_DEBUG_LEVEL` | [Activate debugging logs](https://docs.xmtp.org/agents/deploy/debug-agents) at the specified level | `XMTP_FORCE_DEBUG_LEVEL=Debug` |
 
 ## Core Concepts
 
--- a/sdks/agent-sdk/src/debug/log.test.ts
+++ b/sdks/agent-sdk/src/debug/log.test.ts
@@ -1,0 +1,64 @@
+import { LogLevel } from "@xmtp/node-sdk";
+import { describe, expect, it } from "vitest";
+import { getValidLogLevels, parseLogLevel } from "@/debug/log";
+
+describe("parseLogLevel", () => {
+  it("should parse lowercase log levels", () => {
+    expect(parseLogLevel("off")).toBe(LogLevel.Off);
+    expect(parseLogLevel("error")).toBe(LogLevel.Error);
+    expect(parseLogLevel("warn")).toBe(LogLevel.Warn);
+    expect(parseLogLevel("info")).toBe(LogLevel.Info);
+    expect(parseLogLevel("debug")).toBe(LogLevel.Debug);
+    expect(parseLogLevel("trace")).toBe(LogLevel.Trace);
+  });
+
+  it("should parse uppercase log levels", () => {
+    expect(parseLogLevel("OFF")).toBe(LogLevel.Off);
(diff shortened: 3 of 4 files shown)


sdks/agent-sdk/rollup.config.js
```diff
@@ -0,0 +1,120 @@
...
+    output: {
+      file: "dist/user.js",
+      format: "es",
+      sourcemap: true,
+    },
+    plugins,
+    external,
+  },
+  {
+    input: "src/user/index.ts",
```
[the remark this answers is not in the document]
rygine: this is intentional

sdks/agent-sdk/src/core/MessageContext.ts
```diff
@@ -1,44 +1,41 @@
...
+
+export type MessageContextParams<
+  MessageContentType = unknown,
+  ContentTypes = unknown,
+> = Omit<AgentBaseContext<ContentTypes>, "message"> & {
+  message: DecodedMessageWithContent<MessageContentType>;
 };
 
 export class MessageContext<
+  MessageContentType = unknown,
```
rygine, line 29: added another type param here so that the context could contain a message with a specific type but have a client with multiple possible types.

sdks/agent-sdk/src/core/Agent.ts
```diff
@@ -176,21 +165,10 @@ export class Agent<ContentTypes = unknown> extends EventEmitter<
...
-      new AttachmentCodec(),
-      new MarkdownCodec(),
-      new ReactionCodec(),
-      new ReadReceiptCodec(),
-      new RemoteAttachmentCodec(),
-      new ReplyCodec(),
-      new TransactionReferenceCodec(),
-      new WalletSendCallsCodec(),
-    ];
+    const upgradedCodecs = [...(initializedOptions.codecs ?? [])];
```
bennycode: Nothing is added to the array anymore, so there are no `upgradedCodecs`. I think we can just remove this variable now and further down do the following:

```ts
    const client = await Client.create(signer, {
      ...initializedOptions,
      codecs: initializedOptions.codecs,
    });
```

sdks/agent-sdk/src/demo.ts
```diff
@@ -91,14 +89,6 @@ agent.on("stop", (ctx) => {
   console.log("Agent stopped", ctx);
 });
 
-agent.on("unknownMessage", (ctx) => {
```
bennycode, line 94: Why was this taken out of the demo? To my understanding, it would be still possible to have custom message types that the Agent doesn’t recognize yet. In such cases we can maybe demo that it shows the "fallback"?

sdks/agent-sdk/src/index.ts
```diff
@@ -21,4 +21,4 @@ export {
 export type * from "@xmtp/node-sdk";
 export { isHexString, validHex } from "@xmtp/node-sdk";
 // Agent SDK
-export * from "./core/index.js";
```
bennycode, line 24: Aren’t ES modules supposed to use the .js extension?

sdks/agent-sdk/src/util/AttachmentUtil.ts
```diff
@@ -23,11 +22,17 @@ export type AttachmentUploadCallback = (
...
-export function downloadRemoteAttachment<ContentTypes>(
+export async function downloadRemoteAttachment(
   remoteAttachment: RemoteAttachment,
-  agent: Agent<ContentTypes>,
 ) {
-  return RemoteAttachmentCodec.load<Attachment>(remoteAttachment, agent.client);
+  const response = await fetch(remoteAttachment.url);
+  if (!response.ok) {
+    throw new Error(
+      `unable to fetch remote attachment at ${remoteAttachment.url}: ${response.status} ${response.statusText}`,
```
bennycode: I suggest enclosing parameters in logging strings with double quotes, as this makes it easier to search through code and log files by clearly indicating where static text ends.

sdks/agent-sdk/test/helpers.ts
```diff
@@ -0,0 +1,114 @@
...
+  type Signer,
+} from "@xmtp/node-sdk";
+import { createWalletClient, http, toBytes } from "viem";
+import { generatePrivateKey, privateKeyToAccount } from "viem/accounts";
+import { sepolia } from "viem/chains";
+
+export const sleep = (ms: number) =>
+  new Promise((resolve) => setTimeout(resolve, ms));
+
+export const createUser = (key?: `0x${string}`) => {
```
bennycode: Can we use the `createUser` exposed in `user/User` here?

sdks/agent-sdk/tsconfig.json
```diff
@@ -3,9 +3,10 @@
     "declaration": true,
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
-    "module": "Node18",
+    "module": "ESNext",
```
bennycode, line 6: Do we need `ESNext` (it's a moving target) or can we use `node20` or similar?

sdks/agent-sdk/vitest.setup.ts
```diff
@@ -0,0 +1,11 @@
+import { unlink } from "node:fs/promises";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { glob } from "fast-glob";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
```
bennycode: I think we can just use `import.meta.dirname` here.

bennycode (changes requested):
Great PR! But I would prefer to keep test files co-located and also cut down on timeouts by retaining some unit tests. Maybe we can separate test files by their extension into pure unit tests & tests that involve the network.

sdks/agent-sdk/src/demo.ts
rygine: sorry, meant to add this back once the refactor was done. thanks for pointing it out. added back in https://github.com/xmtp/xmtp-js/pull/1644/commits/8263819ce8f5430165c1b4acba81648268343c03

sdks/agent-sdk/src/core/Agent.ts
rygine: fixed here https://github.com/xmtp/xmtp-js/pull/1644/commits/6d1a66bafdcef8292667ad0bd8007b007f48e68f

sdks/agent-sdk/src/index.ts
rygine: extensions are not needed when we're using a bundler. now that everything is in a single file, there are no relative imports.

sdks/agent-sdk/src/util/AttachmentUtil.ts
rygine: WDYT of this update?

```
`unable to fetch remote attachment at "${remoteAttachment.url}": [${response.status}] ${response.statusText}`
```

sdks/agent-sdk/test/helpers.ts
rygine: fixed here https://github.com/xmtp/xmtp-js/pull/1644/commits/34aebc8514dd9c00fb2405d854b88f821f89789c

sdks/agent-sdk/vitest.setup.ts
rygine: fixed here https://github.com/xmtp/xmtp-js/pull/1644/commits/34aebc8514dd9c00fb2405d854b88f821f89789c

sdks/agent-sdk/tsconfig.json
rygine: `ESNext` is generally the value used when using a bundler

sdks/agent-sdk/src/util/AttachmentUtil.ts
bennycode: I like it! This will make it much easier to trace an error log back to the code that produced it. 👍

sdks/agent-sdk/tsconfig.json
bennycode: Ok, even the TS docs say: 

> You very likely want "nodenext" for modern Node.js projects and "preserve" or "esnext" for code that will be bundled.

- https://www.typescriptlang.org/tsconfig/#module

bennycode (approved):
Just this one left (https://github.com/xmtp/xmtp-js/pull/1644#discussion_r2700096596) but I am approving the PR to unblock. 🥳

sdks/agent-sdk/src/util/test.ts
```diff
@@ -0,0 +1,61 @@
...
+      type: this.contentType,
+      parameters: {},
+      content: new TextEncoder().encode(JSON.stringify(content)),
+    };
+  }
+  decode(content: EncodedContent): Record<string, string> {
+    const decoded = new TextDecoder().decode(content.content);
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return JSON.parse(decoded);
+  }
```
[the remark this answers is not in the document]
rygine: this is for testing only and not important to guard against

PR #1647, opened 2026-01-16
Revert "Allow setting the gateway host"

bennycode:
Reverts xmtp/xmtp-js#1640


bennycode:
Case proven, setting the gateway to "" causes an error.

PR #1700, opened 2026-02-06
Update XMTP CLI

rygine:
# Summary

This PR rebuilds the XMTP CLI from scratch using the `oclif` framework. It exposes more XMTP functionality and provides better guidance on usage through detailed descriptions and examples.

### DIFF
--- a/packages/xmtp-cli/README.md
+++ b/packages/xmtp-cli/README.md
@@ -1,7 +1,7 @@
 # @xmtp/cli
 
 > [!CAUTION]
-> This CLI is in beta status and ready for you to use. Software in this status may change based on feedback.
+> This CLI is in beta status and ready for you to use. Software in this status may contain bugs or change based on feedback.
 
 A command-line interface for [XMTP](https://xmtp.org), the decentralized messaging protocol.
 
@@ -139,8 +139,8 @@ xmtp conversations create-group <address> \
 
 # Manage members
 xmtp conversation members <conversation-id>
-xmtp conversation add-members <conversation-id> <inbox-id>
-xmtp conversation remove-members <conversation-id> <inbox-id>
+xmtp conversation add-members <conversation-id> <address>
+xmtp conversation remove-members <conversation-id> <address>
 ```
 
 ### Identity
@@ -174,3 +174,27 @@ Use `--verbose` to see detailed client initialization info. When combined with `
 ```bash
 xmtp client info --verbose
 ```
+
+## AI Coding Agent Skill
+
+This package includes an [agent skill](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview) (`skills/xmtp-cli/SKILL.md`) that teaches AI coding agents how to use the XMTP CLI.
+
+**Claude Code**
+
+Add the skill directory to your project's `.claude/settings.json`:
+
+```json
+{
+  "skills": ["./node_modules/@xmtp/cli/skills"]
+}
+```
+
+**Other agents** (Cursor, Windsurf, Codex, etc.)
+
+Use [openskills](https://github.com/numman-ali/openskills) to install the skill:
+
+```bash
+npx openskills install ./node_modules/@xmtp/cli/skills
+```
+
+Or point your agent to `node_modules/@xmtp/cli/skills/xmtp-cli/SKILL.md` directly.


packages/xmtp-cli/package.json
```diff
@@ -1,73 +1,109 @@
...
+    "execa": "^9.6.1",
+    "oclif": "^4.22.70",
     "tsx": "^4.21.0",
     "typescript": "^5.9.3",
     "vitest": "^4.0.18"
   },
   "engines": {
     "node": ">=22"
-  },
-  "publishConfig": {
```
bennycode, line 68: Shall we keep the `publishConfig`?

packages/xmtp-cli/src/utils/config.ts
```diff
@@ -0,0 +1,95 @@
...
+      try {
+        loadEnvFile(DEFAULT_ENV_PATH);
+      } catch {
+        // Silently ignore if neither file exists
+      }
+    }
+  }
+
+  return {
+    walletKey: env.XMTP_WALLET_KEY,
```
bennycode, line 66: It got me thinking that we’re setting `process.env.XMTP_WALLET_KEY` and `process.env.XMTP_DB_ENCRYPTION_KEY` globally. That means any other code running in the same process (oclif plugins, npm dependencies, etc.) could potentially read those private keys straight from `process.env`. We just need these keys for client initialization. How about reading those just for that reason and then zeroizing them?

packages/xmtp-cli/src/utils/output.ts
```diff
@@ -0,0 +1,157 @@
+import { stdout } from "node:process";
+import { IdentifierKind } from "@xmtp/node-sdk";
+
+export function isTTY(): boolean {
+  // isTTY can be undefined when stdout is not a TTY (e.g., piped output)
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  return stdout.isTTY ?? false;
+}
+
+export function jsonStringify(data: unknown, pretty = false): string {
```
bennycode, line 10: We’ve also run into the need for `jsonStringify` in xmtp.chat. Long-term perspective: How about exporting these kinds of utilities directly as part of our bindings, so they’re available to the Node SDK and Browser SDK?

packages/xmtp-cli/src/utils/config.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export function mergeConfig(
+  fileConfig: XmtpConfig,
+  flags: Partial<XmtpConfig>,
+): XmtpConfig {
+  return {
+    walletKey: flags.walletKey ?? fileConfig.walletKey,
+    dbEncryptionKey: flags.dbEncryptionKey ?? fileConfig.dbEncryptionKey,
+    dbPath: flags.dbPath ?? fileConfig.dbPath ?? DEFAULT_DB_PATH,
+    env: flags.env ?? fileConfig.env ?? "dev",
```
bennycode, line 88: Do we log that we fallback to "dev"?

packages/xmtp-cli/package.json
rygine: good catch! added back.

packages/xmtp-cli/src/utils/config.ts
rygine: i'm not sure i understand. these values are required every time the CLI runs. the wallet key is needed for signatures and the DB encryption is used to read the local DB.

packages/xmtp-cli/src/utils/output.ts
rygine: i don't think the bindings are the place for something like this. we could have shared utilities written in typescript that are included in both SDKs.

packages/xmtp-cli/src/utils/config.ts
rygine: there's no logging in the CLI for these values. we could add some text to the output that displays some of these values and the command being executed for clarity.

bennycode: That sounds good. Context really matters, especially in a world where AIs read a lot, and it will also make debugging easier when people can clearly state which environment they were using.

packages/xmtp-cli/src/utils/output.ts
bennycode: Okay. The [convos-node-sdk](https://github.com/xmtplabs/convos-node-sdk) could benefit from this as well.

packages/xmtp-cli/src/commands/conversations/stream-all-messages.ts
```diff
@@ -0,0 +1,153 @@
...
+    const stream = await client.conversations.streamAllMessages({
+      ...streamOptions,
+      disableSync: flags["disable-sync"],
+    });
+
+    // Set up timeout if specified
+    let timeoutId: NodeJS.Timeout | undefined;
+    if (timeoutMs) {
+      timeoutId = setTimeout(() => {
+        void stream.return();
```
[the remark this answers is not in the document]
rygine: this call is intentionally fire-and-forget

packages/xmtp-cli/src/commands/conversations/stream-all-messages.ts
```diff
@@ -0,0 +1,153 @@
...
+          senderInboxId: message.senderInboxId,
+          contentType: message.contentType,
+          content: message.content,
+          sentAt: message.sentAt.toISOString(),
+          deliveryStatus: message.deliveryStatus,
+          kind: message.kind,
+        });
+
+        messageCount++;
+        if (maxCount && messageCount >= maxCount) {
```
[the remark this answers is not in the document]
rygine: this is intentional and has the same effect as the mentioned change

packages/xmtp-cli/src/commands/installation-authorized.ts
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  async run(): Promise<void> {
+    const { args } = await this.parse(InstallationAuthorized);
+
+    const config = this.getConfig();
+    const env = config.env ?? "dev";
+
+    // Convert hex string to Uint8Array
+    const installationBytes = toHexBytes(args.installationId);
```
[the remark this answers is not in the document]
rygine: a crash is fine for this as the hex value would be invalid

packages/xmtp-cli/src/utils/output.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export function formatOutput(data: unknown, forceJson: boolean): string {
+  if (forceJson || !isTTY()) {
+    return jsonStringify(data, true);
+  }
+  return formatHuman(data);
+}
+
+export function formatIdentifierKind(kind: IdentifierKind): string {
```
bennycode, line 19: Should we add an exhaustive check using `never` here, in case a new kind gets introduced later?

packages/xmtp-cli/src/utils/client.ts
```diff
@@ -0,0 +1,89 @@
...
+    throw new Error(
+      `Invalid log level: ${value}. Valid levels: ${validLevels}`,
+    );
+  }
+  return LOG_LEVELS[value];
+}
+
+export function toHexBytes(hex: string): Uint8Array {
+  const prefixedHex = hex.startsWith("0x") ? hex : `0x${hex}`;
+  return hexToBytes(prefixedHex as `0x${string}`);
```
bennycode: I would attach the casting already to the `const  prefixedHex` initialization or maybe introduce a `toHex` predicate.

packages/xmtp-cli/src/commands/conversation/update-description.ts
```diff
@@ -0,0 +1,65 @@
...
+    }),
+  };
+
+  static flags = {
+    ...BaseCommand.baseFlags,
+  };
+
+  async run(): Promise<void> {
+    const { args } = await this.parse(ConversationUpdateDescription);
+    const client = await this.initClient();
```
bennycode, line 46: We repeat the combination of `initClient`, `getConversationById`, and `if (!conversation)` in several places. Would it make sense to extract this into a utility function?

packages/xmtp-cli/src/utils/output.ts
rygine: https://github.com/xmtp/xmtp-js/pull/1700/commits/20130f286fac50b0239ec0fa8e33272526953a85

packages/xmtp-cli/src/utils/client.ts
rygine: https://github.com/xmtp/xmtp-js/pull/1700/commits/6e801091a3fee04833f6cf903644c55807b3c519

packages/xmtp-cli/src/utils/config.ts
rygine: https://github.com/xmtp/xmtp-js/pull/1700/commits/9c95172d8dd04a7fc5479ed0078e62c8cfa7a30f

Issue #1641, opened 2026-01-16
Rate Limit 429 Errors Occurring Below Documented Thresholds


Jr-kenny:
Hi Team,
I'm building a messaging dApp using XMTP and experiencing rate limiting issues that don't align with your documented limits. I'd appreciate your help debugging this.
Issue Description
Users are receiving 429 (Too Many Requests) errors after sending only 7-10 messages, even after refreshing the page. According to your documentation, the write operation limit is 3,000 messages per 5-minute rolling window, but we're hitting this limit at a much lower threshold.

Implementation Details
- XMTP SDK Version: @xmtp/browser-sdk
- Platform: Web-based dApp
- Network: Production

Observed Behavior
1. User sends 7-10 messages successfully
2. Subsequent message attempts return 429 error
3. After page refresh, user can send 2 more messages
4. 429 error returns again

This pattern suggests i am either:
Making excessive background API calls i am not aware of
Incorrectly implementing the client
Experiencing a different rate limit than documented

so Could you help me identify
- What might be causing premature rate limiting
- Best practices to avoid this issue
Thank you for your support

jhaaaa:
Hi there @rygine - just transferred this issue that @Jr-kenny opened in the doc repo. I think it may be more appropriate handled here as they are using the Browser SDK. Thank you so much!

PR #1766, opened 2026-03-19
Add RateLimiter middleware to Agent SDK

mvanhorn:
## Summary

Adds a `RateLimiter` middleware to the Agent SDK for per-sender message throttling using a sliding-window counter. No external dependencies.

## Why this matters

Deployed XMTP agents have no built-in protection against message flooding. The consent system ([docs](https://docs.xmtp.org/chat-apps/user-consent/user-consent)) filters at the conversation level, but there is nothing for per-sender rate control within allowed conversations - particularly in group chats where any member can flood the agent.

Network-level rate limiting has also shown reliability issues ([#1641](https://github.com/xmtp/xmtp-js/issues/1641) - 429 errors below documented thresholds), making agent-side throttling a useful defense layer.

HTTP-level rate limiting already exists in the codebase (`apps/xmtp.chat-api-service/src/middleware/rateLimit.ts` using `express-rate-limit`), but nothing equivalent exists for the Agent SDK's message middleware chain.

## Changes

**New files:**
- `sdks/agent-sdk/src/middleware/RateLimiter.ts` - The middleware class
- `sdks/agent-sdk/src/middleware/RateLimiter.test.ts` - 15 unit tests

**Modified:**
- `sdks/agent-sdk/src/middleware/index.ts` - Export the new middleware (also sorted alphabetically to match convention)

**Usage:**

```ts
import { Agent, RateLimiter } from "@xmtp/agent-sdk";

const limiter = new RateLimiter({
  maxMessages: 5,
  windowMs: 30_000,
  behavior: "reply",
  onRateLimit: (senderId) => console.log(`Rate limited: ${senderId}`),
});

const agent = await Agent.create(/* ... */);
agent.use(limiter.middleware());
```

Config options:
- `maxMessages` / `windowMs` - sliding window parameters (defaults: 10 messages per 60s)
- `behavior: "drop" | "reply"` - silently ignore or send a reply
- `replyText` - custom message when rate limited
- `onRateLimit` - callback for logging/metrics
- `reset()` / `resetSender(id)` - manual state control

## Implementation notes

- Follows the same `.middleware()` return pattern as `PerformanceMonitor` and `ActionWizard`
- Uses `Map<string, number[]>` keyed by `senderInboxId`, similar to `ActionWizard`'s in-memory session store
- Expired timestamps are cleaned on each check, so memory stays bounded
- The middleware calls `next()` only when the sender is under their limit. In the `reduceRight` chain (`Agent.ts` line ~570), placing `RateLimiter` early means rate-limited messages never reach downstream handlers.

## Testing

15 unit tests covering: allow/block thresholds, independent sender tracking, sliding window expiry, drop vs reply behavior, custom reply text, onRateLimit callback, reset/resetSender, and memory cleanup. All tests use `vi.useFakeTimers()` for deterministic time control.

```
✓ src/middleware/RateLimiter.test.ts (15 tests) 5ms
```

This contribution was developed with AI assistance (Claude Code).

> [!NOTE]
> ### Add `RateLimiter` middleware to Agent SDK for per-sender message throttling
> - Adds a new [`RateLimiter`](https://github.com/xmtp/xmtp-js/pull/1766/files#diff-eda691511ce8da3a9a5ad69d691d057dfa379a35f8562964a7437276ce1a1672) middleware class that enforces per-sender message rate limits using a sliding window counter (default: 10 messages per 60s).
> - When a sender exceeds the limit, the middleware either drops the message silently (`drop` mode) or sends a single reply per window (`reply` mode) with a configurable message.
> - Supports an optional `onRateLimit` callback, plus `reset()` and `resetSender()` methods to clear window state.
> - Exports `RateLimiter` from the [`middleware` barrel](https://github.com/xmtp/xmtp-js/pull/1766/files#diff-f192e3112946499dc484b5644ca15a29842659564c2aee9764b998d4b005a98a).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2082e4a.</sup>
> 
>

### DIFF
--- a/sdks/agent-sdk/src/middleware/RateLimiter.ts
+++ b/sdks/agent-sdk/src/middleware/RateLimiter.ts
@@ -23,13 +23,19 @@ export interface RateLimiterConfig {
  * agent.use(limiter.middleware());
  * ```
  */
+interface SenderWindow {
+  count: number;
+  windowStart: number;
+  replied: boolean;
+}
+
 export class RateLimiter<ContentTypes = unknown> {
   #maxMessages: number;
   #windowMs: number;
   #behavior: "drop" | "reply";
   #replyText: string;
   #onRateLimit?: (senderInboxId: string) => void;
-  #timestamps = new Map<string, number[]>();
+  #windows = new Map<string, SenderWindow>();
 
   constructor(config: RateLimiterConfig = {}) {
     this.#maxMessages = config.maxMessages ?? 10;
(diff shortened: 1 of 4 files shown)


sdks/agent-sdk/src/middleware/RateLimiter.ts
```diff
@@ -0,0 +1,93 @@
...
+      const senderId = ctx.message.senderInboxId;
+
+      if (this.#isAllowed(senderId)) {
+        await next();
+        return;
+      }
+
+      this.#onRateLimit?.(senderId);
+
+      if (this.#behavior === "reply") {
```
neekolas: I don't think it's a good idea to reply to a sender more than once per period. Otherwise you end up multiplying the DOS, since now you are sending one message for every rate limited message you receive.

sdks/agent-sdk/src/middleware/RateLimiter.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.#maxMessages = config.maxMessages ?? 10;
+    this.#windowMs = config.windowMs ?? 60_000;
+    this.#behavior = config.behavior ?? "drop";
+    this.#replyText =
+      config.replyText ??
+      "You're sending messages too quickly. Please wait a moment.";
+    this.#onRateLimit = config.onRateLimit;
+  }
+
+  #isAllowed(senderInboxId: string): boolean {
```
neekolas: If we were willing to give up some precision, and have the rate limit be calculated for every window period, we could store the limits much more efficiently as a simple count rather than an array of timestamps.

Storing the timestamps maybe isn't so bad on its own, but we don't clean up old timestamps unless another message comes in from the same sender. So if you have an agent that is receiving messages from many addresses memory usage will expand forever.

mvanhorn:
Fixed in 2082e4a6:
1. Reply is now sent once per window, not per blocked message. Subsequent blocked messages are silently dropped.
2. Replaced the timestamp array with a `{ count, windowStart, replied }` struct per sender. No more unbounded growth from inactive senders - the window resets cleanly on expiry.

insipx:

PR #1846, opened 2026-08-13
fix(agent-sdk): LimitedMap.set no longer evicts on update at capacity

### LINKED ISSUE #3981: Bug: LimitedMap evicts a valid entry when updating an existing key at capacity
### Source

Agent SDK

### Describe the bug

`LimitedMap.set()` evicts the oldest entry whenever the map is at its size limit — even when the key being set **already exists**. Updating the value of an existing key does not grow the map, so it should not trigger an eviction. As written, updating an existing key while at capacity removes an unrelated, still-valid entry and shrinks the map below its configured limit.

https://github.com/xmtp/xmtp-js/blob/615d4e71c092ec5db6bb08039117afbb94c0ce36/sdks/agent-sdk/src/util/LimitedMap.ts#L9-L17

```ts
set(key: K, value: V) {
  if (this.#map.size >= this.#limit) {
    const it = this.#map.keys().next();
    if (!it.done) {
      this.#map.delete(it.value);
    }
  }
  this.#map.set(key, value);
}
```

**Impact:** `LimitedMap` backs the name-resolution cache in `sdks/agent-sdk/src/user/NameResolver.ts` (`new LimitedMap<string, string | null>(1000)`). Re-resolving an address that is already cached calls `set()` again; once the cache is full, each such update evicts a *different*, still-valid entry. The cache then holds fewer than 1000 entries and performs avoidable repeat resolutions.

**Suggested fix:** only evict when inserting a *new* key would exceed the limit:

```ts
if (!this.#map.has(key) && this.#map.size >= this.#limit) {
```

### Steps to reproduce the bug

```ts
const map = new LimitedMap<string, number>(2);
map.set("a", 1);
map.set("b", 2); // { a, b } — at limit
map.set("b", 4); // update existing key "b"

map.get("a"); // => undefined  (expected: 1)
map.get("b"); // => 4
// the map now holds only 1 entry instead of 2
```

ygd58:
Fixes xmtp/libxmtp#3981

`set()` evicted the oldest entry whenever the map was at capacity, without checking whether the incoming key already existed. Updating the value of an existing key doesn't grow the map, so evicting in that case drops an unrelated, still-valid entry for no reason.

Now checks `has(key)` first and only evicts when the key is genuinely new and the map is already full.

Added three tests: normal eviction on a new key at capacity, no eviction when updating an existing key at capacity (the bug this fixes), and correct eviction resuming once a genuinely new key is added afterward. Verified the new test fails against the old implementation and passes against the fix.

`yarn vitest run src/util/LimitedMap.test.ts` — 3/3 passing. `yarn typecheck` and `eslint`/`prettier` clean on both changed files.

> [!NOTE]
> ### Fix `LimitedMap.set` to skip eviction when updating an existing key at capacity
> Previously, `LimitedMap.set` evicted the oldest entry whenever the map was at capacity, even if the key being set already existed. The fix adds a check so eviction only occurs when the incoming key is new. Tests cover three cases: new-key eviction, update-with-no-eviction, and new-key eviction after an update.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 62a90f0.</sup>
> 
>

### DIFF
--- a/.changeset/limited-map-eviction-fix.md
+++ b/.changeset/limited-map-eviction-fix.md
@@ -1,0 +1,5 @@
+---
+"@xmtp/agent-sdk": patch
+---
+
+Fixed `LimitedMap.set()` evicting a valid entry when updating an existing key at capacity. It now checks whether the key already exists before evicting, so updating a key's value at capacity no longer drops an unrelated entry.
--- a/sdks/agent-sdk/src/util/LimitedMap.test.ts
+++ b/sdks/agent-sdk/src/util/LimitedMap.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, it } from "vitest";
+import { LimitedMap } from "@/util/LimitedMap";
+
+describe("LimitedMap", () => {
+  it("evicts the oldest entry when a new key is added at capacity", () => {
+    const map = new LimitedMap<string, number>(2);
+    map.set("a", 1);
+    map.set("b", 2);
+    map.set("c", 3);
+
+    expect(map.get("a")).toBeUndefined();
+    expect(map.get("b")).toBe(2);
+    expect(map.get("c")).toBe(3);
+  });
+
+  it("does not evict any entry when updating an existing key at capacity", () => {
+    const map = new LimitedMap<string, number>(2);
+    map.set("a", 1);
+    map.set("b", 2);
+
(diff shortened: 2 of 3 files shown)


insipx:

